PR #283, opened 2015-01-28
Add static convenience method for compiling without exit.

JakeWharton:


JakeWharton:
Doh! https://github.com/google/error-prone/issues/284

PR #285, opened 2015-01-28
Add static convenience method for compiling without exit.

### LINKED ISSUE #284: One-shot static compilation method without exit
In integration with our tooling it would be convenient to have a one-shot static method to perform compilation. Invoking `main` results in `System.exit` being called. Normally I would just add a `compile` overload which only took a `String[]` (#283), but that conflicts with the instance method of the same signature.

This isn't a must-have, of course. I can continue invoking the builder or just pass my own wrapper around `System.err` as a second argument to the existing static method, but it'd be nice.

Some alternatives:

 * Change the static helpers to `compileForResult` (or something) to more accurately describe what they do.
 * Change the instance `compile(String[])` method to be named `run`, `execute`, etc. Not sure how set-in-stone this method is. My guess would be "very" though :)

Thoughts?

JakeWharton:
As this method would result in a collision with an instance method of the same signature, that method has been renamed to 'run'.

Closes #284.

### DIFF
--- a/core/src/main/java/com/google/errorprone/ErrorProneCompiler.java
+++ b/core/src/main/java/com/google/errorprone/ErrorProneCompiler.java
@@ -60,7 +60,7 @@ public class ErrorProneCompiler {
    * @param args the same args which could be passed to javac on the command line
    */
   public static void main(String[] args) {
-    System.exit(new ErrorProneCompiler.Builder().build().compile(args).exitCode);
+    System.exit(compile(args).exitCode);
   }
 
   /**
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/google/errorprone/ErrorProneCompiler.java
```diff
@@ -142,7 +152,7 @@ public Builder report(ScannerSupplier scannerSupplier) {
     }
   }
 
-  public Result compile(String[] args) {
+  public Result run(String[] args) {
```
cushon: Can you also rename the other instance methods named `compile`? The only usages are in `CompilationTestHelper` and `ErrorProneTestCompiler`.

JakeWharton: Yep, good catch. Done.

cushon:
Looks good, thanks.

Issue #287, opened 2015-01-28
Running ant example causes Oracle javac to crash.


JakeWharton:
I'm experiencing problems with the compiler in my regular build, but in trying to craft a reproducible sample (by cannibalizing the example) I ran into the following exceptions internal to `javac` on both JDK 7 and 8.

Test script:
```bash
curl http://www.carfab.com/apachesoftware/ant/binaries/apache-ant-1.9.4-bin.zip > ant.zip
unzip ant.zip

TIMESTAMP=`curl https://oss.sonatype.org/content/repositories/snapshots/com/google/errorprone/error_prone_ant/2.0-SNAPSHOT/maven-metadata.xml | head -8 | tail -1 | sed  's/[^0-9.]//g'`
echo "Timestamp: ${TIMESTAMP}"
BUILD=`curl https://oss.sonatype.org/content/repositories/snapshots/com/google/errorprone/error_prone_ant/2.0-SNAPSHOT/maven-metadata.xml | head -9 | tail -1 | sed 's/[^0-9]//g'`
echo "Build number: ${BUILD}"

curl "https://oss.sonatype.org/content/repositories/snapshots/com/google/errorprone/error_prone_ant/2.0-SNAPSHOT/error_prone_ant-2.0-${TIMESTAMP}-${BUILD}.jar" > error_prone_ant.jar
mv error_prone_ant.jar apache-ant-1.9.4/lib/

git clone https://github.com/google/error-prone.git
cd error-prone/examples/ant

../../../apache-ant-1.9.4/bin/ant
```
If you create a temp directory somewhere you can quickly run this with:
```
curl https://gist.githubusercontent.com/JakeWharton/7bc348ceb1b9c27130f2/raw/676b2c18fbd95e46f6da9f85f4f474d459ecb956/test.sh | sh
```

Oracle JDK 8:
```
compile:
    [mkdir] Created dir: /Users/jw/Desktop/error-prone-test/error-prone/examples/ant/build
    [javac] Compiling 1 source file to /Users/jw/Desktop/error-prone-test/error-prone/examples/ant/build
    [javac] An exception has occurred in the compiler (1.8.0-internal_bootstrap). Please file a bug at the Java Developer Connection (http://java.sun.com/webapps/bugreport)  after checking the Bug Parade for duplicates. Include your program and the following diagnostic in your report.  Thank you.
    [javac] java.lang.NullPointerException
[...]

PR #290, opened 2015-01-30
Make ant work with javac8 and JDK7

cushon:
This generalizes some classloading infrastructure that could be shared
with maven.

### DIFF
--- a/core/src/main/java/com/google/errorprone/internal/NonDelegatingClassLoader.java
+++ b/core/src/main/java/com/google/errorprone/internal/NonDelegatingClassLoader.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright 2015 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.errorprone.internal;
+
+import com.google.common.collect.ImmutableSet;
+
+import java.net.URLClassLoader;
+import java.util.Set;
+
+/**
+ * A non-delegating {@link java.net.URLClassLoader} that searches its own
+ * resource path <i>before</i> the runtime classpath, reversing the usual
+ * classloader delegation model. This makes it possible to override runtime
+ * classes.
+ *
+ * <p>The classloader is also given the classloader from the environment it is
+ * created in, and a whitelist of classes that it should load from that
+ * environment. This list must include anything that crosses classloader
+ * boundaries (for example, the input or return types of the callback).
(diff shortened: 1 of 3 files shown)


cushon:
This also fixes the less interesting part of #287

core/src/main/java/com/google/errorprone/internal/NonDelegatingClassLoader.java
```diff
@@ -0,0 +1,72 @@
...
+
+package com.google.errorprone.internal;
+
+import com.google.common.collect.ImmutableSet;
+
+import java.net.URLClassLoader;
+import java.util.Set;
+
+/**
+ * A non-delegating {@link java.net.URLClassLoader} that searches it's own resource path _before_
```
eaftan: nit: it's -> its

core/src/main/java/com/google/errorprone/internal/NonDelegatingClassLoader.java
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * @author cushon@google.com
+ */
+public class NonDelegatingClassLoader extends URLClassLoader {
+  private final ClassLoader original;
+  private final ImmutableSet<String> whiteList;
+
+  public static NonDelegatingClassLoader create(Set<String> whiteList) {
+    ClassLoader contextLoader = Thread.currentThread().getContextClassLoader();
+    if (!(contextLoader instanceof URLClassLoader)) {
```
eaftan: Can we always be sure the default classloader is a URLClassLoader?

core/src/main/java/com/google/errorprone/internal/NonDelegatingClassLoader.java
cushon: Fixed.

Heh, github markdown recognized `_before_` and `<i>before<i>` as _before_.

core/src/main/java/com/google/errorprone/internal/SandboxedClassLoaderRunner.java
```diff
@@ -0,0 +1,50 @@
...
+import com.google.common.collect.ImmutableSet;
+
+/**
+ * Executes a callback inside a special classloader that searches its own resource path _before_
+ * delegating to the runtime classpath. This allows error-prone's javac8 to be run on JDK7, and is
+ * morally equivalent to invoking java with {@code -Xbootclasspath/p:<path to javac.jar>}.
+ *
+ * @author cushon@google.com
+ */
+public class SandboxedClassLoaderRunner {
```
eaftan: nit: maybe rename to something like NonDelegatingClassLoaderRunner?  "Sandboxed" implies that this is security related.

core/src/main/java/com/google/errorprone/internal/NonDelegatingClassLoader.java
cushon: This assumption is unfortunate, but it holds for all known plugins and I don't know any reasonable alternatives.

eaftan: OK, fine with me

eaftan:
LGTM aside from the nits.  Go ahead and merge when you're done.

core/src/main/java/com/google/errorprone/internal/SandboxedClassLoaderRunner.java
cushon: Done

PR #295, opened 2015-02-05
Add more ant examples

cushon:


examples/ant/compilerclasspath/build.xml
```diff
@@ -0,0 +1,35 @@
...
+  <path id="errorprone">
+    <pathelement location="../../../ant/target/error_prone_ant-2.0.2-SNAPSHOT.jar"/>
+  </path>
+
+  <target name="compile">
+    <mkdir dir="build"/>
+    <mkdir dir="build"/>
+    <javac debug="true"
+      destdir="build"
+      source="1.7"
```
eaftan: As in the other review, are source and target here necessary?

examples/ant/compilerclasspath/build.xml
```diff
@@ -0,0 +1,35 @@
...
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+  -->
+<project name="error prone example" default="compile">
+  <path id="errorprone">
+    <pathelement location="../../../ant/target/error_prone_ant-2.0.2-SNAPSHOT.jar"/>
```
eaftan: It would be nice if the version number were not hardcoded here.  Any ideas?  At worst we should at least document updating this in our release instructions.

Also do we really want this pointing to the SNAPSHOT jar?

examples/ant/compilerclasspath/build.xml
cushon: Nope, fixed.

examples/ant/compilerclasspath/build.xml
cushon: I replaced it with a glob, which at least avoids hardcoding the version number.

The thinking behind using the snapshot jar is that it lets you do a local build and then try the examples without copying / renaming anything.

eaftan: Ah, makes sense.

eaftan:
LGTM

Issue #292, opened 2015-02-03
Ant integration for OpenJDK7 does not work as described


PhilippWendler:
The documentation claims that it is sufficient to add a `compilerclasspath` element to the `javac` task in the ant build file for error prone to work (in addition to the `compiler` attribute).

I am using Error-Prone 2.0.1, OpenJDK7 1.7.0_75, and ant 1.9.3 on Linux and I get the following error:
```
build.xml:17: java.lang.LinkageError: Unable to create runner.
        at com.google.errorprone.internal.NonDelegatingClassLoaderRunner.run(NonDelegatingClassLoaderRunner.java:40)
        at com.google.errorprone.ErrorProneAntCompilerAdapter.execute(ErrorProneAntCompilerAdapter.java:39)
        at org.apache.tools.ant.taskdefs.Javac.compile(Javac.java:1153)
        at org.apache.tools.ant.taskdefs.Javac.execute(Javac.java:930)
        at org.apache.tools.ant.UnknownElement.execute(UnknownElement.java:292)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:606)
        at org.apache.tools.ant.dispatch.DispatchUtils.execute(DispatchUtils.java:106)
        at org.apache.tools.ant.Task.perform(Task.java:348)
        at org.apache.tools.ant.Target.execute(Target.java:435)
        at org.apache.tools.ant.Target.performTasks(Target.java:456)
        at org.apache.tools.ant.Project.executeSortedTargets(Project.java:1393)
        at org.apache.tools.ant.Project.executeTarget(Project.java:1364)
        at org.apache.tools.ant.helper.DefaultExecutor.executeTargets(DefaultExecutor.java:41)
        at org.apache.tools.ant.Project.executeTargets(Project.java:1248)
        at org.apache.tools.ant.Main.runBuild(Main.java:851)
        at org.apache.tools.ant.Main.startAnt(Main.java:235)
        at org.apache.tools.ant.launch.Launcher.run(Launcher.java:280)
        at org.apache.tools.ant.launch.Launcher.main(Launcher.java:109)
[...]

PR #296, opened 2015-02-05
Add installation info to README

cushon:


README.md
```diff
@@ -4,11 +4,70 @@ Catch common Java mistakes as compile-time errors
...
+### Ant
+
+Download the latest release of
+[error_prone_ant.jar](https://repo1.maven.org/maven2/com/google/errorprone/error_prone_ant)
+from maven, and use the following javac task in your `build.xml`.
+
+```xml
+<javac destdir="build"
+  compiler="com.google.errorprone.ErrorProneAntCompilerAdapter"
+  source="1.7" target="1.7" encoding="UTF-8" debug="true"
```
eaftan: Do we need this line?  Shouldn't source and target default to the host JDK version?

README.md
```diff
@@ -4,11 +4,70 @@ Catch common Java mistakes as compile-time errors
...
+## Developing error-prone
+
+To develop and build error-prone, see our documentation on the
+[wiki](https://github.com/google/error-prone/wiki/For-Developers).
 
 ## Links
 
 - [Javadoc](http://errorprone.info/api/latest/)
-- Pre-release snapshots are available from [sonatype's snapshot repository](https://oss.sonatype.org/content/repositories/snapshots/com/google/errorprone/).
+- Pre-release snapshots are available from [sonatype's snapshot
```
eaftan: nit: sonatype -> Sonatype

eaftan:
LGTM other than the minor comments

README.md
alexeagle: do you need to tell where to put the file after its downloaded?

README.md
cushon: Done

README.md
cushon: > Do we need this line? Shouldn't source and target default to the host JDK version?

Thanks, I removed the source/target options.

> do you need to tell where to put the file after its downloaded?

It needs to match the `pathelement` in the config. I tried to make that clearer by using `/path/to/error_prone_ant.jar`. I like this ant configuration option because it doesn't rely on `$ANT_HOME` / `~/.ant/lib`. I lifted the example straight from #292.

PR #297, opened 2015-02-09
Mention @CanIgnoreReturnValue in CheckReturnValue docs

eaftan:


core/src/main/java/com/google/errorprone/bugpatterns/CheckReturnValue.java
```diff
@@ -51,7 +51,12 @@
     summary = "Ignored return value of method that is annotated with @CheckReturnValue",
     explanation = "The JSR 305 @CheckReturnValue annotation marks methods whose return values "
         + "should be checked.  This error is triggered when one of these methods is called but "
-        + "the result is not used.",
+        + "the result is not used.\n\n"
+        + "@CheckReturnValue may be applied to a class or package to indicate that all methods "
+        + "in that class or package must have their return values checked.  For convenience, "
+        + "we provide an annotation, @CanIgnoreReturnValue, to exempt specific methods from "
```
cushon: `@CanIgnoreReturnValue` can also be applied to classes.

eaftan: Thanks for the catch.  Changed it to say that you can use @CanIgnoreReturnValue to exempt specific methods *or classes* from this behavior.

cushon:
LGTM

PR #301, opened 2015-02-11
Replace "assert false" with "throw new AssertionError()"

semonte:
Added a bug pattern for the "assert false" case.

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/AssertFalse.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/AssertFalse.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2015 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.errorprone.bugpatterns;
+
+import com.google.errorprone.BugPattern;
+import com.google.errorprone.VisitorState;
+import com.google.errorprone.bugpatterns.BugChecker.AssertTreeMatcher;
+import com.google.errorprone.fixes.Fix;
+import com.google.errorprone.fixes.SuggestedFix;
+import com.google.errorprone.matchers.Description;
+
+import com.sun.source.tree.AssertTree;
+import com.sun.source.tree.ExpressionTree;
+
+import static com.google.errorprone.BugPattern.Category.JDK;
+import static com.google.errorprone.BugPattern.MaturityLevel.EXPERIMENTAL;
+import static com.google.errorprone.BugPattern.SeverityLevel.WARNING;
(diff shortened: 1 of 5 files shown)


eaftan:
Thanks for the contribution!  I will review in a minute, but in the meantime please take a look at our instructions for contributors (https://github.com/google/error-prone/blob/master/CONTRIBUTING.md).  In particular we need you to sign the Google Individual Contributor License Agreement (https://cla.developers.google.com/about/google-individual?csw=1).

semonte:
Thanks for the quick feedback! I have already signed the agreement so you should be able to see it in the system.

core/src/main/java/com/google/errorprone/bugpatterns/AssertFalse.java
```diff
@@ -0,0 +1,65 @@
...
+import static com.google.errorprone.BugPattern.SeverityLevel.WARNING;
+import static com.google.errorprone.matchers.Matchers.kindIs;
+import static com.sun.source.tree.Tree.Kind.BOOLEAN_LITERAL;
+import static com.sun.tools.javac.tree.JCTree.JCLiteral;
+
+/**
+ * @author sebastian.h.monte@gmail.com (Sebastian Monte)
+ */
+@BugPattern(name = "AssertFalse",
+    summary = "Assert false should not be used",
```
eaftan: The summary and explanation fields are presented to the user as the compile error message and documentation page content on errorprone.info, respectively.  We try to be as specific and clear as possible so that people understand why we're flagging the error and how to fix it.  Otherwise people tend to assume that the tool is wrong and suppress the error instead of fixing it.

For summary, I would say something like, "Assertions may be disabled at runtime and do not guarantee that execution will halt here; consider throwing an exception instead."  For explanation, something like, "Java assertions do not necessarily execute at runtime; they may be enabled and disabled depending on which options are passed to the JVM invocation. An `assert false` statement may be intended to ensure that the program never proceeds beyond that statement.  If the correct execution of the program depends on that being the case, consider throwing an exception instead, so that execution is halted regardless of runtime configuration."

core/src/main/java/com/google/errorprone/bugpatterns/AssertFalse.java
```diff
@@ -0,0 +1,65 @@
...
+                  + " executed except in case of a bug. It is better to"
+                  + " throw an AssertionError so an exception is raised"
+                  + " regardless whether assertions are enabled.",
+    category = JDK, severity = WARNING, maturity = EXPERIMENTAL)
+public class AssertFalse extends BugChecker implements AssertTreeMatcher {
+
+  @Override
+  public Description matchAssert(AssertTree tree, VisitorState state) {
+    ExpressionTree condition = tree.getCondition();
+    if (kindIs(BOOLEAN_LITERAL).matches(condition, state)) {
```
eaftan: You can replace this with `Matchers.booleanLiteral(false)`.  Also, I think it would be generally useful to have a matcher that takes a AssertTree and a Matcher\<ExpressionTree\>, and applies the matcher to the condition.  Could you add something like that to the Matchers class, and use it here?  It would look something like:
```java
public static Matcher<AssertTree> assertionWithCondition(final Matcher<ExpressionTree> conditionMatcher) {
  return new Matcher<AssertTree>() {
    @Override
    public boolean matches(AssertTree tree, VisitorState state) {
      return conditionMatcher.matches(tree.getCondition(), state);
    }
  };
}
```

eaftan:
Thanks, I confirmed that you signed the contributor agreement.

semonte:
Thanks for the comments and sorry for late reply. I updated the code according to your suggestions.

eaftan:
Made some formatting changes to fit with Google style and merged into head in commit da97425121c4b07e6494b2a6ed044f62b1b63493.  Thanks for the contribution!

PR #304, opened 2015-03-04
Delete EndPosTableUtil

### LINKED ISSUE #228: Investigate turning on tree end positions by default
_[Original issue](https://code.google.com/p/error-prone/issues/detail?id=228) created by **eaftan@google.com** on 2014-01-29 at 10:26 PM_

---

Currently javac only computes tree end positions if the -Xjcov option is passed at the command line, or a custom DiagnosticListener is provided.  Turning on tree end positions incurs a ~50% memory penalty in javac.

We need tree end positions to construct suggested fixes, so currently we reparse the tree if we encounter any errors. However, some checks need to know the end position before we know if there are any errors (e.g., LongLiteralLowerCaseSuffix), and anyway we're not happy with our reparsing hack.

Investigate whether it is feasible to turn on -Xjcov by default.  We can try building some larger open source projects with -Xjcov on and see if they pass. 

Note that there has been a patch submitted to upstream javac to reduce memory usage (http://openjdk.5641.n7.nabble.com/javac-ending-positions-generation-and-DiagnosticListener-tt170348.html), but even if accepted it won't help our external users who are on older versions of javac.

cushon:
It's been absolete since I fixed #228.


eaftan:
LGTM.  I'll go ahead and merge.

PR #310, opened 2015-03-25
Bazel

cushon:


cushon:
This is still very exploratory, but anyone else trying to use bazel with Error Prone will need this.

PR #312, opened 2015-03-26
Added a checker for @Nullable annotated primitives.

semonte:


cushon:
Thanks! Looks good overall.

core/src/main/java/com/google/errorprone/bugpatterns/NullablePrimitive.java
```diff
@@ -0,0 +1,61 @@
...
+ * @author sebastian.h.monte@gmail.com (Sebastian Monte)
+ */
+@BugPattern(name = "NullablePrimitive",
+            summary = "@Nullable should not be used for primitive types.",
+            explanation = "Primitives can never be null, annotating a primitive with @Nullable"
+                + " may be hinting at an intent that cannot be fulfilled.",
+            category = JDK, severity = WARNING, maturity = EXPERIMENTAL)
+public class NullablePrimitive extends BugChecker implements BugChecker.AnnotationTreeMatcher {
+
+    private static final String NULLABLE_ANNOTATION = "javax.annotation.Nullable";
```
cushon: Consider matching on any annotation named "Nullable" - there are many different implementations.

core/src/main/java/com/google/errorprone/bugpatterns/NullablePrimitive.java
```diff
@@ -0,0 +1,61 @@
...
+
+    private Matcher<AnnotationTree> nullableAnnotationMatcher = isType(NULLABLE_ANNOTATION);
+
+    @Override
+    public Description matchAnnotation(AnnotationTree tree, VisitorState state) {
+        if (nullableAnnotationMatcher.matches(tree, state)) {
+            if (isPrimitiveType().matches(getAnnotatedNode(state), state)) {
```
cushon: nit: create a constant field for the isPrimitiveType() matcher

core/src/test/resources/com/google/errorprone/bugpatterns/NullablePrimitiveNegativeCases.java
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * @author sebastian.h.monte@gmail.com (Sebastian Monte)
+ */
+public class NullablePrimitiveNegativeCases {
+
+  @Nullable
+  Integer a;
+
+  public void method(@Nullable Integer a) {
```
cushon: Can you add positive and negative tests for methods annotated with @Nullable, to make sure the return type is tested correctly?

core/src/main/java/com/google/errorprone/bugpatterns/NullablePrimitive.java
Stephan202: There's also [javax.annotation.CheckForNull](https://jsr-305.googlecode.com/svn/trunk/javadoc/javax/annotation/CheckForNull.html).

Going further: annotating a primitive `@Nonnull` (or any variation thereof) is redundant. Should the checker flag those as well?

cushon: It's redundant, but it's not wrong the same way `@Nullable int` is. It seems less likely to indicate a problem.

core/src/test/resources/com/google/errorprone/bugpatterns/NullablePrimitiveNegativeCases.java
semonte: Good point! I added a check for methods.

core/src/main/java/com/google/errorprone/bugpatterns/NullablePrimitive.java
semonte: A changed the code so that we match simple names rather than fully qualified names.

cushon:
Committed as 9e9530f6c6502f03e91e68544d4e9441176ab195

PR #319, opened 2015-04-18
Initial bazel build

cushon:
This is still exploratory, but it builds and passes tests.


annotation/BUILD
```diff
@@ -0,0 +1,12 @@
+# Description:
```
eaftan: Are these BUILD files going to collide with the internal copies?

cushon: The BUILD files don't get synced at all right now. It'd be nice to converge them, but that's a ways off.

PR #326, opened 2015-05-13
Stop ignoring everything under .idea

tom-smalls:
As a developer who doesn't work at Google, it is hard to work on error prone (and create pull requests) without violating your styling rules as the project code codeStyleSettings.xml is not commited to GitHub.

By not ignoring everything under .idea this will be possible to do, and make external developer's lives easier.

If this is accepted can someone from Google the correct xml files?

Thanks,

Tom


eaftan:
It looks like we do have a projectCodeStyles.xml file committed:
https://github.com/google/error-prone/blob/master/.idea/projectCodeStyle.xml

It's just that it's wrong.  I'll add the correct one.

cushon:
Also un-ignoring `.idea` sounds reasonable. I think added the suppression to avoid churn when working on different machines / intellij versions.

@tom-smalls - can you add a reference for where the list of files came from? Was it adapted from something like https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore?

tom-smalls:
It's just what we use at work to be perfectly honest

tom-smalls:
It's what we use at work to be honest, although I believe it is based on http://devnet.jetbrains.net/docs/DOC-1186

Shall I close the pull request? Seems to have brought in a load of other changes, bit annoying.

cushon:
Fair enough, I was just curious. I squashed the commits and pushed as 72bd09a5a52f92be7d3bd3e5595b3b6550b5c6fe.

Thanks!

PR #331, opened 2015-05-27
Update bugpattern file generation

cushon:
-add summary to frontmatter
-emit frontmatter xor html metadata
-add script for updating generated content


docgen/src/main/java/com/google/errorprone/BugPatternFileGenerator.java
```diff
@@ -114,18 +115,21 @@ private static MessageFormat constructPageTemplate(
...
         + "-->\n\n");
 
-    result.append("<div style=\"float:right;\"><table id=\"metadata\">\n"
-        + "<tr><td>Category</td><td>{3}</td></tr>\n"
-        + "<tr><td>Severity</td><td>{4}</td></tr>\n"
-        + "<tr><td>Maturity</td><td>{5}</td></tr>\n"
-        + "</table></div>\n\n"
-        + "# Bug pattern: {1}\n"
-        + "__{8}__\n");
+    if (!generateFrontMatter) {
```
eaftan: Could you make this an else for the if on line 100?

eaftan:
Just that one nit, then go ahead and merge.

docgen/src/main/java/com/google/errorprone/BugPatternFileGenerator.java
cushon: I wanted the AUTO-GENERATED comment as early as possible, but it can't be before the frontmatter.

cushon:
Merged as ccdc295fb8ca21dedc0226f111d2c2af099c2d37

PR #332, opened 2015-05-27
Revamp site

cushon:
demo: http://cushon.org/error-prone/bugpattern/DoubleCheckedLocking

This fixes some miscellaneous breakage in the bugpattern docs caused by skew between the internal and external sites. I also changed the appearance to (hopefully) make it more readable. All of the good bits were stolen form bazel.io / bootstrap.

Please don't merge this as-is: I'll rebase it after #331 goes it.


CNAME
```diff
@@ -1 +0,0 @@
-errorprone.info
```
eaftan: I think we do need the CNAME file, based on the revision history of this file.  (I know nothing more than that, but it looks like Alex tried deleting it before, and then had to revert the change)

eaftan:
Looks good!  One comment about the CNAME file, but otherwise rebase and merge whenever you're ready.

CNAME
cushon: Thanks for the catch. I don't want to merge this, it was just for testing with my local fork. (Otherwise github complains because two repositories are trying to use the same CNAME...)

I'll get rid of it before merging.

PR #336, opened 2015-05-27
Update ASTHelpersTest.java

ashwin244:
Update two failing test cases testGetActualStartPosition and testGetActualStartPositionWithWhitespace

### DIFF
--- a/core/src/test/java/com/google/errorprone/util/ASTHelpersTest.java
+++ b/core/src/test/java/com/google/errorprone/util/ASTHelpersTest.java
@@ -62,7 +62,7 @@ public class ASTHelpersTest extends CompilerBasedAbstractTest {
         "    i = -1;",
         "  }",
         "}");
-    assertCompiles(literalExpressionMatches(literalHasActualStartPosition(59)));
+    assertCompiles(literalExpressionMatches(literalHasActualStartPosition(62))); //test failing : updated from 59 -> 62
   }
 
   @Test
@@ -74,7 +74,7 @@ public class ASTHelpersTest extends CompilerBasedAbstractTest {
         "    i = -     1;",
         "  }",
         "}");
-    assertCompiles(literalExpressionMatches(literalHasActualStartPosition(59)));
+    assertCompiles(literalExpressionMatches(literalHasActualStartPosition(62))); //test failing : updated from 59 -> 62
   }
 
   private Matcher<LiteralTree> literalHasActualStartPosition(final int startPosition) {


cushon:
Those tests aren't failing on travis. Can you describe the environment where you saw the failure? (which version of Java were you using, were you running them from an IDE, ...)

ashwin244:
I tested those test cases on Eclipse IDE with Java 1.8.0_31 runtime version. There were several other testcases failing too on my local, but when I see the travis logs they shows to be passing as well.

cushon:
Interesting. Is it possible that the eclipse classpath puts the tools.jar from your local JDK before the version of javac that's distributed with Error Prone?

eaftan:
Looking at the content of the string in the test case, the literal *does* begin at offset 59, not 62.  So the test case is correct and we shouldn't change the expected value in the test.

I'm not sure why the test is failing on your machine.  You mentioned other tests failing; are they also failing because of the wrong positions for elements in the code?

PhilippWendler:
Given that the difference is 3, and there are 3 line breaks before the literal, I would guess that the issue is related to platform-specific line separators (specifically, with Windows line separators the position would be 3 characters after the position with Linux line separators).

And indeed, the strings given to `writeFile` seem to end up in `ErrorProneInMemoryFileManager.forSourceLines`, which uses [`Files.write`](https://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#write(java.nio.file.Path,%20java.lang.Iterable,%20java.nio.charset.Charset,%20java.nio.file.OpenOption...)), which is documented to concatenate the lines with the current platform's line separator.

So, if @ashwin244 is executing the tests on Windows, this would explain the discrepancy.

ashwin244:
Thanks @PhilippWendler  that seems an appropriate reason, @eaftan no the other test cases were not specific due to positions, test cases of GuardedByBinderTest were failing with java.lang.RuntimeException: java.lang.NoSuchFieldError: RELEASE_9 while calling JavacTaskImpl.analyze() but I guess it supports RELEASE_8.

cushon:
The work-around for the `RELEASE_9` thing is: https://github.com/google/error-prone/blob/master/core/pom.xml#L220. Which probably doesn't work on windows...

eaftan:
Thanks @PhilippWendler, that sounds right.  We only develop on Linux (and occasionally Mac), so it's likely we're missing some stuff for Windows.

@ashwin244 Can you pull from this branch and see if ASTHelpersTest passes on your machine now?
https://github.com/google/error-prone/tree/fix_platform_line_separator_issue

eaftan:
Reopening for the issue with java.lang.NoSuchFieldError: RELEASE_9.

cushon:
We have a clean windows CI build these days, I don't think there's anything to do here.

PR #341, opened 2015-06-04
Fix platform line separator issue

eaftan:
This duplicates the tests in ASTHelpersTest that expect a line position into UNIX and Windows versions that use the proper separators.


cushon:
LGTM

PR #344, opened 2015-06-12
Update Gradle example to use latest plugin release

tbroyer:
Uses fixed versions as `latest.release` no longer works with the Gradle Plugin Portal.
Also uses a fixed version for EP as a “best practice” to make the build reproducible.

Note: changes tested locally, but it was just simpler doing the edit through GitHub.

### DIFF
--- a/examples/gradle/build.gradle
+++ b/examples/gradle/build.gradle
@@ -1,17 +1,21 @@
-apply plugin: 'java'
-
 // See https://github.com/tbroyer/gradle-errorprone-plugin
-apply plugin: 'errorprone'
+buildscript {
+  repositories {
+    maven {
+      url "https://plugins.gradle.org/m2/"
+    }
+  }
+  dependencies {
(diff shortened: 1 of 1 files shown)


cushon:
Thanks. Is the `com.google.errorprone:error_prone_core:+` syntax still an option?

tbroyer:
If you don't include the `configurations.errorprone {}` block, then the plugin will default to the `latest.release` version (which works here, it's only for the plugin itself that it doesn't, or more accurately because of the `https://plugins.gradle.org/m2` Maven repository that doesn't serve/update Maven metadata files).

If you prefer that I leave that part out, tell me and I'll amend the commit to use the default plugin behavior.

cushon:
I was just curious, I don't have a strong opinion. Using the latest version of the plugin is recommended, but make builds reproducible is also good. Pinning the version and remembering to run the gradle equivalent of `versions:use-latest-releases` periodically might be the best of both worlds.

Anyway, I'm happy with the patch as-is.

PR #345, opened 2015-06-25
Ant forked compile fixes

garricko:
Fix a few issues with the ant_fork example. This improves matters, but it still isn't working for me on Java 7u79 because of:

java.lang.NoSuchFieldError: NATIVE_HEADER_OUTPUT

Looks to be similar to #287, so I tried copying some of the NonDelegatingClassLoader stuff from the other adapter into ErrorProneCompiler, but didn't have much luck.

### DIFF
--- a/ant/src/main/java/com/google/errorprone/ErrorProneExternalCompilerAdapter.java
+++ b/ant/src/main/java/com/google/errorprone/ErrorProneExternalCompilerAdapter.java
@@ -101,8 +101,9 @@ public class ErrorProneExternalCompilerAdapter extends DefaultCompilerAdapter {
       cmd.createArgument().setPath(classpath);
       cmd.createArgument().setValue(ErrorProneCompiler.class.getName());
       setupModernJavacCommandlineSwitches(cmd);
+      int firstFile = cmd.size();
       logAndAddFilesToCompile(cmd);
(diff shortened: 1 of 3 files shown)


garricko:
That last commit should also fix the NoSuchFieldError as long as you build with Java 8.

eaftan:
Thanks for the contribution!  I'll take a look at this, but I wanted to be clear that we develop on Linux and Mac and don't test on Windows.  I'll check that this works on Linux, but you'll have to do the same for Windows.

eaftan:
Also, sorry for the slow response.  We were very busy leading up to the end of the quarter, and then we were out for the July 4 holiday.

eaftan:
The first 3 commits here look fine.  I think we can avoid the classloader hacks in 4th (f75bca6) with the changes I made in #358.  Please try that one out and see if it fixes the "java.lang.NoSuchFieldError: NATIVE_HEADER_OUTPUT" error you were seeing.

garricko:
I took the approach from #358 of using the boot classpath, and that seems to work, so I reverted the class loader hacks. I incorporated the other improvements from #358, but avoided the need for separately downloading the javac jar (that stuff already exists in the ant jar) and explicitly setting the executable attribute on the javac task (doesn't seem right to make users hard-code that).

eaftan:
Looks good, I'll merge.  Thanks for the fix!

Issue #280, opened 2015-01-27
Create more annotations like @ParametersAreNonnullByDefault


cushon:
Suggested by @Stephan202 in #93 -

> If possible, I'd like for you to consider another annotation, namely `@EverythingIsNonnullByDefault` as defined in [this SO answer](http://stackoverflow.com/a/13429092/74939) or variations thereof, such as `@ReturnTypesAreNonnullByDefault` and `@FieldsAreNonnullByDefault`, as defined in [this SO answer](http://stackoverflow.com/a/11807961/74939). It would do away with the need for everybody to re-implement/redefine these annotations themselves. (I used these in several internal projects, and based on the SO answers it's not uncommon for other to do, too.)

If error-prone starts checking nullness annotations we should consider adding these.

Issue #44, opened 2014-10-31
Consider supporting JSR 305 annotations


cushon:
_[Original issue](https://code.google.com/p/error-prone/issues/detail?id=44) created by **aftandilian** on 2012-09-20 at 05:44 PM_

---

JSR 305 is abandoned but proposes several new annotations to help static analysis tools.  Presumably anyone using these annotations would like to have a tool check for correctness.  FindBugs and IntelliJ already do this.

http://jsr-305.googlecode.com/svn/trunk/javadoc/javax/annotation/package-summary.html

vorburger:
See also vaguely related https://github.com/google/error-prone/issues/891, which proposes opening up null related annotations.

dfabulich:
FYI, there's an Error Prone plugin, [NullAway](https://github.com/uber/NullAway), that covers the stuff I most care about (`@Nullable` and `@Nonnull`).

https://github.com/uber/NullAway/wiki/Error-Messages

It's a bit of a hassle to hook it up with Bazel; I have a sample repository here that shows how to do it. https://github.com/dfabulich/bazel-checker-framework-bug/

WORKSPACE:

```python
maven_jar(
	name='jsr305',
	artifact='com.google.code.findbugs:jsr305:3.0.2',
)

maven_jar(
	name="nullaway",
	artifact="com.uber.nullaway:nullaway:0.3.2"
)

maven_jar(
	name="guava",
	artifact="com.google.guava:guava:22.0",
)
```

BUILD:

```python
java_library(
	name='x',
	srcs=['X.java'],
	deps=['@jsr305//jar'],
	plugins=['nullaway'],
	javacopts=[
		'-Xep:NullAway:ERROR',
		'-XepOpt:NullAway:AnnotatedPackages=com.example',
	],
)

java_plugin(
	name='nullaway',
	deps=[
		'@nullaway//jar',
		'@guava//jar',
	],
)
```

`com.google.code.findbugs:jsr305` provides the `javax.annotation.Nullable` and `Nonnull` annotation classes. [The dependency on Guava is unfortunate](https://github.com/uber/NullAway/issues/119); I've filed a bug on that.

jbduncan:
Another alternative is the [Checker Framework](https://checkerframework.org/), if you want an exhaustive null-checking type system. (AFAIK, NullAway only checks certain common cases with regards to incorrect `null` usage.)

PR #350, opened 2015-06-27
Implement @Nonnull checks for parameters

JeffreyFalgout:
The impementation also takes into account
@ParametersAreNonnullByDefault type annotations on methods, classes,
and packages.


JeffreyFalgout:
Related to:
#44
#280

PR #370, opened 2015-10-28
Re-enable publishing snapshots and documentation

cushon:
These were only enabled for jdk7 builds, so they haven't been running
since 425b9c167800332da711299e7a7ca8211f728889.


cushon:
This won't actually work until we do something about the doclint.

eaftan:
I disabled doclint in a6b3312d29c417f9afa3c5fe37ac0adca584a005, so I think it should work.

cushon:
Ah, right. Travis agrees with you.

PR #372, opened 2015-11-04
Checks on exception handling code

diy1:
In our recent research paper with the title ["Simple Testing Can Prevent Most Critical Failures"](https://www.usenix.org/system/files/conference/osdi14/osdi14-paper-yuan.pdf), we found that a large number of the most deadly failures in production distributed systems are the result of trivial mistakes in exception handling code. Exception handling code is particularly critical to distributed systems as they're often the last line of defence. We observed three simple, yet deadly, bug patterns:
- "TODO"/"FIXME" in the catch block
- Empty catch block
- Abort in exception over-catch

We implemented a simple checker called Aspirator to detect these three types of mistakes, and it found and fixed hundreds of bugs from Hadoop, HBase, Spark, etc. [HBase developers asked us to implement these checks into error-prone](https://issues.apache.org/jira/browse/HBASE-12187), therefore I'm providing these three checks in this pull request.


diy1:
Signed the CLA just now.

diy1:
added the alternative email i used in commit for the purpose of signing the cla.

eaftan:
Sorry for the slow reply.  I see from the HBASE discussion that you tried to submit these checks a while back.  Sorry we missed it; I don't recall ever seeing the patch.

At a high level, I think these checks look pretty reasonable.  To set expectations for this code review: I'm going to be pretty nitpicky to make sure the code is up to Google code quality standards.  Please don't take this the wrong way; it's the same process that we use internally.  Even the project owners have to go through the same nitpicky code review process.  

Each line I comment on is going to result in an update/email to this issue.  I'm sorry about that but I don't know a better way to do it.

core/src/main/java/com/google/errorprone/bugpatterns/AbortInOvercatch.java
```diff
@@ -0,0 +1,138 @@
...
+ * try {
+ *   // stmt 1 throws exception B
+ *   // stmt 2 throws exception A
+ * } catch (Throwable t) {
+ *   // more than one unique exceptions fall into this catch -- overcatch
+ *   abort(); 
+ * }
+ * </pre>
+ *
+ * While it is a best practice to catch the precise exceptions, abort or System.exit()
```
eaftan: nit: in Javadoc, each paragraph after the first should begin with <p>.  Please also fix this for the other bugcheckers in this PR.

core/src/main/java/com/google/errorprone/bugpatterns/AbortInOvercatch.java
```diff
@@ -0,0 +1,138 @@
...
+import com.sun.tools.javac.code.Symbol;
+import com.sun.tools.javac.code.Symbol.MethodSymbol;
+
+import java.util.List;
+
+/**
+ * A bug checker for abort in exception overcatch. See the following example: 
+ *
+ * <pre>
+ * try {
```
eaftan: On the line after `<pre>`,  please add `{@code`, and on the line just before `</pre>` please add `}`.  See http://stackoverflow.com/questions/541920/multiple-line-code-example-in-javadoc-comment to understand why.  And please do the same for the other code samples in javadoc.

core/src/main/java/com/google/errorprone/bugpatterns/AbortInOvercatch.java
```diff
@@ -0,0 +1,138 @@
...
+ * For more detail, refer to the paper:
+ * "Simple Testing Can Prevent Most Critical Failures: 
+ *  An Analysis of Production Failures in Distributed Data-intensive Systems"
+ *  Yuan et al. Proceedings of the 11th Symposium on Operating Systems Design 
+ *  and Implementation (OSDI), 2014
+ *
+ * @author yuan@eecg.utoronto.ca (Ding Yuan)
+ */
+@BugPattern(name = "AbortInOvercatch",
+    summary = "Aborting the system after catching Throwable or Exception. "
```
eaftan: nit: please remove the period and extra space at the end of this string literal

core/src/main/java/com/google/errorprone/bugpatterns/AbortInOvercatch.java
```diff
@@ -0,0 +1,138 @@
...
+            + "Because these exceptions are high-level exceptions, this might accidentally"
+            + " take down the systems on unexpected exceptions.",
+    explanation = "Aborting the system in Throwable or Exception is bad practice.",
+    category = JDK, maturity = MATURE, severity = ERROR)
```
eaftan: Please set maturity to EXPERIMENTAL and severity to WARNING.

core/src/main/java/com/google/errorprone/bugpatterns/AbortInOvercatch.java
```diff
@@ -0,0 +1,138 @@
...
+  @Override
+  public Description matchTry (TryTree tree, VisitorState state) {
+    List<? extends CatchTree> catchList = tree.getCatches();
+    if (catchList == null || catchList.size() == 0) {
+      // TODO: this try block does not have a catch, we should further check the 
+      // finally block!
+      return Description.NO_MATCH;
+    }
+ 
+    CatchTree lastCatch = catchList.get(tree.getCatches().size() - 1);
```
eaftan: You can use Guava's `Iterables.getLast(catchList)` instead.

core/src/main/java/com/google/errorprone/bugpatterns/EmptyCatch.java
```diff
@@ -0,0 +1,389 @@
...
+  };
+
+  @Override
+  public Description matchTry (TryTree tree, VisitorState state) {
+    if (badEmptyCatchBlock(tree, state)) {
+      // If it has finally block, assume it's OK
+      BlockTree bt = tree.getFinallyBlock();
+      if (bt == null || bt.getStatements().size() == 0) {
+        CatchTree lastCatch = tree.getCatches().get(tree.getCatches().size() - 1);
+        LineMap lineMap = state.getPath().getCompilationUnit().getLineMap();
```
eaftan: It looks like lineMap is unused, so please delete it.

core/src/main/java/com/google/errorprone/bugpatterns/AbortInOvercatch.java
```diff
@@ -0,0 +1,138 @@
...
+    if (overcatch(lastCatch, state)) {
+      if (abortInCatch(lastCatch, state)) {
+        LineMap lineMap = state.getPath().getCompilationUnit().getLineMap();
```
eaftan: It looks like lineMap is unused, so please delete it.

core/src/main/java/com/google/errorprone/bugpatterns/AbortInOvercatch.java
```diff
@@ -0,0 +1,138 @@
...
+
+        return describeMatch(lastCatch);
+      }
+    }
+    return Description.NO_MATCH;
+  }
+
+  private boolean overcatch(CatchTree catchTree, VisitorState state) {    
+    String caughtException = catchTree.getParameter().getType().toString();
```
eaftan: I would not do string matching here, it's too brittle.  You can use the symbol table and the types helper to do real type matching:
```
Types types = state.getTypes();
Symtab symtab = state.getSymtab();
Type paramType = ASTHelpers.getType(catchTree.getParameter().getType());
return (types.isSameType(paramType, symtab.throwableType) || types.isSameType(paramType, symtab.exceptionType));
```

core/src/main/java/com/google/errorprone/bugpatterns/AbortInOvercatch.java
```diff
@@ -0,0 +1,138 @@
...
+  
+  private boolean abortInCatch(CatchTree catchTree, VisitorState state) {
+    List<? extends StatementTree> statements = catchTree.getBlock().getStatements();
+    if (statements.isEmpty()) {
+      return false;
+    }
+    StatementTree lastStmt = statements.get(statements.size() - 1);
+    if (lastStmt.getKind() == EXPRESSION_STATEMENT) {
+      ExpressionTree et = ((ExpressionStatementTree) lastStmt).getExpression();
+
```
eaftan: Here I think you can use our MethodMatchers library to do the same thing in a clearer way.  It would be something like:
```
private static final Matcher<ExpressionTree> ABORT_MATCHER = Matchers.anyOf(
  MethodMatchers.staticMethod.onClass("java.lang.System").named("exit"),
  MethodMatchers.anyMethod().anyClass().withNameMatching(Pattern.compile(".*(?i:abort).*")),
  MethodMatchers.anyMethod().anyClass().withNameMatching(Pattern.compile(".*(?i:shutdown).*")));
```

And then this method can just return ```ABORT_MATCHER.matches(et, state)```.

core/src/test/java/com/google/errorprone/bugpatterns/testdata/AbortInOvercatchPositiveCases.java
```diff
@@ -0,0 +1,46 @@
...
+  public void abortInException() {
+    int a = 0;
+    try {
+      error();
+    } // BUG: Diagnostic contains: 
+    catch (Exception e) {
+      System.exit(1);
+    }
+  }
+}
```
eaftan: Can you please add test cases for abort methods other than `System.exit()`?  Please add at least one each for methods named `abort` and `shutdown`.

core/src/test/java/com/google/errorprone/bugpatterns/AbortInOvercatchTest.java
```diff
@@ -0,0 +1,54 @@
...
+  private CompilationTestHelper compilationHelper;
+
+  @Before
+  public void setUp() {
+    compilationHelper = CompilationTestHelper.newInstance(AbortInOvercatch.class, getClass());
+  }
+
+  @Test
+  public void testPositiveCase() throws Exception {
+ //   compilationHelper.assertCompileFailsWithMessages(
```
eaftan: cleanup: can you please remove commented out lines here and elsewhere in this PR?

core/src/main/java/com/google/errorprone/bugpatterns/EmptyCatch.java
```diff
@@ -0,0 +1,389 @@
...
+ * "Simple Testing Can Prevent Most Critical Failures: 
+ *  An Analysis of Production Failures in Distributed Data-intensive Systems"
+ *  Yuan et al. Proceedings of the 11th USENIX Symposium on Operating Systems Design 
+ *  and Implementation (OSDI), 2014
+ *
+ * @author yuan@eecg.utoronto.ca (Ding Yuan)
+ */
+@BugPattern(name = "EmptyCatch",
+    summary = "Empty catch blocks that ignore exceptions",
+    explanation = "Ignoring exception is a bad practice.",
```
eaftan: In the explanation, you can cite Effective Java, Item 65, "Don't ignore exceptions," and the Google Java Style guide section 6.2 (https://google.github.io/styleguide/javaguide.html#s6.2-caught-exceptions).

core/src/main/java/com/google/errorprone/bugpatterns/EmptyCatch.java
```diff
@@ -0,0 +1,389 @@
...
+    category = JDK, maturity = MATURE, severity = WARNING)
+public class EmptyCatch extends BugChecker implements TryTreeMatcher {
+  /* If one of the following exceptions is ignored, we suppress the warning. */
+  String[] harmlessExceptions = {
```
eaftan: Two things:
1. I would use the fully qualified types here, e.g. "java.io.FileNotFoundException".
2. Please use a set rather than a String array: `ImmutableSet<String> harmlessExceptions = ImmutableSet.of("java.io.FileNotFoundException", "java.lang.ReflectiveOperationException", ...);`
Then you can do the membership check by just doing `harmlessExceptions.contains(typeString);`

core/src/main/java/com/google/errorprone/bugpatterns/EmptyCatch.java
eaftan: lineMap is also unused here, please delete it.

core/src/main/java/com/google/errorprone/bugpatterns/TodoInCatch.java
```diff
@@ -0,0 +1,83 @@
...
+    LineMap lineMap = state.getPath().getCompilationUnit().getLineMap();
+    long startLN = lineMap.getLineNumber(TreeInfo.getStartPos((JCTree) tree));
+    long endLN = lineMap.getLineNumber(TreeInfo.endPos((JCTree) (tree.getBlock())));
+    String filename = state.getPath().getCompilationUnit().getSourceFile().getName();
+    String codeStrs[] = state.getSourceCode().toString().split("\\r?\\n");
+    
+    boolean foundError = false;
+    for (int i = (int) (startLN - 1); i < endLN; i++) {
+      // System.out.println("DEBUG: line " + (i+1) + ": " + codeStrs[i]);
+      if (codeStrs[i].contains("TODO") || codeStrs[i].contains("FIXME")) {
```
eaftan: I think this will issue an error even if the TODO or FIXME is not in a comment, e.g. if I have a constant named FIXME and reference it in the catch.  Maybe the best way to do this would be something like:
```
String catchSource = state.getSourceForNode(catchTree);
if (catchSource.contains("TODO") || catchSource.contains("FIXME")) {
  // Getting the tokens for a node is expensive, so we only do it if we think there is a match here
  for (ErrorProneToken token : state.getTokensForNode(catchTree)) {
    for (Comment comment : token.getComments()) {
      if (comment.getText().contains("TODO") || comment.getText().contains("FIXME")) {
        // return match
      }
    }
  }
  // return no match
}
```

core/src/test/java/com/google/errorprone/bugpatterns/testdata/TODOInCatchNegativeCases.java
```diff
@@ -0,0 +1,36 @@
...
+  public void noTodoInCatch() {
+    int a = 0;
+    try {
+      error();
+    } catch (Throwable t) {
+      // Handled!
+      a++;
+    }
+  }
+}
```
eaftan: Can you add a negative test for when the TODO and FIXME are not in a comment?

diy1:
Ping. Any other things I could do here?

PascalSchumacher:
As a user I think this is a useful enhancement of error-prone.

@eaftan As @diy1 updated the pull request according to your review comments I guess this is good for another review?

eaftan:
I'm super sorry I've been so non-responsive on this.  I'll make sure this happens tomorrow.

eaftan:
We're going to follow what the Guava team does for these more substantial PRs and turn them into internal commits.  That way we can use our own code review and style-checking tools, test them against the codebase, etc.  Once it gets through internal code review, we'll push it back out to our Github repo in a commit that closes this PR.

Thanks again for the contribution, and sorry again for my slow response.

eaftan:
I have a few more questions after pulling the code into an internal changelist.  Please don't modify the code in this PR; I'll take care of that on my end.

core/src/main/java/com/google/errorprone/bugpatterns/EmptyCatch.java
```diff
@@ -0,0 +1,372 @@
...
+            + "[Google Java Style guide section 6.2] "
+            + "(https://google.github.io/styleguide/javaguide.html#s6.2-caught-exceptions); and "
+            + "\"[Simple Testing Can Prevent Most Critical Failures] "
+            + "(http://www.eecg.toronto.edu/~yuan/papers/failure_analysis_osdi14.pdf)\" "
+            + "for more detailed discussions on the harm of swallowing the exceptions. ",
+    category = JDK, maturity = EXPERIMENTAL, severity = WARNING)
+public class EmptyCatch extends BugChecker implements TryTreeMatcher {
+  /* If one of the following exceptions is ignored, we suppress the warning. */
+  ImmutableSet<String> harmlessExceptions = ImmutableSet.of("java.io.FileNotFoundException", 
+    "java.lang.ReflectiveOperationException", "java.lang.ClassNotFoundException",
```
eaftan: What's the rationale behind this set of harmless exceptions?

core/src/main/java/com/google/errorprone/bugpatterns/EmptyCatch.java
```diff
@@ -0,0 +1,372 @@
...
+    "java.lang.reflect.InvocationTargetException", "java.lang.NoSuchFieldException",
+    "java.lang.NoSuchMethodException", "java.lang.InstantiationException");
+  
+  /* If the method is one of the following, we allow it to ignore exceptions. */
+  String[] whitelistedMethods = {
```
eaftan: What's the rationale behind this set of whitelisted method names?

core/src/main/java/com/google/errorprone/bugpatterns/EmptyCatch.java
```diff
@@ -0,0 +1,372 @@
...
+  private boolean badEmptyCatchBlock(TryTree tryTree, VisitorState state) {    
+    List<? extends CatchTree> catchList = tryTree.getCatches();
+    if (catchList == null || catchList.size() == 0) {
+      // TODO: this try block does not have a catch, we should further check the 
+      // finally block!
+      return false;
+    }
+    CatchTree catchTree = catchList.get(catchList.size() - 1);
+
+    /* If there are multiple catch blocks with a try tree, we only check the last catch block. */
```
eaftan: Why only match on the last catch block?  Why not match on all of them?

core/src/main/java/com/google/errorprone/bugpatterns/EmptyCatch.java
diy1: This is a great point. Now thinking about it they should not be necessary. We initially came up with the heuristic b/c of the code snippets like the following (both examples are from HBase):
```java
       try {
         return fs.getFileStatus(path);
       } catch (FileNotFoundException e) {
         // Try another file location
       }
```
Or 
```java
htd = null;
try {
   htd = get(FSUtils.getTableName(d));
 } catch (FileNotFoundException fnfe) {
   // inability of retrieving one HTD shouldn't stop getting the remaining
   LOG.warn("Trouble retrieving htd", fnfe);
 }
if (htd == null) continue;
```
Later we came up with the general idea to suppress the false positives if the exception is handled via divergence in control- or data-flow outside of the catch block (the handledByControlflow and handledByDataflow methods). Now the above cases will suppressed by control-flow and data-flow checks respectively. I checked our evaluation results on the 10 systems and indeed those false positives caused by these exceptions can all be covered by handledByControlflow and handledByDataflow. Therefore these harmlessExceptions should no longer be useful and they probably will cause false negatives.

core/src/main/java/com/google/errorprone/bugpatterns/EmptyCatch.java
diy1: These methods are related with system shutdown or resource disposal. We observe it's quite common that exceptions from these methods are ignored and it is unclear whether anything meaningful can be done to handle these exceptions.

core/src/main/java/com/google/errorprone/bugpatterns/EmptyCatch.java
diy1: Good point. I think we should match all of them instead of the last one (we had this heuristic at the beginning simply to reduce the number of warnings aspirator produces).

core/src/main/java/com/google/errorprone/bugpatterns/EmptyCatch.java
eaftan: Thanks, I'll remove them.

core/src/main/java/com/google/errorprone/bugpatterns/EmptyCatch.java
eaftan: Acknowledged.  Added this as a comment on the whitelist.

core/src/main/java/com/google/errorprone/bugpatterns/EmptyCatch.java
eaftan: OK, I'll change it to match on all catch blocks.

diy1:
Hello @eaftan , how is your internal code review on this going?

mattf-horton:
@eaftan , would you please consider finishing this PR?  It would be very useful in the Hadoop community.  Thanks.

diy1:
Hi there, any chance to finish this PR? thx.

cushon:
For empty catch blocks, I think we may want to only discourage the case where the catch block is completely empty (i.e. it doesn't even have a comment): https://google.github.io/styleguide/javaguide.html#s6.2-caught-exceptions.

For `AbortInOvercatch`, most of the examples i saw in our code base that looked potentially bug-prone involved `System.exit`, and there's already a check to report uses of `System.exit` outside of the `main` method (including in catch blocks): http://errorprone.info/bugpattern/SystemExitOutsideMain

I also looked at examples of `TODO`s, and I'm not sure we'd want to discourage them more strongly in catch blocks than in general. That one might be a better fit for a project-specific check, e.g. with a http://errorprone.info/docs/plugins.

cushon:
> For empty catch blocks, I think we may want to only discourage the case where the catch block is completely empty (i.e. it doesn't even have a comment):

`EmptyCatch` was added with that modification in https://github.com/google/error-prone/commit/f067128733b3b60a2339d7ffe0688f2e7a6819d6.

PR #373, opened 2015-11-07
Delete the DepAnn bug pattern.

swankjesse:
This bug pattern is harmful. It encourages developers to repeat themselves with
both a helpful @deprecated Javadoc comment (that typically provides useful
information) and a useless @Deprecated annotation (that serves no practical
purpose, and isn't worth the redundancy).


eaftan:
We need this check because tools are not guaranteed to respect the `@deprecated` javadoc, only the `@Deprecated` annotation.  See http://docs.oracle.com/javase/7/docs/technotes/guides/javadoc/deprecation/deprecation.html#javadoc_tag.  Specifically, [Kythe](https://github.com/google/kythe) does not handle the `@deprecated` javadoc.

You can turn it off with an "-Xep:DepAnn:OFF" command line option.

swankjesse:
@eaftan can we just fix Kythe?

Stephan202:
If you do decide to drop this bug pattern, can't it be turned off by default, rather than be deleted outright?

swankjesse:
@Stephan202 just curious, why would somebody want to turn this on?

Stephan202:
@swankjesse, I see `@Deprecated` as similar to annotations such as `@Nullable`: meta-information about the code made available to both users of the code and static analysis tools. I believe that using a "proper" annotation for this is more appropriate because (a) the compiler guards against typos and (b) I'd wager that there are more static analysis tools out there that handle "proper" annotations than that handle both types of annotation.

I do see your argument about duplication, but I do not feel it to be a problem (especially because deprecated code represents a tiny minority of the code I interact with on a daily basis).

I suppose this position is not uncommon, judging by the existence of the `DepAnn` bug pattern we're currently discussing, as well as the `MissingDeprecatedCheck` provided by both [Checkstyle](https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/checks/annotation/MissingDeprecatedCheck.java) and [Sonar Squid](https://github.com/SonarSource/sonar-java/blob/HEAD/java-checks/src/main/java/org/sonar/java/checks/MissingDeprecatedCheck.java).

JakeWharton:
Consider also [JEP 277](http://openjdk.java.net/jeps/277) which not only will add value to the annotation, but also makes commentary on the Javadoc tag (emphasis mine):

> It has been suggested that the `@deprecated` Javadoc tag be retired in favor of the `@Deprecated` annotation. They are redundant, in a narrow sense, in that the `@deprecated` Javadoc tag and the `@Deprecated` annotation **should always both be present or absent**; it never makes sense to have one without the other. However, they provide different information to different consumers. The `@deprecated` Javadoc tag provides descriptive text to be included in documentation. The `@Deprecated` annotation provides fine-grained information to tooling and to the runtime. Migrating a documentation string to, say, an annotation value would clutter up the runtime image with documentation strings, and it would require additional work for Javadoc to extract the string from the annotation instead of the doc comment where it currently resides.

swankjesse:
The `@Deprecated` annotation is a mistake. Just like `public`, `static` and `final`, deprecated information is already present in both source and `.class` files for tools to consume as they like. We had `@deprecated` long before we had annotations, and IDEs can handle either. Probably the best long term fix is to deprecate the `@Deprecated` annotation.

cushon:
This check reimplements javac's dep-ann xlint warning. It was duplicated here because we wanted an automated fix.

[JDK-4903501](https://bugs.openjdk.java.net/browse/JDK-4903501) says:

> We propose to add a marker annotation java.lang.Deprecated to supplant and
eventually replace the javadoc tag `@deprecated`. This is exactly the sort
of thing metadata is best at expression. This may improve the performance
of compilers, as eventually they will no longer have to scan doc comments
for `@deprecated` tags.

> As a transitional feature, in -source 1.5 javac will give a warning if
a program element is deprecated in the javadoc tag but not with an annotation.
In 1.6 we might either make this a hard error or simply stop recognizing the
tag in doc comments altogether.

It sounds like the plans to get rid of the javadoc tag have been abandoned, but the annotation isn't going anywhere either. If you want metadata about deprecation in class files, prefer the annotation. Compilers shouldn't have to parse javadoc.

swankjesse:
So... continue to require both Javadoc and the annotation forever, because it makes compilation faster / makes it easier to write compilers? I'm not buying it, but it's your project.

Issue #378, opened 2015-12-20
SuggestedFix `importsToRemove` does not work


yigit:
Not sure if I'm reading this API wrong but I'm assuming that

`suggestedFix.removeImport("com.foo.bar")` should remove the import statement for `com.foo.bar`.

Seems like the problem is in `DescriptionBasedDiff` calling the wrong method (and its tests don't cover removeImport).

I already have a fix for this, will upload soon.

PR #379, opened 2015-12-20
Fix removeImport suggestion.

yigit:
This CL fixes a bug in DescriptionBasedDiff where it does not populate
importsToRemove list properly. Also added a test to cover that case.

Issue: #378


cushon:
Looks good, thanks for the fix.

cushon:
Merged as 1d461d282debedb25830b004909dc0031924bb70

PR #400, opened 2016-03-18
Update to new Version of Plexus Compiler

ReneTC:


cushon:
Thanks! We should use the new version everywhere, would you mind also updating the main `pom.xml` and the installation docs?

https://github.com/google/error-prone/blob/gh-pages/docs/installation.md
https://github.com/google/error-prone/blob/master/pom.xml

PR #413, opened 2016-05-24
Add a checker for SLF4J: all org.slf4j.Logger field should be private

KengoTODA:
Hello ALL,
Now I'm using Findbugs to detect potential problem in my usage of SLF4J. Here is [my Findbugs plugin](https://github.com/KengoTODA/findbugs-slf4j/) to check.

I want to use this error-prone to detect problems, to improve agility to develop and to provide auto-fix feature. This PR adds a checker which checks visibility of SLF4J logger field, could you review?
### Questions
- [ ] Can I write unit test to test auto-fix feature?
- [x] Is it better to move `ModifierComparator` to other package? It doesn't related with SLF4J and can be used by other checkers.
- [ ] Is it better to propose new class `FieldVisibility` like `MethodVisibility`? Then I will extract `MethodVisibility.Visibility`.


KengoTODA:
I'me merged latest `master` branch to mine, please check. Thanks!

core/src/main/java/com/google/errorprone/bugpatterns/slf4j/ModifierComparator.java
```diff
@@ -0,0 +1,48 @@
...
+        default:
+            throw new IllegalArgumentException("Unknown modifier:" + input);
+        }
+    }
+
+    @Override
+    public int compare(Modifier left, Modifier right) {
+        return toOrder(left) - toOrder(right);
+    }
+}
```
Stephan202: This class can be replaced with an instance of `com.google.common.collectOrdering.explicit`.

KengoTODA: Thanks, I will check.
https://google.github.io/guava/releases/19.0/api/docs/com/google/common/collect/Ordering.html#explicit(T, T...)

KengoTODA: Updated. bceff13

core/src/main/java/com/google/errorprone/bugpatterns/slf4j/DoNotPublishSlf4jLogger.java
```diff
@@ -0,0 +1,116 @@
...
+            Modifier.PUBLIC, Modifier.PROTECTED, Modifier.PRIVATE, Modifier.ABSTRACT,
+            Modifier.STATIC, Modifier.FINAL, Modifier.TRANSIENT, Modifier.VOLATILE,
+            Modifier.DEFAULT, Modifier.SYNCHRONIZED, Modifier.NATIVE, Modifier.STRICTFP);
+
+    private final String lineSeparator = System.lineSeparator();
+
+    @Override
+    public Description matchVariable(VariableTree tree, VisitorState state) {
+        if (allOf(isField(), SLF4J_LOGGER, not(PRIVATE)).matches(tree, state)) {
+            Fix fix = createSuggestedFix(tree.getModifiers());
```
jbuberel: error: [DeadException] Exception created but not thrown
    new RuntimeException();
    ^
    (see http://errorprone.info/bugpattern/DeadException)
  Did you mean 'throw new RuntimeException();'?

[Apply Fix](https://google.com/) - [Not Useful](https://google.com/)

KengoTODA:
this checker has been moved to plugin, so I'll close this PR.

https://github.com/KengoTODA/errorprone-slf4j

PR #419, opened 2016-06-24
Add Gradle example of plugin check

tbroyer:


cushon:
Thanks!

examples/plugin/gradle/build.gradle
```diff
@@ -0,0 +1,30 @@
...
+    classpath "net.ltgt.gradle:gradle-errorprone-plugin:0.0.8"
+    classpath "net.ltgt.gradle:gradle-apt-plugin:0.6"
+  }
+}
+
+subprojects {
+  repositories {
+    mavenCentral()
+    maven {
+      url "https://oss.sonatype.org/content/repositories/snapshots/"
```
tbroyer: Note that those three lines can be removed when referencing a non-snapshot, leaving only the `mavenCentral()` repository.

PR #435, opened 2016-07-16
Allow custom CheckReturnValue annotations

### LINKED ISSUE #422: CheckReturnValue should allow custom annotations
Currently `CheckReturnValue` only works with `javax.annotation.CheckReturnValue`.  It should also allow for custom annotations with the same name to warn about APIs like assertj which have their own annotation joel-costigliola/assertj-core#695.

andrewgaul:
Previously the checker only used the JSR 305 annotation, not checking
custom annotations like assertj uses.  Fixes #422.


cushon:
This was merged as 6c440ae18e9df017485279068cd62d8c65407094

PR #449, opened 2016-08-17
Enable OverridingJavaxInjectMethod by default, a check to find method…

ronshapiro:
…s that are not annotated with @Inject, but are overriding methods that have @javax.inject.Inject. JSR-330 injectors won't trigger these methods in the subclasses, leading to potential initialization issues.

-------------
Created by MOE: https://github.com/google/moe
MOE_MIGRATED_REVID=130536934


eaftan:
Nice timing -- @nglorioso just enabled this internally.  He's working on the hodgepodge of dependency injection checks we have right now.  

We can drop this PR, and this check will be enabled the next time we sync outward.

ronshapiro:
I should have been more clear - this was actually me running MOE, but I don't have access to the error-prone project

eaftan:
We talked offline, and we decided that someone on Error Prone team will do these MOE syncs and commit them directly.  Closing this PR.

PR #450, opened 2016-08-21
Provide example how to override multiple flags in maven pom.xml

turbanoff:


eaftan:
Thanks for the contribution!

PR #457, opened 2016-09-07
Support suppressions on MockitoCast detector

writeoncereadmany:
MockitoCast detector is currently unsuppressable, because it's a compilation unit matcher and the viable attachment points for suppression (class/method level) are children of a compilation unit. I've added two unit tests which demonstrate the issue.

As far as I can see there's no need to it to be a compilation unit matcher: on the contrary, we're looking specifically for invocations of a specific method, and we have access to all the state we need. Trivial modifications to make it a MethodInvocationTreeMatcher give us a detector which passes all the existing unit tests and also the additional unit tests.


writeoncereadmany:
I signed it!

PR #458, opened 2016-09-07
Support suppressions on MockitoCast detector

writeoncereadmany:
The MockitoCast detector can't be suppressed: it's a compilation unit matcher, which means we haven't visited the code which could suppress the detector before we start analysing it. See the added unit tests to demonstrate the problem.

This is easily rectified by making it a MethodInvocationTreeMatcher and putting all the logic in matchInvocation(). I can't see any reason why it needs to act at the compilation unit level (and nor can its unit tests - they all continue to pass with this change).

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/MockitoCast.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/MockitoCast.java
@@ -16,28 +16,16 @@
 
 package com.google.errorprone.bugpatterns;
 
-import static com.google.common.base.MoreObjects.firstNonNull;
-import static com.google.errorprone.BugPattern.Category.MOCKITO;
-import static com.google.errorprone.BugPattern.MaturityLevel.MATURE;
-import static com.google.errorprone.BugPattern.SeverityLevel.ERROR;
-import static com.google.errorprone.matchers.method.MethodMatchers.staticMethod;
-
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 import com.google.errorprone.BugPattern;
 import com.google.errorprone.VisitorState;
-import com.google.errorprone.bugpatterns.BugChecker.CompilationUnitTreeMatcher;
 import com.google.errorprone.fixes.SuggestedFix;
 import com.google.errorprone.fixes.SuggestedFixes;
 import com.google.errorprone.matchers.Description;
 import com.google.errorprone.matchers.Matcher;
 import com.google.errorprone.util.ASTHelpers;
-import com.sun.source.tree.AssignmentTree;
-import com.sun.source.tree.CompilationUnitTree;
-import com.sun.source.tree.ExpressionTree;
-import com.sun.source.tree.MethodInvocationTree;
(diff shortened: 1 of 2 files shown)


cushon:
> I can't see any reason why it needs to act at the compilation unit level

It scans the entire file and records the answer strategy used to initialize mock variables, and then uses that information when analyzing calls to `when`. This change causes it to scan the entire compilation unit once for each invocation of `when`.

We can still make it respect `@SuppressWarnings`, but if you don't want to run this check I'm curious why you'd use suppression annotations instead of just disabling it on the command line?

writeoncereadmany:
Okay, thanks, that makes sense. If there's a more performant way of achieving the same goal, then great. 

Reasons why I think it's important it respects @SuppressWarnings:

Firstly, it's surprising that it isn't, especially as it's documented as doing so. I was trialling out error-prone on a new project and this was the only failing check and for a while I thought suppression was broken in general, as opposed to just this check. I had to read and debug the source code in order to understand what was going on here, and that's an unreasonable burden for most users.

Secondly, I don't think we want the warning disabled on the command line. One of two things will happen first: we'll upgrade to a JDK such that this is a bug, or we'll upgrade Mockito such that it isn't. 

In the former case, if we're suppressing in place then we'll see our tests fail with the ClassCastException, see that it's co-located with the suppression, remove the suppression and fix the code - and any future violations will also get picked up, which is what we want. 

If we're suppressing on the command line, nothing will happen to draw attention to the suppression, so it will just stay there as residual cruft and we'll miss out on a useful warning. Experience shows that changes to the build tend to be purely additive.

In the case where Mockito is upgraded to it's no longer an issue - that's the appropriate time to suppress on the command line, but it's not dangerous, just a bit crufty, to suppress in code.

Thirdly, it's a matter of what sorts of workflow I want to encourage if I introduce it as a new tool for the team. Disabling checks at the command line *in general* is a nuclear option and not usually appropriate: it's situationally appropriate here only because it applies to a narrow intersection of JDK and library versions, so I want to go by a mantra of fix-or-suppress and discourage disabling.

writeoncereadmany:
I've modified the check to collect the bad answer strategies on a per-compilation-unit basis, but only check for violations on a per-method-invocation basis. This means we don't do the rework, and the usual suppression mechanisms fall out as expected.

This does mean that the checker has become stateful, which slightly offends my sensibilities, but as far as I can see the alternatives are:
* Doing everything at the compilation unit level and not respecting suppressions
* Doing everything at the compilation unit level and re-implementing suppression logic
* Doing everything at the method invocation level and re-calculating answer strategies on demand
* Having two different entry points and communicating between them with shared state

Of those options, I'm favouring multiple entry points communicating with shared state.

cushon:
The mockito issue this check helped mitigate has been closed (https://github.com/mockito/mockito/issues/357), and the javac versions affected by the change are getting older, so I think it might be time to get rid of MockitoCast.

cushon:
> I think it might be time to get rid of MockitoCast.

I'm going to proceed with that, the change should show up on github in the next day or so.

PR #478, opened 2016-11-15
Whitelist many java.time classes as well known immutable types

### LINKED ISSUE #477: recognize java.time types as immutable
I see you have a [whitelist](https://github.com/google/error-prone/blob/b47de27e8e9d577d498ca1366dd1eb62d83752ef/core/src/main/java/com/google/errorprone/bugpatterns/threadsafety/WellKnownMutability.java) for known-immutable types recognized by the `@Immutable` annotation / bug pattern.  Would you consider adding the immutable `java.time` types to that list?

I have lots of types that should be `@Immutable`, but maybe half of them have a `LocalDateTime` or something similar that trips up error prone.  I could add a bunch of `@SuppressWarnings` but it would be cool if I didn't have to.

michaelhixson:
This list does not include:

 * Any enums, even though all the enums in `java.time` and subpackages are immutable, because enums are assumed to be immutable already.

 * Any interfaces, even when the contract of the interface declares that all concrete implementations must be immutable (e.g. `ChronoLocalDate`).

 * Any singletons, even when those singletons are immutable (e.g. `IsoChronology`).

Fixes #477


michaelhixson:
poke @googlebot

cushon:
Thanks! This looks good.

Is there any reason not to whitelist interfaces and singletons? Adding interfaces to the whitelist will cause the check to enforce subtypes are also immutable, the same as if the interface was explicitly annotated.

michaelhixson:
I didn't realize that error prone would interact with `@Immutable` interfaces that way.  That sounds perfect.  I'll go ahead and add those interfaces to the list.

Out of curiosity, do you consider [`java.nio.file.Path`](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Path.html) to be in the same boat?

I didn't add the singletons to the list because I figured that zero people would add a field of that concrete type to one of their own classes -- what would be the point?  But since they are implementations of must-be-immutable interfaces, and I'm going to add those interfaces to the list, I'm guessing that forces me to add the singleton types as well.  So I will.

michaelhixson:
I've updated the PR to include those interfaces and singletons.

cushon:
> I didn't add the singletons to the list because I figured that zero people would add a field of that concrete type to one of their own classes -- what would be the point? But since they are implementations of must-be-immutable interfaces, and I'm going to add those interfaces to the list, I'm guessing that forces me to add the singleton types as well. So I will.

The singletons shouldn't be required unless you're compiling the JDK with Error Prone, instead of just reading those class from the bootclasspath. But I think we may as well include them.

And I think `Path` should also be whitelisted. The only issue is that there are more existing subtypes that will now be checked for immutability.

Issue #482, opened 2016-11-23
net.ltgt.errorprone 0.0.8 not compatible with dagger 2.8


koral--:
Not sure if it is bug in errorprone or android/errorprone gradle plugin but first reporting here. Combination mentioned in title causes `NoSuchMethodError: com.google.common.base.Preconditions.checkState` because `errorprone` configuration still uses guava 19.0.

More details: https://github.com/google/dagger/issues/511#issuecomment-262435064

tbroyer:
Note: the `net.ltgt.errorprone` gradle plugin issue tracker is at https://github.com/tbroyer/gradle-errorprone-plugin; yet I don't think this is a bug with the plugin specifically, but rather with ErrorProne itself; possibly due to [this code in JavaC](https://github.com/google/error-prone-javac/blob/90149b22e202e7b24bfafc80df7509c8fb490ee7/src/jdk.compiler/share/classes/com/sun/tools/javac/util/BaseFileManager.java#L133) (called via [this code](https://github.com/google/error-prone-javac/blob/90149b22e202e7b24bfafc80df7509c8fb490ee7/src/jdk.compiler/share/classes/com/sun/tools/javac/processing/JavacProcessingEnvironment.java#L232)), therefore leaking ErrorProne's own class-loader to the annotation processors' class-loader.
Should ErrorProne set the `procloader` option when delegating to JavaC?

cushon:
Error Prone should already be installing a processor classloader that avoids leaking the internal version of guava: https://github.com/google/error-prone/blob/390d90ea894e8270f1c9ddc09df20a01feb1647a/check_api/src/main/java/com/google/errorprone/BaseErrorProneCompiler.java#L176

Either that's broken, or there's something else going on here.

koral--:
Here is example of project which reproduces this issue: https://github.com/koral--/GithubClient
To reproduce run eg. `./gradlew assembleDebug`.

tbroyer:
[...]

PR #504, opened 2017-01-06
Unshade dependencies in error_prone_core

### LINKED ISSUE #501: error prone artifact provides guava library
Is this expected? Bazel and Buck tool chains include their own versions of guava on their classpath (Bazel 21.0-rc1 and Buck 19.0) and after consuming the error prone there is yet another guava version (20 in error prone 2.0.15), so that for example `buck autodeps` command is failing after adding error prone as native compiler option in Buck, with: [1].

```
WARNING: Multiple providers for com.google.common.base.Preconditions: //third-party/java/errorprone:error_prone, //third-party/java/guava:guava. Consider adding entry to .buckconfig to eliminate ambiguity:
[autodeps]
java-package-mappings = com.google.common.base.Preconditions => //third-party/java/errorprone:error_prone
```

[1] http://paste.openstack.org/show/593952

tbroyer:
error_prone_core had some dependencies shaded that caused problems
with build-system integrations. Moreover, they were still referenced
them in its POM, leading to unnecessary duplications in the classpath.

This commit changes the default error_prone_core artifact to no
longer use shaded dependencies, and moves the shaded artifact to
an additional JAR with 'with-dependencies' classifier. This JAR
is needed to keep the error_prone_refaster JAR identical.

An alternative would be to move the list of the dependencies we want to shade from `core` to `refaster`; that means splitting the information into 2 files, and the shade configuration would have to be duplicated somehow if a new module is added that needs the same shading as `refaster` (that being said, I'm not familiar with Refaster and maybe the shading isn't even necessary/useful)

Fixes #501
Updates #482

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -283,6 +283,8 @@
               <goal>shade</goal>
             </goals>
             <configuration>
+              <shadedArtifactAttached>true</shadedArtifactAttached>
+              <shadedClassifierName>with-dependencies</shadedClassifierName>
               <createDependencyReducedPom>false</createDependencyReducedPom>
               <artifactSet>
                 <!-- Include only dependencies with compatible licenses.
--- a/refaster/pom.xml
+++ b/refaster/pom.xml
@@ -33,6 +33,7 @@
             <groupId>com.google.errorprone</groupId>
             <artifactId>error_prone_core</artifactId>
             <version>${project.version}</version>
+            <classifier>with-dependencies</classifier>
             <scope>compile</scope>
         </dependency>
         <dependency>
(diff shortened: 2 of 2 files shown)


cushon:
This looks fine to me, thanks. The shading was added for the ant integration in 0a19fc39218f0845d5b1b86a40a1eb56dfff6531, I'm not sure why core was ever shaded. Maybe @eaftan remembers?

And @nglorioso for the refaster implications.

eaftan:
I think we shaded core mostly so you could use Error Prone as a command-line, drop in replacement for javac.  At the time we didn't package javac, so everything was in one jar.  Now we probably don't need to shade it.

PR #518, opened 2017-01-24
Add documents to gh-pages to describe Refaster/Patching

nglorioso:


docs/refaster.md
```diff
@@ -0,0 +1,83 @@
...
+
+  @AfterTemplate
+  @AlsoNegation
+  boolean optimizedMethod(String string) {
+    return string.isEmpty();
+  }
+}
+```
+
+Refaster templates are any class with multiple methods with the same return type and list of arguments. One of the methods should be annotated `@AfterTemplate`, and every other method should be annotated with `@BeforeTemplate`. With this template, any code calling String#equals passing in the empty string literal, or calling String#length and comparing it to 0 will be replaced by a call to String#isEmpty. Notably, no matter how the String expression is generated, Refaster will do the replacement:
```
cushon: nit: backticks around e.g. String#equals, and @BeforeTemplate lower down?

docs/refaster.md
```diff
@@ -0,0 +1,83 @@
+---
+title: Refaster templates
+layout: documentation
+---
+
+In addition to [patching] your code using the checks built-in to Error Prone, we've developed a mechanism to refactor your code using before-and-after templates (we call them Refaster templates). Once you write these templates, you compile them into .refaster files, then use the Error Prone compiler to refactor your code according to those rules.
+
+## Building Refaster Templates
+
+Writing refaster rules is best done by example:
```
cushon: This sentence is a little off, should it be something like "Explaining how to write refaster rules is best done by example"?

docs/refaster.md
```diff
@@ -0,0 +1,83 @@
+---
+title: Refaster templates
+layout: documentation
+---
+
+In addition to [patching] your code using the checks built-in to Error Prone, we've developed a mechanism to refactor your code using before-and-after templates (we call them Refaster templates). Once you write these templates, you compile them into .refaster files, then use the Error Prone compiler to refactor your code according to those rules.
```
cushon: Can you add a link to: https://research.google.com/pubs/archive/41876.pdf

nglorioso: Done.

nglorioso:
Pushing these for now. Will followup after looking at @lowasser's internal docs and borrowing liberally from them.

PR #531, opened 2017-02-16
Fix active navigation tab on index and logo link.

JakeWharton:
Before on `index.html`, note the activated tab:

screen shot 2017-02-16 at 6 50 13 pm

After:

screen shot 2017-02-16 at 6 50 36 pm

----

Before on any page, note the link preview at the bottom while hovered on the logo:

screen shot 2017-02-16 at 6 50 26 pm

After:

screen shot 2017-02-16 at 6 50 42 pm

PR #538, opened 2017-02-18
Update pattern index to use correct template.

JakeWharton:
`master` doesn't select the "Bug Patterns" tab whereas `bugpatterns` does.


JakeWharton:
I had manually fixed this in #531 but then it was reverted by docgen in https://github.com/google/error-prone/commit/b25de8954b7b9e681dca6877e8b035b83f9b2e7b#diff-e5cd5138b89a0b12f3cfc2c51e22ac70.

cushon:
Thanks, merged as dfa5b488e4660731244752b4952bbc9d53849fc9.

Issue #497, opened 2016-12-24
NPE upon unused import in `package-info.java`


Stephan202:
Consider a `package-info.java` containing the following:
```java
package com.example;

// The actual type imported is irrelevant.
import java.util.Map;
```

Then when the `RemoveUnusedImports` bug pattern is enabled, the following exception is thrown:
```
/some/path/to/com/example/package-info.java:1: error: An unhandled exception was thrown by the Error Prone static analysis plugin.
package com.example;
^
     Please report this at https://github.com/google/error-prone/issues/new and include the following:
  
     error-prone version: 2.0.15
     Stack Trace:
     java.lang.NullPointerException
  	at com.google.errorprone.fixes.AppliedFix$Applier.apply(AppliedFix.java:96)
  	at com.google.errorprone.JavacErrorDescriptionListener.lambda$new$0(JavacErrorDescriptionListener.java:61)
  	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
  	at java.util.Iterator.forEachRemaining(Iterator.java:116)
  	at java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1801)
  	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
  	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
  	at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
  	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
  	at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:499)
  	at com.google.errorprone.JavacErrorDescriptionListener.onDescribed(JavacErrorDescriptionListener.java:75)
  	at com.google.errorprone.VisitorState.reportMatch(VisitorState.java:159)
  	at com.google.errorprone.scanner.Scanner.reportMatch(Scanner.java:160)
  	at com.google.errorprone.scanner.ErrorProneScanner.visitCompilationUnit(ErrorProneScanner.java:644)
  	at com.google.errorprone.scanner.ErrorProneScanner.visitCompilationUnit(ErrorProneScanner.java:144)
  	at com.sun.tools.javac.tree.JCTree$JCCompilationUnit.accept(JCTree.java:561)
  	at com.sun.source.util.TreePathScanner.scan(TreePathScanner.java:53)
[...]

PR #543, opened 2017-02-18
Fix for #497: support deletion of last line of file

Stephan202:
This PR proposes a fix for issue #497. The problem turns out not to be specific to `package-info.java`, though perhaps cannot reasonably be triggered in other contexts.

Some notes about the tests I added:
1. The test added to `AppliedFixTest` matches the call made by `RemoveUnusedImports` when the issue is triggered. I added this test because `AppliedFix` is the class that is actually being fixed.
2. The test added to `RemoveUnusedImportsTest` uses text-matching rather than AST-matching because the latter simply isn't supported for `package-info.java` files, it seems. (This appears to be a limitation of the `compile-testing` library.)
3. The test added to `SuggestedFixesTest` is arguably redundant, and a bit out of place, since it's the only test in that class which references a "real"/non-test bug pattern. If you want me to delete it, let me know. On the other hand, it does test the "Did you mean to remove this line" phrasing, which is otherwise only tested in `NonOverridingEqualsTest`, which arguably should not be the primary/only place to test this.

### DIFF
--- a/check_api/src/main/java/com/google/errorprone/fixes/AppliedFix.java
+++ b/check_api/src/main/java/com/google/errorprone/fixes/AppliedFix.java
@@ -93,10 +93,16 @@ public class AppliedFix {
           lineNumberReader.readLine();
         }
         // TODO: this is over-simplified; need a failing test case
-        snippet = lineNumberReader.readLine().trim();
-        // snip comment from line
-        if (snippet.contains("//")) {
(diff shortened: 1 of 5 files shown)


Stephan202:
I signed it!

Stephan202:
Thanks for accepting the contribution :).

PR #561, opened 2017-03-06
Bump URLs in refaster to 2.0.18 concrete versions

nglorioso:


cushon:
2.0.18?

nglorioso:
I don't see jars in central yet :)

nglorioso:
Looks like the directory index is broken, but the jars are there. Updating to 2.0.18 instead

PR #580, opened 2017-03-17
Reduce serverity of LiteralClassName to SUGGESTION

### LINKED ISSUE #579: Lower severity of LiteralClassName to SUGGESTION
Consider the following pattern:
```
Class<?> clz = null;
try {
  clz = Class.forName("SomethingOptional");
} catch (ClassNotFoundException e) {
  // log that the class was not found
}
```
ErrorProne would flag that this should be warned about, even though it may depend on runtime dependencies.  There aren't any [examples of this being dangerous](http://errorprone.info/bugpattern/LiteralClassName) and hardly any explanation as to how this could lead to an error happening.

Can this be lowered in severity?

carl-mastrangelo:
Fixes #579


nglorioso:
Ah, sorry for letting this sit for so long. This change looks good

PR #594, opened 2017-04-03
check When value for CheckReturnValue annotation

ellisjoe:


ellisjoe:
I signed it!

ellisjoe:
based on the failing test `noCRVonClasspath` it seems like this may not be a desired feature since this requires having `jsr305` available at compile time. Would be great to hear if this is worth pursuing more.

eaftan:
What's the use case for setting the `when` property of `@CheckReturnValue` to anything other than `ALWAYS`?

ellisjoe:
we have ping endpoints that looks like:
```
@CheckReturnValue(when = When.NEVER)
String ping();
```

cushon:
We created a [`@CanIgnoreReturnValue`](http://errorprone.info/api/latest/com/google/errorprone/annotations/CanIgnoreReturnValue.html) annotation that has the same behaviour. Maybe we weren't aware of `when` at the time?

ellisjoe:
how do you feel about supporting both `@CanIgnoreReturnValue` and `when`?

eaftan:
Do other tools respect the when property?  If so, which ones?

`@CheckReturnValue(when = When.NEVER)` seems to say that you should *never* consume the result of the call, which is weird.

cushon:
I think we're going to stick with `@CanIgnoreReturnValue` for now. The Error Prone `@CRV` annotation doesn't support `when`, and the JSR-305 annotations have an [uncertain future](https://github.com/google/guava/issues/2960).

PR #598, opened 2017-04-11
Make error prone compiler return a JavacTask

kageiit:
Build systems like buck rely on the `CompilatonTask` to be a `JavacTask`. Wired up the returned interface to the real implementation.

See https://github.com/facebook/buck/issues/1285


cushon:
I think `BaseErrorProneJavaCompiler` can be refactored to avoid subclassing `CompilationTask`, which should also resolve #608.

kageiit:
The reason we are still using the `BaseErrorProneJavaCompiler` directly instead of using it as a plugin is because of #609 . If refactoring support is part of the plugin, then a bunch of these issues could also be avoided potentially.

Issue #511, opened 2017-01-15
Feature Request: Skip certain files


vanniktech:
There are some auto generated files that should not be processed by Error Prone. So far I haven't found a way to tell Error Prone that. So it would be nice to be able to do that.

One example is the `R` class in android.

vanniktech:
Closing in favor of #648 which has more information and ideas.

mernst:
For others who stumble across this feature request, Error Prone now has a [`-XepExcludedPaths` command-line flag](https://errorprone.info/docs/flags).

PR #599, opened 2017-04-11
Add an option to exclude paths from being analyzed

### LINKED ISSUE #523: Disable checks for an entire package.
I'm adding error-prone to a legacy project and have the need to disable/ignore an entire package. 

This package is a transplant from a third-party dependency that isn't published as a consumable artifact but also requires a very small tweak to work in my usage. Since I want to reduce the changes to this package so future updates are easy, I'm hesitant to go in and annotation every type with a suppression for every violation therein. 

Unfortunately the `@SuppressWarnings` annotation can't be targeted at packages so I can't use it in a `package-info.java` either (not sure this would have worked anyway).

Is there any other way of suppressing checks for an entire package? And if not, what do you think about introducing a mechanism for this (e.g., an exclude flag which was a list of globs)?

kageiit:
This adds an option to suppress some paths entirely from being analyzed by the error prone analyzer. This is particularly useful for generated code or third party code which consumers may not want to analyze.

Fixes #523

The input patterns are treated as regular expressions for path matching. For example:
`-XepExcludedPaths:".*/build/generated/.*\.java"` will exclude all java files inside any `build/generated/` directories

### DIFF
--- a/check_api/src/main/java/com/google/errorprone/ErrorProneAnalyzer.java
+++ b/check_api/src/main/java/com/google/errorprone/ErrorProneAnalyzer.java
@@ -138,17 +138,15 @@ public class ErrorProneAnalyzer implements TaskListener {
     DescriptionListener descriptionListener =
         descriptionListenerFactory.getDescriptionListener(log, compilation);
     try {
-      if (path.getLeaf().getKind() == Tree.Kind.COMPILATION_UNIT) {
-        // We only get TaskEvents for compilation units if they contain no package declarations
-        // (e.g. package-info.java files).  In this case it's safe to analyze the
-        // CompilationUnitTree immediately.
-        if (!isExcludedPath(compilation.getSourceFile().toUri())) {
+      if (!isExcludedPath(compilation.getSourceFile().toUri())) {
+        if (path.getLeaf().getKind() == Tree.Kind.COMPILATION_UNIT) {
+          // We only get TaskEvents for compilation units if they contain no package declarations
+          // (e.g. package-info.java files).  In this case it's safe to analyze the
+          // CompilationUnitTree immediately.
           transformer.get().apply(path, subContext, descriptionListener);
-        }
-      } else if (finishedCompilation(path.getCompilationUnit())) {
-        // Otherwise this TaskEvent is for a ClassTree, and we can scan the whole
-        // CompilationUnitTree once we've seen all the enclosed classes.
-        if (!isExcludedPath(compilation.getSourceFile().toUri())) {
+        } else if (finishedCompilation(path.getCompilationUnit())) {
+          // Otherwise this TaskEvent is for a ClassTree, and we can scan the whole
+          // CompilationUnitTree once we've seen all the enclosed classes.
           transformer.get().apply(new TreePath(compilation), subContext, descriptionListener);
         }
       }
--- a/check_api/src/main/java/com/google/errorprone/ErrorProneAnalyzer.java
+++ b/check_api/src/main/java/com/google/errorprone/ErrorProneAnalyzer.java
@@ -24,6 +24,7 @@ import com.google.common.base.Suppliers;
 import com.google.common.base.Throwables;
 import com.google.errorprone.scanner.ErrorProneScannerTransformer;
 import com.google.errorprone.scanner.ScannerSupplier;
+import com.google.errorprone.util.ASTHelpers;
 import com.sun.source.tree.CompilationUnitTree;
 import com.sun.source.tree.Tree;
 import com.sun.source.util.TaskEvent;
@@ -172,7 +173,7 @@ public class ErrorProneAnalyzer implements TaskListener {
    */
   private boolean isExcludedPath(URI uri) {
     Pattern excludedPattern = errorProneOptions.getExcludedPattern();
-    return (excludedPattern != null) && excludedPattern.matcher(uri.toString()).matches();
+    return (excludedPattern != null) && excludedPattern.matcher(ASTHelpers.getFileNameFromUri(uri)).matches();
   }
 
   /**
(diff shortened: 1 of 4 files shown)


kageiit:
cc @msridhar

vanniktech:
This would also fix #511

kageiit:
@cushon ping on this

nglorioso:
Hello! We discussed this a bit:

We believe that the combination of `-XepDisableWarningsInGeneratedCode` and separating the compilation of generated code into its own compiler invocation (perhaps without Error Prone enabled, or with Error Prone configured to turn off a select set of checks) may be a better long term idea.

The flag allows error prone to skip generating warnings for code annotated with `@Generated`, but will still break the build if the generated code violates the error-level patterns (i.e.: the stuff that would break your build if you wrote it by hand). We've selected the list of error-level checks based on our impression that these represent real issues that should be corrected.

If you find that the generated code is consistently violating these error-level checks, we may want to consider reclassifying the checks, or fixing the code generator if the code it's generation contains the buggy code.

kageiit:
Hey @nglorioso This is for more than just generated code. There are cases where we have vendor/third party code that needs to be part of the same sourceset. Also, not all code generators add the `@Generated` annotation, so just the existing `-XepDisableWarningsInGeneratedCode` options will not work for those cases.

I think having an option like this would make it very flexible and let consumers suppress generated code issues etc. and also any other potential future use cases like during a migration when two modules are merged etc, to turn off EP on a particular path so it can be done incrementally etc. when dealing with large projects at scale

hzsweers:
Ping on this?

vladgurovich:
Furthermore, this seems like something that would enable a gradual rollout of this tool on a legacy code that would otherwise generate hundreds of errors or warnings

check_api/src/main/java/com/google/errorprone/ErrorProneAnalyzer.java
```diff
@@ -160,7 +167,17 @@ public void finished(TaskEvent taskEvent) {
     }
   }
 
-  /** Returns true if all declarations inside the given compilation unit have been visited. */
+  /**
+   * Returns true if all declarations inside the given compilation unit have been visited.
+   */
+  private boolean isExcludedPath(URI uri) {
+    Pattern excludedPattern = errorProneOptions.getExcludedPattern();
+    return (excludedPattern != null) && excludedPattern.matcher(uri.toString()).matches();
```
cushon: Can you use `ASTHelpers.getFileNameFromUri` instead of `URI.toString`? The former has special handling for jar entries, which I think it's possible to see here.

kageiit:
@cushon updated as per your suggestion

cushon:
LGTM, but can you add test coverage?

msridhar:
I'm ok with my commits being included

check_api/src/main/java/com/google/errorprone/ErrorProneAnalyzer.java
```diff
@@ -138,11 +142,15 @@ public void finished(TaskEvent taskEvent) {
...
         // CompilationUnitTree immediately.
-        transformer.get().apply(path, subContext, descriptionListener);
+        if (!isExcludedPath(compilation.getSourceFile().toUri())) {
+          transformer.get().apply(path, subContext, descriptionListener);
+        }
       } else if (finishedCompilation(path.getCompilationUnit())) {
         // Otherwise this TaskEvent is for a ClassTree, and we can scan the whole
         // CompilationUnitTree once we've seen all the enclosed classes.
-        transformer.get().apply(new TreePath(compilation), subContext, descriptionListener);
+        if (!isExcludedPath(compilation.getSourceFile().toUri())) {
```
cushon: It would be nice to de-duplicate this logic.

check_api/src/main/java/com/google/errorprone/ErrorProneOptions.java
```diff
@@ -412,4 +435,12 @@ public static ErrorProneOptions processArgs(String[] args) {
     Preconditions.checkNotNull(args);
     return processArgs(Arrays.asList(args));
   }
+
+  private static Pattern getExcludedPattern(Iterable<String> excludeSubStrings) {
```
cushon: WDYT about making the flag take a regex, instead of a comma-separated list that gets processed into a regex?

check_api/src/test/java/com/google/errorprone/ErrorProneOptionsTest.java
```diff
@@ -164,6 +166,20 @@ public void recognizesCompilingTestOnlyCode() {
   }
 
   @Test
+  public void recognizesExcludedPaths() {
```
cushon: Can you add an integration test that verifies the option is actually used to filter the compilation units that get analyzed?

check_api/src/main/java/com/google/errorprone/ErrorProneOptions.java
msridhar: Makes sense; will change

check_api/src/test/java/com/google/errorprone/ErrorProneOptionsTest.java
msridhar: Will look into it; is there an existing test whose structure I can copy?

cushon: There are test cases in [`ErrorProneJavaCompilerTest`](https://github.com/google/error-prone/blob/a9f4c95bb5648a8e1ef56e6570033ebaa4ad95f1/core/src/test/java/com/google/errorprone/ErrorProneJavaCompilerTest.java) that use flags to enable/disable checks and change severities. Maybe something like that?

msridhar: Cool, thanks!  I added a test case there.

cushon:
Thanks! I'm merging this change internally and it will show up on github in the next day or so.

Can you also create a PR to add documentation to http://errorprone.info/docs/flags?

msridhar:
Awesome!  Will create a documentation PR.

vorburger:
FTR: Not release as of right now, but https://github.com/google/error-prone/issues/821 says about to...

PR #607, opened 2017-04-26
Fix Markdown of links in README

ejona86:
```
The []() and [][] styles of linking can't be separated by spaces/newline
on GitHub. I kept the []()-style for short URLs, but for long URLs that
caused line-wrapping issues I swapped to the [][] to make the text more
readable.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,23 +29,27 @@ its type int is not compatible with its collection's type argument Short
 
 Our documentation is at [errorprone.info](http://errorprone.info).
 
-Error Prone works with [Bazel](http://bazel.io), [Maven]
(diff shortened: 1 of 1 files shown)


ejona86:
:-/ It seems  https://github.com/google/error-prone/wiki/For-Developers and https://github.com/google/error-prone/wiki/Writing-a-check are also broken and could use fixing.

https://github.com/google/error-prone/wiki/Writing-a-check also accidentally uses ````java (instead of ```java; 3 vs 4) prevents it from "closing" the code literal.

nglorioso:
Thanks for the pointers! I've cleaned up the referenced pages.

PR #635, opened 2017-05-29
Remove redundant calls to `String.format()`

jbduncan:
I found these redundant calls to `String.format()` using IntelliJ IDEA's "Inspect Code" feature. Hopefully the changes are rather self-explanatory, but if not then I'd be happy to explain them in detail.

Overall, I submit this PR in the hope that the error-prone team will find it worth merging. :)


jbduncan:
Thanks @eaftan! :+1:

PR #637, opened 2017-05-29
Optimize string concatenations within `StringBuilder.append()` calls

jbduncan:
The change in this PR was made using IntelliJ IDEA's "Inspect Code" feature. The reason for this change is explained in the commit message, which I've pasted below.

```text
Concatenating strings as an argument to `StringBuilder.append()` causes
a new, temporary StringBuilder to be created just for that argument on
modern JVMs, which is a little wasteful. This commit replaces those
concatenations with separate calls to `StringBuilder.append()` so as to
make sure that only one StringBuilder is created, rather than
(1 + concat-args) StringBuilders.
```

Overall, I submit this PR in the hope that the error-prone team will find it worth merging.

### DIFF
--- a/check_api/src/main/java/com/google/errorprone/fixes/SuggestedFix.java
+++ b/check_api/src/main/java/com/google/errorprone/fixes/SuggestedFix.java
@@ -68,8 +68,13 @@ public class SuggestedFix implements Fix {
     StringBuilder result = new StringBuilder("replace ");
     for (Replacement replacement : getReplacements(compilationUnit.endPositions)) {
       result
-          .append("position " + replacement.startPosition() + ":" + replacement.endPosition())
-          .append(" with \"" + replacement.replaceWith() + "\" ");
+          .append("position ")
+          .append(replacement.startPosition())
+          .append(":")
+          .append(replacement.endPosition())
+          .append(" with \"")
+          .append(replacement.replaceWith())
+          .append("\" ");
     }
     return result.toString();
   }
--- a/check_api/src/main/java/com/google/errorprone/fixes/SuggestedFix.java
+++ b/check_api/src/main/java/com/google/errorprone/fixes/SuggestedFix.java
@@ -67,14 +67,10 @@ public class SuggestedFix implements Fix {
   public String toString(JCCompilationUnit compilationUnit) {
     StringBuilder result = new StringBuilder("replace ");
     for (Replacement replacement : getReplacements(compilationUnit.endPositions)) {
-      result
-          .append("position ")
-          .append(replacement.startPosition())
-          .append(":")
-          .append(replacement.endPosition())
-          .append(" with \"")
-          .append(replacement.replaceWith())
-          .append("\" ");
+      result.append(
+          String.format(
+              "position %d:%d with \"%s\"",
+              replacement.startPosition(), replacement.endPosition(), replacement.replaceWith()));
     }
     return result.toString();
   }
--- a/check_api/src/main/java/com/google/errorprone/fixes/SuggestedFix.java
+++ b/check_api/src/main/java/com/google/errorprone/fixes/SuggestedFix.java
@@ -69,7 +69,7 @@ public class SuggestedFix implements Fix {
     for (Replacement replacement : getReplacements(compilationUnit.endPositions)) {
       result.append(
           String.format(
-              "position %d:%d with \"%s\"",
+              "position %d:%d with \"%s\" ",
               replacement.startPosition(), replacement.endPosition(), replacement.replaceWith()));
     }
     return result.toString();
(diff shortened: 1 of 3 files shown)


check_api/src/main/java/com/google/errorprone/fixes/SuggestedFix.java
```diff
@@ -68,8 +68,13 @@ public String toString(JCCompilationUnit compilationUnit) {
     StringBuilder result = new StringBuilder("replace ");
     for (Replacement replacement : getReplacements(compilationUnit.endPositions)) {
       result
-          .append("position " + replacement.startPosition() + ":" + replacement.endPosition())
-          .append(" with \"" + replacement.replaceWith() + "\" ");
+          .append("position ")
```
ronshapiro: What about using `String.format`? Seems easier to me to read.

```java
result.append(String.format("position %d:%d with \"%s\", replacement.startPosition(), replacement.endPosition(), replacement.replaceWith())
```

lowasser:
None of this code is, I think, performance sensitive?  I'd definitely prioritize readability over performance here.

check_api/src/main/java/com/google/errorprone/fixes/SuggestedFix.java
jbduncan: @ronshapiro Good point! That's definitely readable, and that wouldn't report any warnings in my copy of IntelliJ, which was my main reason for submitting this PR (as I'm a huge stickler for these sorts of things :stuck_out_tongue_winking_eye:).

jbduncan:
@lowasser That's a very good point. My main reason for submitting this PR was because I'm a stickler for reducing the number of warnings that my copy of IntelliJ produces. Would replacing these string concatenations with `String.format()` or something similar be a good compromise, like @ronshapiro suggested in https://github.com/google/error-prone/pull/637#discussion_r118989742?

core/src/main/java/com/google/errorprone/bugpatterns/AbstractSuppressWarningsMatcher.java
```diff
@@ -73,7 +73,7 @@ protected final Fix getSuggestedFix(AnnotationTree annotationTree) {
       StringBuilder sb = new StringBuilder("@SuppressWarnings({\"" + values.get(0) + "\"");
       for (int i = 1; i < values.size(); i++) {
         sb.append(", ");
-        sb.append("\"" + values.get(i) + "\"");
+        sb.append('"').append(values.get(i)).append('"');
```
ronshapiro: Can this entire string be:

```java
"@SuppressWarnings({\"" + Joiner.on("\", \"").join(values) + "})"
```

May be more or less readable, depending on your style

core/src/main/java/com/google/errorprone/bugpatterns/PreconditionsCheckNotNullPrimitive.java
```diff
@@ -157,10 +157,10 @@ private Fix createCheckArgumentOrStateCall(
 
     // Was the original call to Preconditions.checkNotNull a static import or not?
     if (methodInvocationTree.getMethodSelect().getKind() == Kind.IDENTIFIER) {
-      replacement.append(replacementMethod + "(");
+      replacement.append(replacementMethod).append("(");
```
ronshapiro: What about pulling this out of the if/else entirely? Then the else has append("Preconditions") and below has .append(replacementMethod).append('(') for both cases?

core/src/main/java/com/google/errorprone/bugpatterns/AbstractSuppressWarningsMatcher.java
jbduncan: I'd consider your suggestion to be more readable, as everything is in one line and easier to parse (in my mind, at least). Thanks!

core/src/main/java/com/google/errorprone/bugpatterns/PreconditionsCheckNotNullPrimitive.java
jbduncan: @ronshapiro Am I right to understand that you're suggesting something like this?

```java
if (methodInvocationTree.getMethodSelect().getKind() == Kind.IDENTIFIER) {
  fix.addStaticImport("com.google.common.base.Preconditions." + replacementMethod);
} else {
  replacement.append("Preconditions.")
}
replacement..append(replacementMethod).append('(');
```

core/src/main/java/com/google/errorprone/bugpatterns/AbstractSuppressWarningsMatcher.java
jbduncan: @ronshapiro Is there an established preference within error-prone for `Joiner`, or would Java 8's `String.join()` be fine instead?

jbduncan:
I'll rename the PR and squash the commits if it passes the review.

core/src/main/java/com/google/errorprone/bugpatterns/AbstractSuppressWarningsMatcher.java
jbduncan: I've decided to use `Joiner` for consistency.

jbduncan:
Travis and AppVeyor are fixed now.

PR #638, opened 2017-05-30
Moe sync

ronshapiro:


ronshapiro:
Both of the commits that were warned on by @googlebot were made by googlers.

PR #642, opened 2017-06-14
Moe sync

ronshapiro:


ronshapiro:
@googlebot both of those authors are done their internships and no longer at Google, but authored those commits while at the company

ronshapiro:
@cushon @eaftan these continue to fail - but maybe we should merge it anyway and fix-forward since a lot of commits are in here?

eaftan:
Sorry for the slow reply, @ronshapiro.  I'm going to disable the failing test, then we can push the update.

PR #665, opened 2017-07-02
Fix URL markdown problems.

tasomaniac:
I noticed that the markdown links were broken. Here is the fix.


eaftan:
Thanks for the fix.  This happens because of differences in the markdown implementation we use internally and the one that Github uses.

PR #671, opened 2017-07-10
Allow long non-file arguments to be spilled

novalis:
Allow long non-file arguments (e.g. -classpath) to go to a file.  This
prevents E2BIG when the classpath argument is too long.

This can't easily be fixed in upstream ant because:
(a) it would require an API change, and
(b) there is no way for ant to know which arguments are safe to spill
as that depends on the semantics of the external compiler.


novalis:
My company has signed the corporate CLA, and @googlebot has in the past gotten this correct -- see e.g. https://github.com/google/google-java-format/pull/163

novalis:
Oh, hm, maybe it was because I wasn't part of the Two Sigma group on github yet?  Let me retry...

novalis:
I signed it!

novalis:
Our Google accounts were messed up, and I think are now fixed, so maybe if I retry now the CLA-bot will work?

novalis:
Just a ping on this.  Thanks.

PR #677, opened 2017-07-14
Update maven-project-info-reports-plugin version.

tawanjairew:
The previous version we were using was printing a stack trace on
invokedynamic bytecodes.  This filled up the logs and caused our
Travis build to fail.


eaftan:
Thanks!

PR #686, opened 2017-07-20
Add standalone (and backport) JSR-310 library as immutable.

JakeWharton:
The listed types are equivalent to those in `java.time.*`.


eaftan:
Thanks @JakeWharton!

PR #691, opened 2017-07-26
Document how plugin checkers can accept configuration flags

msridhar:
In particular, document the fact that in addition to the
constructor taking an ErrorProneFlags object, plugin checkers
must also define a zero-argument constructor, in order to be
loaded by the ServiceLoader.


eaftan:
Thanks!

Issue #472, opened 2016-11-01
New *Nullable bugpatterns not made available?


Stephan202:
The [release announcement](https://groups.google.com/forum/#!topic/error-prone-discuss/E2OXIVKOPJE) mentions three new suggestions related to nullability analysis, namely:

* `FieldMissingNullable`
* `ParameterNotNullable`
* `ReturnMissingNullable`

When I try use use any of these I get an error of the form:

> FieldMissingNullable is not a valid checker name

Running the following script in the `error-prone` source root, it seems the above checkers are the only ones not listed in `BuiltInCheckerSuppliers.java`:
```sh
#!/bin/sh

find ./core/src/main/java/com/google/errorprone/bugpatterns -name "*.java" \
  | xargs grep -l @BugPattern \
  | xargs -I{} basename '{}' .java \
  | while read C; do
      grep -q "$C" ./core/src/main/java/com/google/errorprone/scanner/BuiltInCheckerSuppliers.java \
        || echo "$C"
    done
```

m-titov:
Please fix this bug

eaftan:
We'd be happy to take a PR for this.  Those checks should go in the disabled section because they are intended as opt-in suggestions rather than errors or even warnings.

m-titov:
Please approve pull request and release it as soon as possible. Thanks.

m-titov:
Hi @eaftan When it could be released?

Stephan202:
This was fixed version 2.1.2. Thanks!

PR #699, opened 2017-08-05
New *Nullable bugpatterns not made available?

originalrusyn:
https://github.com/google/error-prone/issues/472


originalrusyn:
I signed it!

PR #700, opened 2017-08-05
New *Nullable bugpatterns not made available?

originalrusyn:
https://github.com/google/error-prone/issues/472


eaftan:
These should not be errors because not everyone uses nullability annotation in their code, and even if they do, different checkers use different nullability semantics.  Can you change this to add them to the disabled list, and then you can enable them for your project by passing command line flags?

originalrusyn:
> Can you change this to add them to the disabled list, and then you can enable them for your project by passing command line flags?

What do you mean? It's already in ImmutableSet<BugCheckerInfo> DISABLED_CHECKS only.

eaftan:
Sorry, misread the diff.  Thanks for the contribution.

PR #740, opened 2017-09-06
Moe sync

ronshapiro:


cushon:
CLAs were signed in #721 and #671

ronshapiro:
@cushon can you submit if everything is OK?

PR #747, opened 2017-09-12
clear out Checker Framework cache after compilation to avoid leak

msridhar:
This is a workaround for typetools/checker-framework#1482, which can affect any build process that runs javac in memory as part of a daemon.  Change is pretty simple; after compilation, we clear out the caches to avoid the leak.  This is not an ideal fix for concurrent builds (since there is just one global cache), but that would require some deeper changes to the CF.

### DIFF
--- a/check_api/src/main/java/com/google/errorprone/BaseErrorProneJavaCompiler.java
+++ b/check_api/src/main/java/com/google/errorprone/BaseErrorProneJavaCompiler.java
@@ -31,9 +31,6 @@ import com.sun.tools.javac.util.Context;
 import com.sun.tools.javac.util.JavacMessages;
 import com.sun.tools.javac.util.Log;
 import com.sun.tools.javac.util.Log.WriterKind;
-
-import org.checkerframework.javacutil.AnnotationUtils;
-
 import java.io.InputStream;
 import java.io.OutputStream;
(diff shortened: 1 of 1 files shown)


check_api/src/main/java/com/google/errorprone/BaseErrorProneJavaCompiler.java
```diff
@@ -31,6 +31,9 @@
 import com.sun.tools.javac.util.JavacMessages;
 import com.sun.tools.javac.util.Log;
 import com.sun.tools.javac.util.Log.WriterKind;
+
+import org.checkerframework.javacutil.AnnotationUtils;
+
```
jbduncan: @msridhar IIRC, Google Java style now dictates that all non-static imports must stay together, so I think it's invalid for there to be any blank lines before and after this `org.checkerframework.javacutil.AnnotationUtils` import.

msridhar: @jbduncan fixed (also put it in alphabetical order).  Thanks!

jbduncan: You're very welcome @msridhar! :+1:

PR #752, opened 2017-09-18
Moe sync

ronshapiro:
Fixes https://github.com/google/error-prone/issues/497


ronshapiro:
Both of those commits have an accompanying PR

PR #753, opened 2017-09-21
Update Gradle example to latest version

tbroyer:
Use new Gradle plugins DSL, and new way to define errorprone version;
makes the build script much simpler and easier to understand.


nglorioso:
Thanks, this looks much cleaner!

We're committing this internally and it should be pushed out in the next MOE sync PR (ala #752)

Issue #762, opened 2017-10-07
LinkedList accessed through via index


zero-88:
java.util.LinkedList accessed through via index.
Reason: This is bad performance since random access via index on a linked list is O(n)
List methods to check:
- add(int i, Object obj)
- get(int i)
- remove(int i)
- set(int i, Object obj)

zero-88:
@nglorioso @lowasser ping on this.
Please help me review this issue.
Thanks

cushon:
Thanks for the PR, someone will take a look at this next week.

cpovirk:
As @ronshapiro points out in #763, it's hard to identify that a `List` is a `LinkedList` unless that's its static type. You might get more bang for your buck for flagging all creations of `LinkedList`. My understand is that, unless you are doing something very strange, `LinkedList` is slower than `ArrayList`, `ArrayDeque`, or both.

zero-88:
Hi @cpovirk @ronshapiro 
Sorry about checking for `matchInstanceOf` when initialize `LinkedList`. I intended to check define `LinkedList` by `List` interface. But it seems impossible.
Use `addFirst` instead of `myLinkedList.add(0, item)`, and similar with other methods: `remove`, `get`...
Because when using index with LinkedList, this class recomputes list item and make worst performance. You can see method in `LinkedList`:
```
/**
     * Returns the (non-null) Node at the specified element index.
     */
    Node<E> node(int index) {
        // assert isElementIndex(index);

        if (index < (size >> 1)) {
            Node<E> x = first;
            for (int i = 0; i < index; i++)
                x = x.next;
            return x;
        } else {
            Node<E> x = last;
            for (int i = size - 1; i > index; i--)
                x = x.prev;
            return x;
        }
    }
```

PR #763, opened 2017-10-07
Implement #762

zero-88:
@cushon  #762

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/LinkedListChecker.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/LinkedListChecker.java
@@ -1,0 +1,40 @@
+package com.google.errorprone.bugpatterns;
+
+import static com.google.errorprone.BugPattern.SeverityLevel.ERROR;
+import static com.google.errorprone.matchers.Description.NO_MATCH;
+import com.google.auto.service.AutoService;
+import com.google.errorprone.BugPattern;
+import com.google.errorprone.BugPattern.Category;
+import com.google.errorprone.VisitorState;
+import com.google.errorprone.bugpatterns.BugChecker.MethodInvocationTreeMatcher;
+import com.google.errorprone.matchers.Description;
+import com.google.errorprone.matchers.Matcher;
+import com.google.errorprone.matchers.Matchers;
+import com.google.errorprone.matchers.method.MethodMatchers;
+import com.sun.source.tree.MethodInvocationTree;
+
+@AutoService(BugChecker.class)
+@BugPattern(name = "LinkedListChecker", summary = "java.util.LinkedList accessed through via index",
+    severity = ERROR, category = Category.JDK)
+public final class LinkedListChecker extends BugChecker implements MethodInvocationTreeMatcher {
+
+  private static final String LINKED_LIST_FQN = "java.util.LinkedList";
+  private static final Matcher<MethodInvocationTree> INSTANCE_LINKEDLIST_MATCHER = Matchers.anyOf(
+      MethodMatchers.instanceMethod().onDescendantOf(LINKED_LIST_FQN).named("add")
+          .withParameters("int", "java.lang.Object"),
+      MethodMatchers.instanceMethod().onDescendantOf(LINKED_LIST_FQN).named("addAll"),
+      MethodMatchers.instanceMethod().onDescendantOf(LINKED_LIST_FQN).named("get"),
+      MethodMatchers.instanceMethod().onDescendantOf(LINKED_LIST_FQN).named("remove")
+          .withParameters("int"),
+      MethodMatchers.instanceMethod().onDescendantOf(LINKED_LIST_FQN).named("set")
+          .withParameters("int", "java.lang.Object"));
+
+  @Override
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/google/errorprone/bugpatterns/LinkedListChecker.java
```diff
@@ -0,0 +1,48 @@
...
+
+@AutoService(BugChecker.class)
+@BugPattern(name = "LinkedListChecker", summary = "java.util.LinkedList accessed through via index",
+    severity = ERROR, category = Category.JDK)
+public final class LinkedListChecker extends BugChecker
+    implements MethodInvocationTreeMatcher, InstanceOfTreeMatcher {
+
+  private static final String LINKED_LIST_FQN = "java.util.LinkedList";
+  private static final Matcher<MethodInvocationTree> INSTANCE_LINKEDLIST_MATCHER = Matchers.anyOf(
+      MethodMatchers.instanceMethod().onDescendantOf(LINKED_LIST_FQN).named("add")
```
ronshapiro: This isn't necessarily bad. `myLinkedList.add(0, item)` is actually probably a good intended use of a `LinkedList`.

Maybe at the least special-case `0`, or even any small contant (or perhaps all constants)?

The same applies for the other methods below.

core/src/main/java/com/google/errorprone/bugpatterns/LinkedListChecker.java
```diff
@@ -0,0 +1,48 @@
...
+  @Override
+  public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+    if (!INSTANCE_LINKEDLIST_MATCHER.matches(tree, state)) {
+      return NO_MATCH;
+    }
+    return this.describeMatch(tree);
+  }
+
+  @Override
+  public Description matchInstanceOf(InstanceOfTree tree, VisitorState state) {
```
ronshapiro: Did you mean to remove this (and the interface that requires it)?

ronshapiro:
This check requires people to store references as a `LinkedList<T>` instead of just a plain `List<T>` which we tend to prefer (Effective Java 2nd edition Item 52). In general, the performance of `LinkedList` isn't so great compared to `LinkedList`, so I'm not sure if this will get many hits.

zero-88:
@ronshapiro 
I already updated code. Could you help me to review it again?
Anything misunderstood, I left comment in #762

cushon:
> In general, the performance of `LinkedList` isn't so great compared to [`ArrayList`], so I'm not sure if this will get many hits.

Given that, I think we're better off discouraging uses of `LinkedList` altogether (see [`JdkObsolete`](https://github.com/google/error-prone/blob/master/core/src/main/java/com/google/errorprone/bugpatterns/JdkObsolete.java)), rather than warning about specific uses of `LinkedList`.

zero-88:
Hi @cushon, I agree with your solution. 
However, I just wondering:
- Should we point out exact issue in description? In this case, it is a list method that I checked.
- Seems `JdkObsolete` not yet checked inheritance. E.g: Clazz extends HashTable/LinkedList

cushon:
> Should we point out exact issue in description? In this case, it is a list method that I checked.

I'm not sure if I understand. Are you suggesting having `JdkObsolete` emit more specific diagnostics, or also handling method calls on the obsolete types (like this check does), or something else? My thinking was that checking instantiation (and now extending/implementing) the obsolete classes is enough. We want to prevent new uses of these types from being introduced, but if you're calling `ArrayList#get` it might be because some API you don't have control over returned that type, so it would be annoying to get a warning about it.

> Seems JdkObsolete not yet checked inheritance. E.g: Clazz extends HashTable/LinkedList

Good idea: 319c50539eeab6442963e23a26f8b7119539a778.

zero-88:
Actually, I would like to indicate exact operation that slow down performance than restrict use any class.
For example, linkedlist is faster than arraylist in case of using operation without index. It is good point to use it.
Also, it helps to scan old system when using class operation instead of checking new instance. Somehow, client don't want a change their structure but can refactor some method.
However, the current limitation is we cannot check if var's declared as List interface.
That's my idea.

Stephan202:
>  For example, linkedlist is faster than arraylist in case of using operation without index. It is good point to use it.

Can you expand on that? ([Potentially relevant video](https://www.youtube.com/watch?v=YQs6IC-vgmo))

cushon:
Our benchmarking shows that `ArrayList` or `ArrayDeque` are effectively always a better choice. I don't think it's worth trying to discourage particular uses of `LinkedList` instead of discouraging *all* uses. I can imagine people seeing this check and getting the idea that `LinkedList` is a good choice as long as they don't need to index into it, but it really isn't.

If you'd still like to use this check on your code base, you could add it as a [plugin check](http://errorprone.info/docs/plugins).

cushon:
Also, https://twitter.com/joshbloch/status/583813919019573248.

zero-88:
Thank @cushon and @Stephan202 for interesting information.
I will have a look on plugin check because based on my requirements. 
We can close topic at here 👍

PR #769, opened 2017-10-11
Add Date equality checker

tschwery:
This check verifies that two dates are not compared using the instance `equals` or `Objects.equals` because of problems when one of the instances are descendants of Date, such as `sql.Timestamp`. In such cases, the comparison can return false every time. The `compareTo` method has no such problems.

This check is added as disabled and with a level of suggestion because code bases that don't use dates from `sql.Timestamp` (for example from entities) are not really impacted by this behavior.

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/DateEquals.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/DateEquals.java
@@ -1,0 +1,95 @@
+/*
+ * Copyright 2012 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.errorprone.bugpatterns;
+
+import static com.google.errorprone.matchers.Description.NO_MATCH;
+import static com.google.errorprone.matchers.Matchers.allOf;
+import static com.google.errorprone.matchers.Matchers.anyOf;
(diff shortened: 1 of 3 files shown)


cushon:
Thanks for the PR!

You can simplify this by using this abstract class: https://github.com/google/error-prone/blob/master/core/src/main/java/com/google/errorprone/bugpatterns/AbstractReferenceEquality.java

We also have a more general check for reference equality: http://errorprone.info/bugpattern/ReferenceEquality. Are you interested in a separate check for this so you can make reference equality an error specifically for `Date`s, but not for other types?

tschwery:
I don't know if I misunderstood some of these checks, but from what I understand, these check that objects are not compared (by reference) using `==` but using `equals` methods. 

In this check, I want to check that `Date` objects are not compared using `equals` methods but only by using `compareTo`, which give correct `Date` comparison even in its subclasses.

If it's already possible with the existing checks, I failed to see how before writing this and would gladly appreciate a pointing in the correct direction.

cushon:
I mis-read the check, sorry. You're right that `AbstractReferenceEquality` isn't useful here.

Is this only a problem for `sql.Timestamp`, this might be a good
http://errorprone.info/docs/plugins

core/src/main/java/com/google/errorprone/bugpatterns/DateEquals.java
```diff
@@ -0,0 +1,95 @@
...
+              + "period equality.",
+  severity = SUGGESTION,
+  providesFix = ProvidesFix.REQUIRES_HUMAN_ATTENTION
+)
+public class DateEquals extends BugChecker implements MethodInvocationTreeMatcher {
+  /** Matches when the equals instance method is used to compare two Date objects. */
+  private static final Matcher<MethodInvocationTree> instanceEqualsMatcher =
+      Matchers.allOf(
+          instanceMethod().onClass(TypePredicates.isDescendantOf("java.util.Date")).named("equals"),
+          argument(0, Matchers.<ExpressionTree>isSubtypeOf("java.util.Date")));
```
cushon: Are you sure you want to only match if the static type of the argument is a Date, or would it make sense to match all calls to Date.equals even if the static type of the argument was e.g. an Object?

tschwery: I wanted to limit the scope of this test so that it only impact you if you compare two dates but I'm not opposed to broadening the match.

PR #785, opened 2017-10-21
Add -XepDisableAllWarnings flag

msridhar:
This flag turns off all WARNING-level checks, while leaving on those at ERROR level.  This is useful for those who want to compile with `-Werror` but don't want any EP WARNING-level issue to block the build.  Individual WARNING level checks can still be set to ERROR level as desired.


thalescm:
Is there any chance this might ever be merged to this repo?

danielkaneider:
Any chance to rebase this PR and get it merged?

msridhar:
I'm happy to do the rebase work if there is interest in merging this PR

msridhar:
Let me pitch another use case for this flag.  For an existing project trying to adopt Error Prone, the initial number of errors + warnings can be overwhelming.  While the warnings will not block the build, they do introduce a lot of noise in the build logs, which could turn off new users.  (I just ran into this with a student trying to introduce Error Prone to a project today.)  The proposed flag could give users a more incremental way to adopt Error Prone (first fix errors, then warnings), or a less noisy way if they never intend to fix the warnings.

Again, I am happy to rebase and update the PR if there is interest in landing it.  And if there is no interest, then we can close it out.

/cc @eaftan

Issue #595, opened 2017-04-07
XepPatchChecks, default checks and refaster check


BKer:
Hi,

At the moment I’m on my phone, so typing isn’t the most comfortable thing to do.

I tried to combine `-XepPatchChecks` with a refaster check. When, deliberately adding errors the checks generate a patch file, however the custom refaster check is not present in the patch file. Running refaster without the other checks generates the patch file with the custom refaster refactorings.
This is my `-XepPatchChecks` parameter passed to the compiler:
`-XepPatchChecks:MissingOverride,DefaultCharset,DeadException,refaster:/path/to/my/CustomRule.refaster`
How can I have a patch file containing both the modifications made by the default checks and my custom refaster rule?
I hope someone can point me into the right direction :)

eaftan:
@nglorioso, can you take a look?

nglorioso:
https://github.com/google/error-prone/blob/master/check_api/src/main/java/com/google/errorprone/ErrorProneOptions.java#L322 currently we don't allow concatenating Refaster rules and built-in checkers. We should allow this, but it's a bit hacky due to the way these are constructed (there's a separation from the options parsing and loading the built-in checkers).

We have plans to improve the experience here, but I can see if this functionality is easy to wedge into the command line flags without being too awkward :)

Issue #552, opened 2017-03-02
`RefasterRuleCompiler` ignores all but one input file


Stephan202:
First of all: Refaster is awesome. Defining new rules is child's play. Super cool!

While playing around with it I defined a number of rewrite rules. If possible I'd like pass all of these to Error Prone in one go. But, while cursory look at the `RefasterRuleCompiler` indicates that it outputs a _list_ of `CodeTransformer`s to the file specified using the `--out` flag, only one of the input `.java` files actually contributes to the output. Would it be possible to combine multiple before/after templates into a single `.refaster` file?

(I'm not familiar with `javac` internals, but if you give me some pointers I'm willing to implement this generalization. Provided that you think this is a sound idea in the first place, of course.)

nglorioso:
Thanks for the bug file!

It looks like the RefasterRuleAnalyzer appears to output after each compilation unit: https://github.com/google/error-prone/blob/master/refaster/src/main/java/com/google/errorprone/refaster/RefasterRuleCompilerAnalyzer.java#L61

I think we can restructure this to collect over all of the classes then emit at the end.

In the meantime: does this pattern work?

```java
class MyWrapper {
   static class Rule1 { @BeforeTemplate ... }
   static class Rule2 { @BeforeTemplate ... }
}
```

and compile MyWrapper.java?

lowasser:
FWIW: internally the refaster rule compiler only accepts one file at a time, but we also have a separate tool that assembles a "suite" (which can include mixed Refaster and Error Prone refactorings).  We might be able to externalize that bit?

Stephan202:
@nglorioso: GMTA, I did try the wrapper trick ;). Unfortunately that yields an empty list of code transformers. (I.e., same as if one would compile an empty class.) I do think there are good use cases for supporting that feature, though: if one has several closely-related rewrite rules (e.g. rewriting `toList()` and `toSet()` to `toImmutableList()` and `toImmuntableSet()`, respectively), then it makes sense to group those together in the same class.

@lowasser: that sounds exactly like what I'm look for :). What I'd like to do is create a Maven project which documents, in Error-Prone compatible form, a number of company-internal best-practices. The idea would be to use Refaster where possible, and fall back to Error Prone plugins where necessary. Ideally Error Prone would treat them identically (discovery through service loading, automatic patching, etc).

Stephan202:
[...]

PR #787, opened 2017-10-23
Generalize `-XepPatchChecks` command line parsing

Stephan202:
This is a minimal (?) fix for #595. It adds support for:
- `-XepPatchChecks` arguments combining built-in and Refaster rules.
- `-XepPatchChecks` arguments listing multiple Refaster rules.

Open points and questions:
- Is this approach acceptable at all?
- If so, I'll add some tests.
- Right now Refaster is treated specially in both `ErrorProneOptions` and `BaseErrorProneJavaCompiler`; a nicer solution would entail making Refaster a first-class citizen. That'd be more work, but if you provide some pointers on the desired "end goal" or boundaries of the design space, I'm willing to have a crack at that, too. (In the context of a separate PR, ideally.)


check_api/src/main/java/com/google/errorprone/BaseErrorProneJavaCompiler.java
```diff
@@ -199,24 +201,31 @@ static ErrorProneAnalyzer createAnalyzer(
...
     return ErrorProneAnalyzer.createWithCustomDescriptionListener(
-        codeTransformer, epOptions, context, refactoringCollection[0]);
+        compoundTransformer, epOptions, context, refactoringCollection[0]);
+  }
+
+  private static CodeTransformer createBuiltinRefactorer(
+      ScannerSupplier scannerSupplier,
+      ErrorProneOptions epOptions,
+      Context context,
+      boolean forceFilter) {
```
Stephan202: This `forceFilter` flag makes sure that the default patch mode of applying all built-in checks is disabled when a Refaster rule is supplied. I added this for backwards compatibility, but it does mean there's (still) no way to apply all built-in checks _and_ any additional Refaster rules. That's another way in which Refaster is "special" (contrast this with service-loaded custom plugins:  those are applied by default).

This is another thing I'd like to improve; input welcome!

Stephan202:
Rebased the branch and resolved a conflict. What does the Error Prone team think of this approach?

knutwannheden:
@Stephan202 Is this patch also available in the forked v2.3.4-picnic-2 release? I tried specifying the path to multiple .refaster check files, but the compiler then failed.

Stephan202:
@knutwannheden based on the code I'd say that should work, yes. What error do you see?

NB: there's a possibility that this indeed doesn't work, because I don't think combining multiple Refaster templates like this has been tested recently. While [`v2.3.4-picnic-2`](https://github.com/PicnicSupermarket/error-prone/commits/v2.3.4-picnic-2) does indeed contain this patch, it also contains a fix for #552, so the current integration within Picnic actually compiles many Refaster templates into one, and then only passes that one `.refaster` file to Error Prone.

(I believe a longer term fix entails a more user-friendly integration as described under the second bullet point in https://github.com/google/error-prone/issues/552#issuecomment-516054278, but I'll admit that while this works, I'm no closer to polishing and open-sourcing that code. Chronic lack of time.)

knutwannheden:
With my code being compiled through Maven I didn't really see any helpful error message at all, just some very generic message. I can report that back here, if you think it helps. Or maybe you know how I can get Maven to print the root cause here.

Stephan202:
Shot in the dark: add `<verbose>true</verbose>` to the `maven-compiler-plugin` configuration and see whether that helps. If not, run with `mvn -e -X` and see whether anything stands out. If still none the wiser, dig through the `-X` output to find the exact `javac` command executed and manually run it on the command line. This can be pretty fiddly, but I recall that at least once in the past this allowed me to see an error message that was somehow not bubbled up by the `maven-compiler-plugin`.

knutwannheden:
The error message wasn't displayed because of the `<fork>true</fork>` I had in my configuration (isn't the first time this happens to me) :unamused: 

So the error I get is `Can't load Refaster rule from /path/to/first.refaster,refaster:/path/to/second.refaster`. In the `pom.xml` file I have:

```xml
    <arg>-Xplugin:ErrorProne -XepPatchChecks:refaster:/path/to/first.refaster,refaster:/path/to/second.refaster -XepPatchLocation:${basedir}</arg>
```

Stephan202:
Okay, I stand corrected. This changeset is _not_ included in the fork. :man_facepalming: And then the error message makes perfect sense. Apologies! Can you try to compile your Refaster rules in one go, using the changes from #552? (Those _are_ included in the fork, I double-checked now :sweat_smile:.)

Otherwise I can try to create a new release that does include this change, but it might take a bit before I have time for that.

knutwannheden:
Thanks for the heads-up. For my use case (part of jOOQ, which I think you are familiar with :smile:) I actually want to have separate `.refaster` files, which I am building in Maven with separate executions of the Java compiler.

But since we are only in an exploratory phase, it isn't yet important to have the ability to use multiple `.refaster` files, so no need to make a new release for that.

Stephan202:
Cool! jOOQ Refaster templates... we're definitely interested in those! (Say hi to @lukaseder :) )

(Going off-topic big time here, but the idea that library authors provide migration or best-practice Refaster rules for their libraries is one of the motivations behind the stuff written under the second bullet point in https://github.com/google/error-prone/issues/552#issuecomment-516054278.)

lukaseder:
Hi @Stephan202

cushon:
Hi @Stephan202,

Thanks for the PR! I was taking a fresh look at this, it seems like a nice thing to support.

I was trying to think a bit about how this interacts with #947 and #4028, I wonder if we should decide what the semantics should be for #4028 first and then figure out how to make the refaster patch options consistent with that.

Issue #792, opened 2017-10-25
ParameterName fails in certain cases where two classes reside in one file


nicktrav:
In the case of multiple classes residing in the same .java file, say `Foo` (the public class) and `Bar` in the file `Foo.java`, if a second class `Baz`, residing in file `Baz.java` extends `Foo`, error-prone can fail in what appears to be the case when `Baz.java` is compiled and checked before `Foo.java`.

Foo.java:

```java
class Bar {
}

public class Foo {

  public void acceptInteger(Integer integer) {
  }

  public void callAccept() {
    acceptInteger(0);
  }
}
```

Baz.java:

```java
public class Baz extends Foo {
}
```

Specifically, the issue is in the `ParameterName` check, which is failing with an `IndexOutOfBoundsException` [here](https://github.com/google/error-prone/blob/v2.1.2/core/src/main/java/com/google/errorprone/bugpatterns/ParameterName.java#L85).

[Here's a link](https://github.com/nicktrav/error-prone/commit/a2be36b2bf563edd64d326cd4c57d8f62a190944) to a commit reproducing the issue.

Running the following test case:

```
mvn test -pl core -Dtest='com.google.errorprone.bugpatterns.MultipleClassWithExtendsTest'
```

leads to the following failure for one of the two tests:

```
Failed tests:
  MultipleClassWithExtendsTest.failingTest:40 An unhandled exception was thrown by the Error Prone static analysis plugin.
     Please report this at https://github.com/google/error-prone/issues/new and include the following:

     error-prone version: unknown version
     Stack Trace:
     java.lang.IndexOutOfBoundsException
        at java.nio.HeapCharBuffer.subSequence(HeapCharBuffer.java:580)
        at java.nio.HeapCharBuffer.subSequence(HeapCharBuffer.java:42)
        at com.google.errorprone.bugpatterns.ParameterName.checkArguments(ParameterName.java:85)
        at com.google.errorprone.bugpatterns.ParameterName.matchMethodInvocation(ParameterName.java:64)
        at com.google.errorprone.scanner.ErrorProneScanner.visitMethodInvocation(ErrorProneScanner.java:907)
        at com.google.errorprone.scanner.ErrorProneScanner.visitMethodInvocation(ErrorProneScanner.java:146)
        at com.sun.tools.javac.tree.JCTree$JCMethodInvocation.accept(JCTree.java:1644)
     ... snip ...
```

Affects v2.1.2, but also confirmed to exist at HEAD on master.

[...]

PR #795, opened 2017-10-26
Add a regression test for #792

nicktrav:
Adds a regression test to cover the case where one java file contains
two class declarations, and another class in a separate file extends the
public class in the first file.

This test would fail at v2.1.2, but was patched in 73e24807d.


nicktrav:
I signed it!

PR #801, opened 2017-11-01
Remove synchronizedList suggestion from LinkedList

gaul:
`LinkedList` does not provide synchronization so this recommendation is
incorrect.


cushon:
Thanks - I have a pending change to fix this. That sentence was supposed to be under `java.util.Stack` (which is synchronized).

cushon:
See bd9802edfbddaefc75d2256f4062c2ded9a3d2b6

PR #817, opened 2017-11-08
fixes #716 respect any @CanIgnoreReturnValue annotation

### LINKED ISSUE #716: error-prone doesn't follow relocated dependencies
I'm working on integrating error-prone into the compile for Apache HBase, and we're running into a strange false positive, I think.

Steps to reproduce:
```
git clone https://github.com/apache/hbase
cd hbase
mvn clean package -PerrorProne
```

Output error:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.6.1:compile (default-compile) on project hbase-common: Compilation failure
[ERROR] /Users/mdrob/IdeaProjects/hbase/hbase-common/src/main/java/org/apache/hadoop/hbase/io/crypto/Context.java:[89,31] [CheckReturnValue] Ignored return value of method that is annotated with @CheckReturnValue
[ERROR]     (see http://errorprone.info/bugpattern/CheckReturnValue)
[ERROR]   Did you mean to remove this line?
```

That line is a call to a relocated Guava 22 `Preconditions.checkNotNull` which is annotated `@CanIgnoreReturnValue` - this can be verified in the [artifact](https://repo1.maven.org/maven2/org/apache/hbase/thirdparty/hbase-shaded-miscellaneous/1.0.0/hbase-shaded-miscellaneous-1.0.0.jar) (No source release available yet, unfortunately!)

There may be older versions of guava on the dependency classpath, but this class specifically uses `import org.apache.hadoop.hbase.shaded.com.google.common.base.Preconditions;`

I don't yet have a smaller reproducible case. Before I start down that path, do we think the problem is most likely in error-prone, the plexus compiler, our configuration of the plugins, a combination of these, or something else entirely?

madrob:


madrob:
I signed it!

madrob:
@cushon WDYT?

PR #843, opened 2017-12-02
attempt to fix formatting issue with JdkObsolete docs

msridhar:
The footnote on the [JdkObsolete doc](http://errorprone.info/bugpattern/JdkObsolete) is formatted incorrectly.  This is an attempt to fix that by removing newlines.  Not sure if it worked; don't know how to test locally.  Is there any way to test this?

### DIFF
--- a/docs/bugpattern/JdkObsolete.md
+++ b/docs/bugpattern/JdkObsolete.md
@@ -57,11 +57,4 @@ An ancient precursor to `Iterator`.
 
 Replaced by `NavigableSet` and `NavigableMap` in Java 6.
 
-[^1]: People generally choose `LinkedList` because they want fast insertion and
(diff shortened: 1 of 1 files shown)


cushon:
It's possible but inconvenient. We update `gh-pages` using [`util/update-local-pages.sh`](https://github.com/google/error-prone/blob/263c4056cf6edd49490741ed08359439c3fa3dcc/util/update-local-pages.sh), and then you can run jekyll locally to test the changes. But I'm fine with making this change and seeing if it helps.

(And it looks like not all markdown parsers have trouble with the current formatting, but that's not a new problem...)

msridhar:
I don't have jekyll installed locally.  If you want to go ahead and merge to see if it works, I'm fine with that 😄

msridhar:
Sigh, doesn't seem to have worked 🙁

PR #872, opened 2017-12-19
Add String.split check, and make all existing code pass it

### LINKED ISSUE #867: proposed new check: String.split with only one argument
Consider the code at 
https://github.com/google/error-prone/blob/92254ee562f449c3aef387949094221b1ff8c0d1/check_api/src/main/java/com/google/errorprone/ErrorProneOptions.java#L241

If you pass -Xep:: (two colons) to that function, you'll get an ArrayIndexOutOfBoundsException. 

This is because String.split has a funny behavior inherited from Perl (see http://konigsberg.blogspot.com/2009/11/final-thoughts-java-puzzler-splitting.html ) 

I think String.split should only ever be called with two arguments (e.g. (arg, -1)), to prevent this gotcha.

novalis:
Fixes #867.  I decided not to do work to suggest the Guava Splitter.  That could always be added later, and String.split is just fine in its two-argument form.

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/StringSplit.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/StringSplit.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2017 The Error Prone Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.errorprone.bugpatterns;
+
+import static com.google.common.collect.Iterables.getOnlyElement;
+import static com.google.errorprone.BugPattern.Category.JDK;
+import static com.google.errorprone.BugPattern.SeverityLevel.WARNING;
+import static com.google.errorprone.matchers.Description.NO_MATCH;
+import static com.google.errorprone.matchers.method.MethodMatchers.instanceMethod;
+
+import com.google.errorprone.BugPattern;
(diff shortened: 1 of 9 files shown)


core/src/main/java/com/google/errorprone/bugpatterns/StringSplit.java
```diff
@@ -0,0 +1,54 @@
+/*
+ * Copyright 2017 Two Sigma Open Source. All Rights Reserved.
```
cushon: This should be

```
Copyright 2017 The Error Prone Authors.
```

You're welcome to add Two Sigma to https://github.com/google/error-prone/blob/master/AUTHORS. (There's some documentation on copyright and AUTHORS files here: https://opensource.google.com/docs/releasing/authors/)

core/src/main/java/com/google/errorprone/bugpatterns/StringSplit.java
```diff
@@ -0,0 +1,54 @@
...
+      instanceMethod().onDescendantOf("java.lang.String").withSignature("split(java.lang.String)");
+
+  @Override
+  public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+    if (!MATCHER.matches(tree, state)) {
+      return NO_MATCH;
+    }
+    List<? extends ExpressionTree> arguments = tree.getArguments();
+    return describeMatch(
+        tree, SuggestedFix.builder().postfixWith(arguments.get(0), ", -1").build());
```
cushon: consider `Iterables.getOnlyElement` instead of `get(0)` to make the assumption about the number of arguments explicit.

core/src/main/java/com/google/errorprone/bugpatterns/StringSplit.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.List;
+
+/** @author dturner@twosigma.com (David Turner) */
+@BugPattern(
+  name = "StringSplit",
+  category = JDK,
+  summary = "String.split should never take only a single argument; it has surprising behavior",
+  severity = WARNING
+)
+public class StringSplit extends BugChecker implements MethodInvocationTreeMatcher {
```
cushon: The check needs to be added to this list to get wired up: https://github.com/google/error-prone/blob/2623f228cece3a24fd7ccb77be478c4e81f2883d/core/src/main/java/com/google/errorprone/scanner/BuiltInCheckerSuppliers.java#L525

PR #874, opened 2017-12-20
Moe sync

ronshapiro:


core/src/main/java/com/google/errorprone/bugpatterns/TypeParameterNaming.java
```diff
@@ -0,0 +1,138 @@
...
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import javax.lang.model.element.Name;
+
+/** Enforces type parameters match the google style guide. */
+@BugPattern(
+  name = "TypeParameterNaming",
+  summary =
+      "Type Parameters must be a single letter with a numeric suffix,"
```
jbduncan: @nglorioso @ronshapiro I think that this portion of the sentence is a bit misleading, as it seems to suggest that the numeric suffix is mandatory and can be of any length, whereas the guide seems to say the numeric suffix is optional and should only be one digit long if used. What do you both think?

ronshapiro: I didn't write it (just synced) but I think this should read:
```java
"Type Parameters must be a single letter with an optional numeric suffix"
```

jbduncan: @ronshapiro Sounds a very good idea to me! I'd be perfectly happy with that rewording.

PR #877, opened 2017-12-21
Moe sync

ronshapiro:


cushon:
CLAs were signed in #406 and #872.

Issue #873, opened 2017-12-20
2.1.0 no more supports encodings other than UTF-8


lasselindqvist:
Example command line:
java -Xbootclasspath/p:error_prone_ant-2.1.0.jar com.google.errorprone.ErrorProneCompiler Constants.java

Below is an example of an ANSI file which produces an error with version 2.1.0 but not with 2.0.9.
...
package com.constants;

/**
 * Mikäli
 */
public class Constants {	
	public final static String SALA = "SALA";
	public final static String SÄLÄ = "SÄLÄ";
	public final static String SÄLÄ = "SALA";
	public final static String SALA = "SÄLÄ";
}
...

Constants.java:4: error: unmappable character (0xE4) for encoding UTF-8
 * Mik?li
      ^
Constants.java:8: error: unmappable character (0xC4) for encoding UTF-8
        public final static String S?L? = "S?L?";
                                    ^
Constants.java:8: error: unmappable character (0xC4) for encoding UTF-8
        public final static String S?L? = "S?L?";
                                      ^
Constants.java:8: error: unmappable character (0xC4) for encoding UTF-8
        public final static String S?L? = "S?L?";
                                            ^
Constants.java:8: error: unmappable character (0xC4) for encoding UTF-8
        public final static String S?L? = "S?L?";
                                              ^
Constants.java:9: error: unmappable character (0xC4) for encoding UTF-8
        public final static String S?L? = "SALA";
                                    ^
Constants.java:9: error: unmappable character (0xC4) for encoding UTF-8
        public final static String S?L? = "SALA";
                                      ^
Constants.java:10: error: unmappable character (0xC4) for encoding UTF-8
        public final static String SALA = "S?L?";
                                            ^
Constants.java:10: error: unmappable character (0xC4) for encoding UTF-8
        public final static String SALA = "S?L?";
                                              ^
Constants.java:8: error: illegal character: '\ufffd'
        public final static String S?L? = "S?L?";
                                    ^
Constants.java:8: error: illegal character: '\ufffd'
        public final static String S?L? = "S?L?";
                                      ^
Constants.java:9: error: illegal character: '\ufffd'
        public final static String S?L? = "SALA";
                                    ^
Constants.java:9: error: illegal character: '\ufffd'
        public final static String S?L? = "SALA";
                                      ^
4 errors

lasselindqvist:
[...]

PR #912, opened 2018-01-26
Support other file encodings than UTF-8

pskiwi:
The source file encoding _UTF-8_ is hard coded into [MaskedClassLoader.java](https://github.com/google/error-prone/blob/71f27667eb723c18e574f905e0fe3a64786a2c3c/check_api/src/main/java/com/google/errorprone/MaskedClassLoader.java#L67). 
This makes it impossible to use error-prone for otherwise encoded sources.
This MR leaves the default to _UTF-8_ and only changes it if the parameter "-encoding" is given on the command line.

This problem ist reported in #873  and mentioned in https://groups.google.com/d/msg/error-prone-discuss/075NR0aTzm4/LsHe6t7BBgAJ.

### DIFF
--- a/check_api/src/main/java/com/google/errorprone/BaseErrorProneCompiler.java
+++ b/check_api/src/main/java/com/google/errorprone/BaseErrorProneCompiler.java
@@ -18,7 +18,6 @@ package com.google.errorprone;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
-import com.google.common.base.Charsets;
 import com.google.common.collect.ImmutableList;
 import com.google.errorprone.MaskedClassLoader.MaskedFileManager;
 import com.google.errorprone.scanner.ScannerSupplier;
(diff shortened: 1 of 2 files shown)


pskiwi:
I signed it!

pskiwi:
I signed it!

jbduncan:
@pskiwi I think that the email address you used to sign your commits is different to your GitHub email address. The reason I think this is because if you used the same email address in both places, you'd see your GitHub username under each commit, in a similar fashion to this:

capture2

However, in this case, we just see your real name instead:

capture

Have you checked and updated your Git settings as per [the instructions in the Git book](https://git-scm.com/book/en/v2/Getting-Started-First-Time-Git-Setup) to make sure that the email address you registered on your local installed copy of Git is the same as your GitHub email address?

jbduncan:
Oops, never mind, too quick for my own good! Looks like you fixed things on your end, so please feel free to disregard my earlier comment. :)

pskiwi:
I will start a new MR with fixed email addresses.

PR #926, opened 2018-02-07
Add option to ignore missing @Override annotations in interfaces

msridhar:
In the [mobile RIB architecture](https://github.com/uber/RIBs), we often have cases where Dagger Component-related interfaces form an inheritance chain, such that a method gets added to a Component super-interface that was already present in the Component sub-interface.  In such a case, the EP MissingOverride check forces an `@Override` annotation to be added to the sub-interface.  Forcing this change goes against some of the modularity in the RIB architecture (@AttwellBrian can explain more).  Further, in this case the annotation does not really add safety, as (I think) Dagger code gen will complain if there is some ambiguity in the set of methods exposed by the component sub-interface.

In general, it's not clear whether forcing `@Override` annotations in interfaces is as much of a win as it is for classes.  (Kotlin forces it on interfaces, but Swift only requires it on classes, not sub-protocols.)  This PR adds an option to the MissingOverride check to disable checking for `@Override` annotations in interfaces.

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/MissingOverride.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/MissingOverride.java
@@ -48,12 +48,12 @@ import javax.lang.model.element.Modifier;
 public class MissingOverride extends BugChecker implements MethodTreeMatcher {
 
   /**
-   * if true, don't warn on missing @Override annotations inside interfaces
+   * if true, don't warn on missing {@code @Override} annotations inside interfaces
    */
   private final boolean ignoreInterfaceOverrides;
 
   public MissingOverride() {
-    this.ignoreInterfaceOverrides = false;
+    this(ErrorProneFlags.empty());
   }
 
   public MissingOverride(ErrorProneFlags flags) {
--- a/core/src/main/java/com/google/errorprone/bugpatterns/MissingOverride.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/MissingOverride.java
@@ -22,6 +22,7 @@ import static com.google.errorprone.BugPattern.SeverityLevel.WARNING;
 import com.google.errorprone.BugPattern;
 import com.google.errorprone.BugPattern.ProvidesFix;
 import com.google.errorprone.BugPattern.StandardTags;
+import com.google.errorprone.ErrorProneFlags;
 import com.google.errorprone.VisitorState;
 import com.google.errorprone.bugpatterns.BugChecker.MethodTreeMatcher;
 import com.google.errorprone.fixes.SuggestedFix;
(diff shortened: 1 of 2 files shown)


core/src/test/java/com/google/errorprone/bugpatterns/MissingOverrideTest.java
```diff
@@ -89,4 +90,32 @@ public void deprecatedMethod() throws Exception {
...
+        .addSourceLines(
+            "Test.java",
+            "public interface Test extends Super {",
+            "  // BUG: Diagnostic contains: f implements method in Super; expected @Override",
+            "  void f();",
+            "}")
+        .doTest();
+  }
+
+  // this doesn't work!  need new APIs in CompilationTestHelper
```
msridhar: Would appreciate some advice on how to test.  The most obvious route I could was to add APIs to `CompilationTestHelper` to allow for passing in custom `ErrorProneFlags`.  But I wanted to check if there is an easier way before going ahead with that.

core/src/main/java/com/google/errorprone/bugpatterns/MissingOverride.java
```diff
@@ -46,6 +47,20 @@
 )
 public class MissingOverride extends BugChecker implements MethodTreeMatcher {
 
+  /**
+   * if true, don't warn on missing @Override annotations inside interfaces
+   */
+  private final boolean ignoreInterfaceOverrides;
+
+  public MissingOverride() {
+    this.ignoreInterfaceOverrides = false;
```
Stephan202: You could keep the default (i.e., `false`) in one place by replacing this constructor implementation with `this(ErrorProneFlags.empty());`.

core/src/test/java/com/google/errorprone/bugpatterns/MissingOverrideTest.java
Stephan202: Hmm, why doesn't this work? In some private code I do the following, which does work:
```java
@RunWith(JUnit4.class)
public final class LexicographicalAnnotationAttributeListingCheckTest {
  ...
  private final CompilationTestHelper restrictedCompilationTestHelper =
      CompilationTestHelper.newInstance(
              LexicographicalAnnotationAttributeListingCheck.class, getClass())
          .setArgs(
              ImmutableList.of(
                  "-XepOpt:LexicographicalAnnotationAttributeListing:Includes=pkg.A.Foo,pkg.A.Bar",
                  "-XepOpt:LexicographicalAnnotationAttributeListing:Excludes=pkg.A.Bar#value"));
  ...
  @Test
  public void testFiltering() {
    restrictedCompilationTestHelper
        .addSourceLines(...)
        .doTest();
  }
}
```

core/src/main/java/com/google/errorprone/bugpatterns/MissingOverride.java
```diff
@@ -46,6 +47,20 @@
 )
 public class MissingOverride extends BugChecker implements MethodTreeMatcher {
 
+  /**
+   * if true, don't warn on missing @Override annotations inside interfaces
```
Stephan202: Maybe write `{@code @Override}`, so as not to confuse the javadoc tool?

msridhar:
@Stephan202 thanks for the review!  And you were right about the tests; I must have been invoking maven incorrectly before.

msridhar:
@eaftan any thoughts on this one?

msridhar:
@eaftan another bump on this one, if you have time

msridhar:
@cushon should I open a separate PR to update the docs?

PR #930, opened 2018-02-12
Add FieldCanBeStatic checker

gaul:
This finds final fields which are initialized with a literal.  Adding
static reduces memory overhead from per-instance to per-class and
allows initialization at compile-time instead of run-time.

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/FieldCanBeStatic.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/FieldCanBeStatic.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright 2018 The Error Prone Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.errorprone.bugpatterns;
+
+import static com.google.errorprone.BugPattern.SeverityLevel.SUGGESTION;
+import static com.google.errorprone.fixes.SuggestedFixes.addModifiers;
+import static com.google.errorprone.matchers.Description.NO_MATCH;
+import static com.google.errorprone.matchers.Matchers.isPrimitiveType;
+import static com.google.errorprone.util.ASTHelpers.getType;
+import static com.google.errorprone.util.ASTHelpers.isSubtype;
+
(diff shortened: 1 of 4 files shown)


core/src/test/java/com/google/errorprone/bugpatterns/FieldCanBeStaticTest.java
```diff
@@ -0,0 +1,62 @@
+/*
+ * Copyright 2017 The Error Prone Authors.
```
jbduncan: s/2017/2018 ?

core/src/main/java/com/google/errorprone/bugpatterns/FieldCanBeStatic.java
```diff
@@ -0,0 +1,84 @@
...
+import com.sun.source.tree.VariableTree;
+import com.sun.tools.javac.code.Symbol.VarSymbol;
+import javax.lang.model.element.ElementKind;
+import javax.lang.model.element.Modifier;
+
+/** @author andrew@gaul.org (Andrew Gaul) */
+@BugPattern(
+  name = "FieldCanBeStatic",
+  category = JDK,
+  summary = "A final field with an immutable type can be static.",
```
jbduncan: I wonder if this summary would explain its corresponding bug checker's purpose better with a description like the following?

> A final field **initialized at compile-time with an instance of an immutable type** can be static.

core/src/test/java/com/google/errorprone/bugpatterns/FieldCanBeStaticTest.java
```diff
@@ -0,0 +1,62 @@
...
+            "  static {",
+            "    staticFinalInitializer = \"string\";",
+            "  }",
+            "  private static final String staticFinal = \"string\";",
+            "  private int nonFinal = 3;",
+            "  private static int staticNontFinal = 4;",
+            "  private final Object finalMutable = new Object();",
+            "  private final int nonLiteral = new java.util.Random().nextInt();",
+            "}")
+        .doTest();
```
jbduncan: @gaul I think it's worth adding another negative case where the `final` fields are initialized in a constructor, like one would with a POJO. So something like this:

```java
class Person {
  final String name;
  final int age;

  Person(String name, int age) {
    this.name = name;
    this.age = age;
  }
}
```

WDYT?

core/src/main/java/com/google/errorprone/bugpatterns/FieldCanBeStatic.java
gaul: Done.

core/src/test/java/com/google/errorprone/bugpatterns/FieldCanBeStaticTest.java
gaul: Done.

core/src/test/java/com/google/errorprone/bugpatterns/FieldCanBeStaticTest.java
gaul: Done.

davidhart82:
Any chance this merge conflict can be resolved?

PR #947, opened 2018-02-21
Support running all available patch checks

Stephan202:
The `BaseErrorProneJavaCompiler` [supports](https://github.com/google/error-prone/blob/master/check_api/src/main/java/com/google/errorprone/BaseErrorProneJavaCompiler.java#L225) running _all_ patch checks; it just requires that one doesn't provide any explicitly named checkers.

In practice this doesn't work, however, because:
- `PatchingOptions` [requires](https://github.com/google/error-prone/blob/master/check_api/src/main/java/com/google/errorprone/ErrorProneOptions.java#L142) a named checker or a custom refactorer.
- Specifying `-XepPatchChecks:` without any explicit check will [cause](https://github.com/google/error-prone/blob/master/check_api/src/main/java/com/google/errorprone/ErrorProneOptions.java#L404) the empty string to be stored as a named checker.

This PR proposes to resolve the above issues by:
1. Requiring `-XepPatchLocation` when `-XepPatchChecks` is specified, but not vice versa.
2. Omitting empty strings from `-XepPatchChecks:`.

### DIFF
--- a/check_api/src/main/java/com/google/errorprone/ErrorProneOptions.java
+++ b/check_api/src/main/java/com/google/errorprone/ErrorProneOptions.java
@@ -137,21 +137,7 @@ public class ErrorProneOptions {
 
       abstract Builder importOrganizer(ImportOrganizer importOrganizer);
 
-      abstract PatchingOptions autoBuild();
-
-      final PatchingOptions build() {
-
-        PatchingOptions patchingOptions = autoBuild();
-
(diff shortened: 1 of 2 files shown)


Stephan202:
Rebased the branch; would still be interested in seeing this feature included upstream.

cushon:
(@Stephan202 thanks for the PR, and for reminding me about this one in #4028)

cushon (approved):

PR #950, opened 2018-02-22
Relocate com.google classes in Ant artifact, excluding Error Prone classes

msridhar:
Context: we have Error Prone integrated with Buck, leveraging the `error_prone_ant` shaded jar and Buck's [javac_jar](https://buckbuild.com/concept/buckconfig.html#tools.javac_jar) setting.  One issue with the integration is that if annotation processors use a more recent version of Guava than what EP is using, the version in the shaded jar wins.  This change relocates all `com.google` classes in the EP jar except for the `com.google.errorprone` classes (which need to be accessed by plugin checks).
This change should prevent the most common conflicts with annotation processors and EP plugin checks.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -87,7 +87,7 @@
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-shade-plugin</artifactId>
-          <version>2.1</version>
+          <version>3.1.0</version>
           <configuration>
             <dependencyReducedPomLocation>${project.build.directory}/dependency-reduced-pom.xml</dependencyReducedPomLocation>
           </configuration>
--- a/ant/pom.xml
+++ b/ant/pom.xml
@@ -66,6 +66,15 @@
                   <exclude>org.apache.ant:ant-launcher</exclude>
                 </excludes>
               </artifactSet>
+              <relocations>
+                <relocation>
+                  <pattern>com.google</pattern>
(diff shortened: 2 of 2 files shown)


pom.xml
```diff
@@ -87,7 +87,7 @@
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-shade-plugin</artifactId>
-          <version>2.1</version>
+          <version>3.1.0</version>
```
msridhar: I had to bump this dependence; with version 2.1, I got a crash in ASM during the class renaming process.

msridhar:
Travis failure looks transient; all tests passed locally for me.

msridhar:
Actually, now I am wondering if there are any APIs that an EP plugin check would use that refer to a Guava type.  Is this possible?  We haven't run into it yet, but if there is such an API, I'm not sure how to fix, besides sharing the Guava classes between the EP plugins and EP itself.

ronshapiro:
> Actually, now I am wondering if there are any APIs that an EP plugin check would use that refer to a Guava type.

This definitely happens.

Also, it seems to me like this is probably better served outside of the default EP configuration.

msridhar:
Ok, if Guava types are exposed to EP plugins then this relocation is a bad idea.

@ronshapiro so the only solution I see then is for individual annotation processors to package a shaded / relocated version of Guava if they need something newer than what EP uses.  If you have other suggestions I'd be interested.

Issue #649, opened 2017-06-24
Availability of Google's Refaster templates?


JnRouvignac:
[Refaster's documentation](http://errorprone.info/docs/refaster) is great! Thanks for that.

It mentions the following:

> At Google, we have a tool that uses thousands of these simplification patterns to automatically comment on code reviews, suggesting improvements to code. We hope to release many of these soon.

I am the author of JnRouvignac/AutoRefactor.
I am very interested by Refaster as a faster approach to write refactoring rules.
However I do not want to waste time trying to contribute rules that already exist at Google.

What are your plans for open sourcing your Refaster rules?
Do you have an expected timeframe?

JnRouvignac:
Any idea @eaftan @nglorioso @cushon ?

jbduncan:
@eaftan @nglorioso @cushon I'm interested in this as well. Are there any plans for this?

eaftan:
Assigned to @lowasser.  He can comment further on timeframe but it might be a while before he can get to it.  Let's say sometime in Q4.

JnRouvignac:
Hello again @eaftan @nglorioso @cushon @lowasser, Q4 is behind us :)
I am checking if you made some progress on this issue?

eaftan:
@lowasser is actively working on this.

JnRouvignac:
Excellent news!

@lowasser Do you have an ETA for this, or rather a date at which I can ping you back if I see no progress?

Let me know if I can be of any help.

JnRouvignac:
Hello @eaftan @lowasser !

This is my periodical checking for progress :)
Did you have a chance to complete this item?

Thanks!

cal101:
Hello @eaftan @lowasser!

I would be interested that googles refaster templates are made public. Is there any progress on this issue?

eaftan:
@cal101 I no longer work at Google, sorry!

JnRouvignac:
@lowasser @cpovirk, by any chance, would you have an update to provide on this?

JnRouvignac:
I meant to also ask @cushon

cushon:
I unfortunately don't have any updates: this is something I'd like to see happen, but I can't promise when we'll find time to get to it.

PR #951, opened 2018-02-26
Use valueOf method instead of constructor since valueOf is likely to …

reudismam:
…yield significantly better space and time performance by caching frequently requested values.


lowasser:
Related to #649.  This is a best practices/optimization issue, not a correctness issue, so we group it separately from Error Prone internally (and already have this check, just not open sourced yet).

eaftan:
Retrying the AppVeyor CI build.  The change looks fine to me.  Note that the code modified was in test inputs, not actual Error Prone code.

PR #952, opened 2018-02-26
Use equals method of string literal to prevent NPE.

reudismam:


lowasser:
Why would you not want an NPE?  When is it appropriate to pass null to these methods?

jbduncan:
I can't speak for the error-prone team, but I'd argue that this is not such a good change to make, because if `null`s are allowed to pass through like this, it would then make debugging much harder if they cause `NullPointerException`s to be thrown in other, seemingly unrelated parts of the code base. If the `NullPointerException`s get caught sooner during a run of error-prone, it then becomes easier to identify where wrong assumptions were made, and to fix those areas of the code base accordingly.

Just my 2 cents. :)

eaftan:
I agree that in this case we do want the NPEs as these methods should not accept null, and we would want to get NPEs if anyone does pass null to them.

@lorisdanto and @reudismam, thanks for the contribution and let me know if you want more explanation.  In general this is a good suggestion, but not in this specific case.

PR #957, opened 2018-03-05
add endCompilationUnit callback for BugChecker

msridhar:
This PR adds a new interface `BugChecker.CompilationUnitTreeEndMatcher` that can be implemented by Error Prone checkers.  Its `endCompilationUnit` method gets invoked after the entire tree of a compilation unit has been visited, and it can return a `Set<Description>` to report errors over the file.

We've found this functionality to be quite useful in a couple of internal checkers.  If this functionality were available in open-source Error Prone we would probably also update NullAway to use it.  I've included a sample `UnusedPrivateFields` check to show how it could be useful.  This check is primarily included for exemplary purposes; we haven't enabled this check at the moment.

As to why `endCompilationUnit` but not all the other tree nodes: it seemed like the minimal API change to enable the functionality we wanted.  Also, compilation units are not nested; an `endClass` callback (for example) might be more error prone in the presence of inner / anonymous classes.


eaftan:
@nglorioso and I discussed this, and we're not comfortable with adding mutable state to BugCheckers.  If you want to collect state from a whole compilation unit, you could implement a `TreeScanner` and then scan the compilation unit from `matchCompilationUnit()`.  There's an example of this in MockitoCast: https://github.com/google/error-prone/blob/master/core/src/main/java/com/google/errorprone/bugpatterns/MockitoCast.java#L100

msridhar:
@eaftan thanks for the feedback and the pointer to the MockitoCast check.  I didn't know about some of the `VisitorState` APIs used there, pretty handy.  Our main concern about the `TreeScanner` approach was performance.  If we have several checks each doing their own AST traversals, we thought that could eventually add up to a significant performance overhead compared to piggy-packing on Error Prone's AST traversal.  I don't have numbers to back this up, though.  Do you have any thoughts on whether this is a real concern or not?

msridhar:
@cushon I'm tinkering around with the `TaskListener` approach.  From what I see, for the check to observe the finish of `ANALYZE` events it needs to be registered as a listener before the `ANALYZE` phase starts to run.  Further, its listener should run after the `ErrorProneAnalyzer` listener.  I don't see any obvious way for an Error Prone plugin check to do accomplish this based on current APIs.  I don't think it'll work to try to report errors at the start of `GENERATE`, and even if it does, it seems gross.  Any suggestions?

msridhar:
I haven't yet figured out a way to get the `TaskListener` approach to work.  Another thought / proposal here.  Rather than having a fairly unrestricted `CompilationUnitTreeEndMatcher` interface, we could have an abstract class `WholeCompilationUnitChecker` that relevant checks need to extend.   `WholeCompilationUnitChecker` could then explicitly manage the lifecycle of any relevant (mutable) state via a `CompilationUnitState` interface.  Subclasses would be required to define a method `CompilationUnitState create()`, and `WholeCompilationUnitChecker` would be responsible for calling `create()` at the beginning of a compilation unit and nulling out the pointer at the end.  This would strongly encourage limiting the scope of mutable state.  It would not completely prevent checks from keeping mutable state elsewhere, but currently Error Prone doesn't do that anyway.

@eaftan @nglorioso if the above approach sounds more promising let me know and I can update the code in this PR for further discussion.  We have some other checks we'd like to upstream, but we are hesitant to adopt the pattern of every check doing its own AST traversal due to performance.

msridhar:
@eaftan @nglorioso any update / thoughts on the above?  We have some other checks we'd like to open source / upstream that are blocked on getting a resolution here.

eaftan:
Can you try the TreeScanner/matchCompilationUnit approach and see if the performance is acceptable?  We would really like to avoid mutable state in our BugCheckers, and we would want convincing numbers before going down that path.

msridhar:
Ok, finally had some time to dig into performance.  I wrote a simple Error Prone check that just does a configurable number of AST traversals; code is [here](https://gist.github.com/msridhar/9e5d92774bc1bbe9aa4c589533922b61).  I then compiled a large module (~2300 Java source files) running only this check, and varying the number of additional traversals it does.  Results (measured on my laptop in a fairly non-rigorous way):

| # Traversals | Running Time |
| ------------ | -------------- |
|             0       |    28s               |
|         10         |    28.4s            |
|        100        |    28.8s            |
|        250       |      32.5s           |
|        500       |      36.6s           |
|       1,000     |       45.7s          |
|       2,000     |      64.5s           |

Any difference less than 2 seconds is in the noise for these runs.  So, it looks like there needs to be more than 100 extra traversals going on to add anything measurable to compile time on my machine.  At the larger number of passes, the overhead starts looking more linear, and each pass is adding somewhere between 0.015 and 0.02 seconds of compile time (again, for ~2300 source files).  I'm not going to speculate on percentage overheads as that will depend a lot on annotation processors, etc.

Bottom line, I think the cost of extra traversals is low, but in the medium to long term maybe not a cost that should be completely ignored.  As #994 shows, these costs can add up.  That said, there is also a maintenance / readability cost to writing checks with mutable state and I can understand the desire to not encourage it.  I'll close this PR for now, but we may want to revisit down the line.

PR #964, opened 2018-03-15
Moe Sync

ronshapiro:

Commits:
=====
<p> Make ImmutableAnalysis.checkForImmutability() public.

RELNOTES: N/A

6fa9e9f815bd4fb96f052778bd18730c239eedae

-------

<p> Manually regenerate test protos.

RELNOTES: N/A

0d2216ea8c6f29722d59ab019119cb7660866062

-------

<p> Adding checker for inconsistent capitalization under the same scope for variables.

RELNOTES: Checker for inconsistent capitalization under the same scope for variables.

844e4401495ed2cc3144870c977f317f89f2b0cc

-------

<p> Update documentation on ASTHelpers.getAnnotation, pointing out the danger of using it with Annotations that contain Class members.

RELNOTES: n/a

407c1a2420459d05cfe2d3362147b234d3447d53

-------

<p> Extend SizeGreaterThanOrEqualsZero check to check proto repeated fields.

Rolled forward from d4e90c3e7185eae3dba61c93cf4ff37015856df0 but eliminating false positives caused by proto non-repeated field names ending in "count".

RELNOTES: Extend SizeGreaterThanOrEqualsZero check to check proto repeated fields.

84bf55a603c254585f28b6d3f1e4fed517b18dda

-------

<p> Adding test for listSuperClasses() method in ASTHelpers where class is in
default package.

RELNOTES: None.

57e88d2dfc4836c1333c30ea248b16b8ce6bc793

-------

<p> Add ContiguousSet to the list of known-immutable classes

RELNOTES: n/a

2e2a9a1bce40a73002d355386c200a511c6fda82


core/src/main/java/com/google/errorprone/bugpatterns/SizeGreaterThanOrEqualsZero.java
```diff
@@ -242,40 +238,48 @@ private Description provideReplacementForStaticMethodInvocation(
...
-                public CharSequence apply(ExpressionTree expressionTree) {
-                  return state.getSourceForNode((JCTree) expressionTree);
-                }
-              });
+          callToSize
+              .getArguments()
+              .stream()
+              .map(state::getSourceForNode)
+              .collect(toImmutableList());
       String argumentString = Joiner.on(',').join(argumentSourceValues);
```
jbduncan: @epmjohnston I believe that this portion of the code could be shortened and made a bit more efficient with something like:

```java
String argumentString =
  callToSize
    .getArguments()
    .stream()
    .map(state::getSourceForNode)
    .collect(joining(","));
```

PR #970, opened 2018-03-20
False positive FutureReturnValueIgnored in java Standard Library 

### LINKED ISSUE #883: False positive FutureReturnValueIgnored in java Standard Library
From the javadocs of [ScheduledExecutorService#scheduleAtFixedRate](https://docs.oracle.com/javase/9/docs/api/java/util/concurrent/ScheduledExecutorService.html#scheduleAtFixedRate-java.lang.Runnable-long-long-java.util.concurrent.TimeUnit-):
> ##### Returns:
>a ScheduledFuture representing pending completion of the series of repeated tasks. The future's get() method will never return normally, and will throw an exception upon task cancellation or abnormal termination of a task execution.

So perhaps an exception for `ScheduledFuture` Futures would be nice?

madrob:
fixes #883

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/FutureReturnValueIgnored.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/FutureReturnValueIgnored.java
@@ -64,6 +64,7 @@ import java.util.Queue;
 import java.util.Set;
 import java.util.concurrent.CompletionService;
 import java.util.concurrent.ForkJoinTask;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.regex.Pattern;
 import javax.lang.model.type.TypeKind;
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/google/errorprone/bugpatterns/FutureReturnValueIgnored.java
```diff
@@ -110,6 +111,11 @@
           instanceMethod()
               .onDescendantOf("io.netty.channel.ChannelFuture")
               .withNameMatching(Pattern.compile("addListeners?")),
+          // ScheduledExecutorService.scheduleAtFixedRate represents a series of tasks where it
+          // doesn't make sense to check for the return value
+          instanceMethod()
+              .onDescendantOf(ScheduledExecutorService.class.getName())
+              .withNameMatching(Pattern.compile("schedule(At|With)Fixed(Rate|Delay)")),
```
Stephan202: :thinking: but the `Future` can be used to cancel the scheduling. (Though most users perhaps shut down the whole service instead. I wonder which of those approaches prevails within the Google codebase.)

madrob: The future can be kept and used for things, of course, but in the specific case of `SES.scheduleAtFixedRate` it's arguably not a mistake to discard the return value. Quoting from the original issue filed:

> The future's get() method will never return normally

So unless you explicitly plan to cancel the future at some point, there's no reason to keep it around. You may want to keep it around to monitor the task, I guess, and restart it if it fails?

cushon: I left a note in https://github.com/google/error-prone/issues/883#issuecomment-1724432418

cushon:
Closing based on https://github.com/google/error-prone/issues/883#issuecomment-1724432418

PR #997, opened 2018-04-18
Fix Travis build

eaftan:
Context: https://github.com/travis-ci/travis-ci/issues/9512

Turns out we don't need to explicitly install JDK8 anymore, it's part of the image.

Tested: https://travis-ci.org/google/error-prone/jobs/368277832

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,16 +15,6 @@ matrix:
 # use travis-ci docker based infrastructure
 sudo: false
 
-# TODO(eaftan): uncomment addons below when
-# https://github.com/travis-ci/travis-ci/issues/9512 is fixed
-#
-# https://github.com/travis-ci/travis-ci/issues/3259#issuecomment-130860338
-#addons:
-#  apt:
-#    packages:
-#
-#      - oracle-java8-installer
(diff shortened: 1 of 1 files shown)


tbroyer:
> Turns out we don't need to explicitly install JDK8 anymore, it's part of the image.

IIRC, explicitly installing it ensured you had the latest version rather than the one built into the image. This is critical for those who deploy right from Travis I guess, but probably not for Error Prone.

cushon:
Do you know which version they are providing? In the past they have been slow to update.

It probably doesn't matter to much for 8 at this point--we ran in to some bugs that were fixed in the first few update releases, but it's had a while to stabilize.

eaftan:
@cushon it's 1.8.0_151-b12, which was released on [10/17/2017](http://www.oracle.com/technetwork/java/javase/8u151-relnotes-3850493.html)

eaftan:
The upstream bug has been fixed, so I can drop this if you want.

eaftan:
ping @cushon

cushon:
LGTM, sorry for the delay.

PR #998, opened 2018-04-18
Moe Sync

ronshapiro:

Commits:
=====
<p> Add FieldCanBeFinal - an optional check to identify fields that are only
assigned within a constructor.

RELNOTES: New optional (disabled by default) check: FieldCanBeFinal - to
identify fields that could be made final as they are only assigned once in
constructors or field initializers.

GOOGLE: moving it from the refaster directory and leaving it as a refaster
cleanup.

bf52036d2d3f5c311d930f6d39f520095c785cd8

-------

<p> Flag toString() on Objects without toString() overrides and final type

RELNOTES: Flag toString() on Objects without toString() overrides and final type

b9ea0177d57a88ac8abe898a3985c32f9751e2cb

-------

<p> Fix bug in comment parser when there is a comment on a nested method call.

RELNOTES: Fix bug in comment parser when there is a comment on a nested method call.

22e8298e3c82f250bb7d2fe0db581a8ddd839b39

-------

<p> Change cast in SuggestedFixes.compilesWithFix.

BasicJavacTask is a superclass of JavacTaskImpl, and provides the necessary getContext() method.

RELNOTES: n/a

736928cc0358e3031c4045d2cff3648b18b20696

-------

<p> Add a check to discourage assertThrows with block lambdas with >1 statement

RELNOTES: [AssertThrowsMultipleStatements] The lambda passed to assertThows should contain exactly one statement

b3442865df98d2252e0e35607652f89985f329a1

-------

<p> Make OrphanedFormatString detect instances where format string literals are
passed to @FormatMethod methods in the right position.

RELNOTES: OrphanedFormatString detects string literals correctly passed to
@FormatMethod methods

df39c21f6637cc4b985d17690ec074ef6ca04d14

-------

<p> Update examples for 2.3.0 release.

1ca0d1c0f33b8f453ac29b0a69d05abed1fe7783

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/AssertThrowsMultipleStatements.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/AssertThrowsMultipleStatements.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2018 The Error Prone Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.errorprone.bugpatterns;
+
+import static com.google.common.collect.Iterables.getLast;
+import static com.google.errorprone.matchers.Description.NO_MATCH;
+import static com.google.errorprone.matchers.method.MethodMatchers.staticMethod;
+
+import com.google.errorprone.BugPattern;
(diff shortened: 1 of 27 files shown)


ronshapiro:
All of the authors are googlers

core/src/main/java/com/google/errorprone/bugpatterns/AssertThrowsMultipleStatements.java
```diff
@@ -0,0 +1,85 @@
...
+@BugPattern(
+    name = "AssertThrowsMultipleStatements",
+    summary = "The lambda passed to assertThows should contain exactly one statement",
+    severity = SeverityLevel.WARNING,
+    providesFix = ProvidesFix.REQUIRES_HUMAN_ATTENTION)
+public class AssertThrowsMultipleStatements extends BugChecker
+    implements MethodInvocationTreeMatcher {
+
+  private static final Matcher<ExpressionTree> MATCHER =
+      staticMethod().onClass("org.junit.Assert").named("assertThrows");
```
jbduncan: @cushon I believe that this could be extended to JUnit 5's `org.junit.jupiter.api.Assertions.assertThrows` at some point in the future. WDYT?

PR #1007, opened 2018-04-21
Allow javac options to have directories that end in .java

cheister:
I ran into a problem where the output directory getting passed to Errorprone as a javac option ended in `.java` and it was mistakenly getting used for a source file.

e.g.

```
javac -d ouptut/com/hello.java com/exmplar/HelloWorld.java
```

I looked at using the Java Main or Arguments class to do the parsing via `processArgs()` instead but they changed it between Java 8 and Java 9 and say the API is use at your own risk.

### DIFF
--- a/check_api/pom.xml
+++ b/check_api/pom.xml
@@ -141,4 +141,18 @@
     </dependency>
   </dependencies>
 
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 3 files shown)


cushon:
`BaseErrorProneCompiler` is deprecated (https://github.com/google/error-prone/commit/e17752771ea771cc044768a95575682135b71657), I recommend using `BaseErrorProneJavaCompiler` instead, which among other things does flag handling in a more robust way.

PR #1011, opened 2018-04-25
BugChecker for TreeMap using either comparable keys or explicit comparator

madrob:
work in progress patch, functional but inelegant in places, specifically TreeMapNonComparableKeyWithoutComparator.java line 53 where I couldn't understand the API well enough to discover a better way.


madrob:
Bump. Can anybody take a look at this?

cpovirk:
Is this covered by the new https://github.com/google/error-prone/blob/master/core/src/main/java/com/google/errorprone/bugpatterns/Incomparable.java ?

madrob:
Yes

Issue #849, opened 2017-12-05
False positive in `UnnecessaryDefaultInEnumSwitch` triggered by (presumed) compiler bug


Stephan202:
I cannot provide a small reproduction case yet, but let me describe the situation.

Some code triggers the following warning:
```
[WARNING] /home/sschroevers/workspace/picnic/picnic-platform/wms/wms-logic/src/main/java/com/picnic/wms/logic/receiving/issue/IssueServiceImpl.java:[430,13] [UnnecessaryDefaultInEnumSwitch] Switch handles all enum values; an explicit default case is unnecessary and defeats error checking for non-exhaustive switches.
    (see http://errorprone.info/bugpattern/UnnecessaryDefaultInEnumSwitch)
  Did you mean to remove this line?
```

In actuality this code is covered by [case 3](https://github.com/google/error-prone/blob/master/core/src/main/java/com/google/errorprone/bugpatterns/UnnecessaryDefaultInEnumSwitch.java#L130): the `default` clause cannot be removed without breaking the code.

Above this warnings the following stacktrace is logged:
```
java.lang.RuntimeException: java.lang.NullPointerException
	at com.sun.tools.javac.api.JavacTaskImpl.handleExceptions(JavacTaskImpl.java:158)
	at com.sun.tools.javac.api.JavacTaskImpl.analyze(JavacTaskImpl.java:363)
	at com.google.errorprone.fixes.SuggestedFixes.compilesWithFix(SuggestedFixes.java:660)
	at com.google.errorprone.bugpatterns.UnnecessaryDefaultInEnumSwitch.matchSwitch(UnnecessaryDefaultInEnumSwitch.java:143)
	at com.google.errorprone.scanner.ErrorProneScanner.visitSwitch(ErrorProneScanner.java:1030)
	at com.google.errorprone.scanner.ErrorProneScanner.visitSwitch(ErrorProneScanner.java:146)
	at com.sun.tools.javac.tree.JCTree$JCSwitch.accept(JCTree.java:1223)
	at com.sun.source.util.TreePathScanner.scan(TreePathScanner.java:82)
	at com.google.errorprone.scanner.Scanner.scan(Scanner.java:82)
	at com.google.errorprone.scanner.Scanner.scan(Scanner.java:42)
	at com.sun.source.util.TreeScanner.scanAndReduce(TreeScanner.java:90)
	at com.sun.source.util.TreeScanner.scan(TreeScanner.java:105)
[...]

PR #1020, opened 2018-05-03
Moe Sync

ronshapiro:

Commits:
=====
<p> Create checker to enforce that names aren't named like type parameters.

RELNOTES: New, disabled-by-default checker [ClassNamedLikeTypeParameter], to
detect classes that looks like type parameters (e.g.: T, X, R1, R2)

bab06007ef1cb27a83d15aba9dae76091365dec8

-------

<p> Focus on the 'Builder' import case for now

RELNOTES: N/A

7e814fc7df0f35feb407f39da5b4b0973cbc0aef

-------

<p> Remove a call to printStackTrace

RELNOTES: N/A

b298498c6ccbfe95be41d902d97524352d79029e

-------

<p> Change CompilationTestHelperTest's assertions to use assertThrows to avoid issues with catching AssertionError from fail statements.

RELNOTES: n/a

549b98f4c875430534d2f1ff4cb9ac35104136b4

### DIFF
--- a/check_api/src/main/java/com/google/errorprone/fixes/SuggestedFixes.java
+++ b/check_api/src/main/java/com/google/errorprone/fixes/SuggestedFixes.java
@@ -734,7 +734,7 @@ public class SuggestedFixes {
     try {
       newTask.analyze();
     } catch (Throwable e) {
-      e.printStackTrace();
+      // ignored
     }
     return countErrors(diagnosticListener) == 0;
   }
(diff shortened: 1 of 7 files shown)


check_api/src/main/java/com/google/errorprone/fixes/SuggestedFixes.java
```diff
@@ -734,7 +734,7 @@ public CharSequence getCharContent(boolean ignoreEncodingErrors) throws IOExcept
     try {
       newTask.analyze();
     } catch (Throwable e) {
-      e.printStackTrace();
+      // ignored
```
Stephan202: @cushon, have you considered doing `return false` instead, as suggested in #849? Cause as-is this change only makes the issue in #849 harder to debug.

PR #1026, opened 2018-05-10
Moe Sync

ronshapiro:

Commits:
=====
<p> Check that start <= end in IndexedPosition

RELNOTES: None

7918e0916c88615c6e7b6f3267a67e4e965745da

-------

<p> BadComparable test case with File.

This was a motivating case for BadComparable and so it makes a nice test.

RELNOTES: extra BadComparable test.

a99ab45c00a519e53b05ed020be82f3430387bfa

-------

<p> Generate nicer replacements when having to convert literals.

This avoids waffle like `(double) 1f`.

RELNOTES: N/A

ee6215107e41ae91ee670f9be9359b1d0c14deea

-------

<p> AppliedFix: reject replacement ranges which exceed the source length

RELNOTES: None

7840996af21e32d0cf32d379f84cefb3eb0953fe

-------

<p> BadNestedImport: match on import, report on first occurrence only.

RELNOTES: None

836db1748a9006e5051d33b740c9505c0d3ac4e8

-------

<p> Add a check to catch chained proto builders setting the same field twice.

The most popular occurrence is single fields, but this will also match setting repeated or map fields if the index/key is a compile-time constant.

A suggestion to remove the overwritten set() is only generated if the two contain the same code--this isn't foolproof (they could be set to the result of an impure function), but I haven't seen any obvious false positives.

RELNOTES: [ProtoRedundantSet] Catch occurrences of proto fields being set twice in the same chained expression

eee92ae20eac111c80714b519f21a2dfe1651204

-------

<p> Update NullnessPropagationTransfer to recognize that if equals() returns true, argument in not null

equals() implementation should return false on null arg :
https://docs.oracle.com/javase/9/docs/api/java/lang/Object.html#equals-java.lang.Object-

RELNOTES: N/A

b072f22fdcb1a0776a14ed9e4ee94d092d0befcb

-------

<p> Flags equals() implementation which throw NPE when passed null argument

RELNOTES: Adds a new check which flags equals methods which throw NPE when passed null argument.

0d54d5a1328d7300587714105f8575f4a8a8aedd

-------

<p> Remove extraneous blank line in markdown.

9b3a398436afd4060754702be401cd3c334fcab6

-------

<p> Make the ExpectedException refactoring tool work when we can't determine the expected exception class.

ATM, when the ExpectedException refactoring tool encounters a «thrown.expect(...)» it assumes that it's of the form «thrown.expect(SomeClass.class)» and crashes otherwise.
This is definitely a common pattern, but it's not the case in general.
When refactoring some of our tests that don't follow this pattern, the tool failed on some targets.

RELNOTES: Make the ExpectedException refactoring tool work when we can't determine the expected exception class.

9276d795cfa020bc26a508c041a80b36c6a9a28a

### DIFF
--- a/annotation/pom.xml
+++ b/annotation/pom.xml
@@ -21,7 +21,7 @@
   <parent>
     <groupId>com.google.errorprone</groupId>
     <artifactId>error_prone_parent</artifactId>
-    <version>2.3.2-SNAPSHOT</version>
+    <version>2.2.1-SNAPSHOT</version>
   </parent>
 
   <name>@BugPattern annotation</name>
--- a/annotations/pom.xml
+++ b/annotations/pom.xml
@@ -21,7 +21,7 @@
   <parent>
     <groupId>com.google.errorprone</groupId>
     <artifactId>error_prone_parent</artifactId>
-    <version>2.3.2-SNAPSHOT</version>
+    <version>2.2.1-SNAPSHOT</version>
   </parent>
 
   <name>error-prone annotations</name>
--- a/ant/pom.xml
+++ b/ant/pom.xml
@@ -19,7 +19,7 @@
   <parent>
     <artifactId>error_prone_parent</artifactId>
     <groupId>com.google.errorprone</groupId>
-    <version>2.3.2-SNAPSHOT</version>
+    <version>2.2.1-SNAPSHOT</version>
   </parent>
   <modelVersion>4.0.0</modelVersion>
 
--- a/check_api/pom.xml
+++ b/check_api/pom.xml
@@ -31,7 +31,7 @@
   <parent>
     <groupId>com.google.errorprone</groupId>
     <artifactId>error_prone_parent</artifactId>
-    <version>2.3.2-SNAPSHOT</version>
+    <version>2.2.1-SNAPSHOT</version>
   </parent>
 
   <name>error-prone check api</name>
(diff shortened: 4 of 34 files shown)


annotation/pom.xml
```diff
@@ -21,7 +21,7 @@
   <parent>
     <groupId>com.google.errorprone</groupId>
     <artifactId>error_prone_parent</artifactId>
-    <version>2.3.2-SNAPSHOT</version>
+    <version>2.2.1-SNAPSHOT</version>
```
jbduncan: Hmm, I'm a bit puzzled by this drop in version. @awturner @ronshapiro Can you explain the reasoning behind this change for me?

jbduncan:
@ronshapiro Am I right to think that the ExpectedException refactoring tool mentioned in https://github.com/google/error-prone/pull/1026/commits/9276d795cfa020bc26a508c041a80b36c6a9a28a is Google-internal?

annotation/pom.xml
ronshapiro: Hrm, it looks like when @eaftan ran maven-release-plugin, the changes weren't synced internally, so our source of truth is inconsistent. I'll fix this and it will be in the next change. Thanks for the close inspection!

ronshapiro:
@jbduncan I think it refers to this: https://github.com/google/error-prone/blob/931be32e59a7c578491074eb1a8266d23ae620db/core/src/main/java/com/google/errorprone/bugpatterns/ExpectedExceptionChecker.java

annotation/pom.xml
jbduncan: 👍😊

jbduncan:
@ronshapiro Cool, thanks for the clarification!

PR #1045, opened 2018-06-04
Moe Sync

ronshapiro:

Commits:
=====
<p> Add a new option to UngroupedOverloads: FindingsOnFirstOverload

If true, then UngroupedOverloads will only report a finding on the *first*
overload for each method name, as opposed to reporting a finding on each:

void a(); // Always reported
void b(); // Always reported
void a(int a); // Not reported if flag is on
void b(int b); // Not reported if flag is on

RELNOTES: You can use '-XepOpt:UngroupedOverloads:FindingsOnFirstOverload' to
limit UngroupedOverloads' warnings to the first overload of each method name.

GOOGLE: And enable the option on the finding generator flume

1d18cb8eaab091af561951427301c55f7baa5894

-------

<p> In ApiDiffChecker, also recognize APIs that have a given annotation.

GOOGLE:
  And modify Java7ApiCheckerTest to recognize a new @RequiresJava8 annotation.
  Additionally, introduce RequiresJava8MustBeUsedWithMacro to check that @RequiresJava8 is used only with our macro (which doesn't exist yet). And enable it, which is safe because no one uses @RequiresJava8 yet.

RELNOTES: none

6a174ef20f8f7ee1706bf5ff6fba6311dcd3055f

-------

<p> Qualified super accesses are not static

RELNOTES: N/A

bc53d65d674517baba24fa0f079e27b8f2361fbf

-------

<p> Add "copyOf" as a bad import.

If "of" is bad, this is too.

RELNOTES: Add "copyOf" to BadImport

7aeb77d741cc9e56b6c7b78bf5d846ed607183e1

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/BadImport.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/BadImport.java
@@ -63,7 +63,7 @@ public class BadImport extends BugChecker implements ImportTreeMatcher {
       ImmutableSet.of(
           "Builder");
   private static final ImmutableSet<String> BAD_STATIC_IDENTIFIERS =
-      ImmutableSet.of("of", "from", "INSTANCE", "builder", "newBuilder");
+      ImmutableSet.of("copyOf", "of", "from", "INSTANCE", "builder", "newBuilder");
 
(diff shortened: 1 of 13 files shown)


eaftan:
@cpovirk your change 6a174ef appears to have broken our Windows build.  I commented on the internal CL, but could you PTAL?

PR #1047, opened 2018-06-05
Attempt to fix "cannot access unnamed package."

cpovirk:
DO NOT SUUBMIT; stacked on top of the other internal commits.


cpovirk:
No luck. Will re remove tests.

PR #1050, opened 2018-06-06
Another attempt to fix "cannot access unnamed package."

cpovirk:
Plus other commits.

DO NOT SUBMIT


cpovirk:
And that didn't help either :\

PR #1052, opened 2018-06-07
To see if ApiDiffCheckerTest fails on Windows when run alone

cpovirk:
DO NOT SUBMIT


cpovirk:
Got it working. Sent a CL to fix internally.

PR #1066, opened 2018-07-03
[Proposal] new checker for detecting unused variables with Dagger annotations

mkruber:
We have built a checker for detecting unused variables that have dagger annotations. More specifically, when a field with @Inject annotation is declared and unused, unnecessary code is generated. Similarly, when a method with @Provides annotation contains a parameter that is unused in its body, that also results in unnecessary code. The tests provide the necessary example scenarios where this checker can be useful. 

This is a proposal PR and we are happy to update the implementation based on the feedback. Let us know. 

Thanks!

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/inject/dagger/UsageCheckers.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/inject/dagger/UsageCheckers.java
@@ -24,22 +24,21 @@ import java.util.HashMap;
 import static com.google.errorprone.BugPattern.Category.DAGGER;
 import static com.google.errorprone.BugPattern.SeverityLevel.WARNING;
 
+/** @authors Murali Krishna Ramanathan, Manu Sridharan */
 /** Class that analyzes the use of variables to determine unused variables. */
 @AutoService(BugChecker.class)
-
-/** @authors Murali Krishna Ramanathan, Manu Sridharan */
 @BugPattern(
   name = "UsageCheckers",
-  altNames = {"UnusedParam", "UnusedInject"},
+  altNames = {"UnusedProviderParam", "UnusedMemberInjection"},
   summary = "Unused field with @Inject annotation.",
   category = DAGGER,
   severity = WARNING
 )
 public class UsageCheckers extends BugChecker implements BugChecker.ClassTreeMatcher {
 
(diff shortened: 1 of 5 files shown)


docs/bugpattern/UsageCheckers.md
```diff
@@ -0,0 +1,5 @@
+###UnusedInject
```
hzsweers: I think this should be `UnusedMemberInjection` since this doesn't cover constructor injection (which is probably the most common form of injection outside of android projects).

hzsweers:
@ronshapiro any thoughts on this?

core/src/main/java/com/google/errorprone/bugpatterns/inject/dagger/UsageCheckers.java
```diff
@@ -0,0 +1,126 @@
...
+        if (s.equals(checkerName)) return true;
+      }
+    }
+    return false;
+  }
+
+  @Override
+  public Description matchClass(ClassTree tree, VisitorState state) {
+    UsageCheckers.CallScanner callScanner = new UsageCheckers.CallScanner(state);
+    callScanner.scan(state.getPath(), null);
```
ronshapiro: I think we should invert this. We should be cautious that this will run on every Java class, so creating maps of every variable usage could be very wasteful.

Instead, I think we should have individual matchers for (a) fields and methods with `@Inject` and (b) methods with `@Provides` (and we should also add `@dagger.producers.Produces` too) and only when we find those should we then check to see if they're used. Then we probably don't have to do any state keeping of what is used or not - we can scan looking for a single symbol and once we find it, stop scanning. If it's never found, then we declare it unused.

msridhar:
@ronshapiro I was wondering how @mkruber's effort here relates to the EP [Unused checker](https://github.com/google/error-prone/blob/master/core/src/main/java/com/google/errorprone/bugpatterns/Unused.java) (which we didn't know about when this PR was opened).  Just wanted to check if the two efforts are related and whether they should be merged in some way.

ronshapiro:
I think I mentioned that checker to @hzsweers and was wondering what the additional value of this one was. I think that should subsume this one, no?

hzsweers:
Ah ok, seems like we can shutter this then

PR #1070, opened 2018-07-14
Create parent directory before writing patch file

AlexandreCarlton:
We may choose to place the generated `error-prone.patch` file in a
subdirectory of the base which might not already exist. For example, we
may wish to place it in target/patches instead of the base directory.


AlexandreCarlton:
I signed it!

Issue #1078, opened 2018-08-04
Refaster docs on the web site are out-of-date


dmitry-timofeev:
Hi,
"Running the Refaster refactoring" section of the Refaster documentation page is out of date, 
the instructions no longer work with 2.3.1: https://errorprone.info/docs/refaster

Would it be possible to update them, or add a link to https://github.com/google/error-prone/tree/master/examples/refaster if the latter page is supposed to contain an up-to-date example?

Thanks!

cushon:
Thanks, I pushed an update to https://errorprone.info/docs/refaster in 0ede1de0a59a6c782a242ca85981b8b71c92f944. Please re-open or file additional bugs if you run into any issues with the updated instructions.

dmitry-timofeev:
Brilliant, thank you!

PR #1079, opened 2018-08-05
Update the Refaster rule build instructions

dmitry-timofeev:
## Overview

Update the Refaster rule build instructions
to match the ones on the web site. Remove
no longer needed `J--add-exports` compiler
options.

See also:
  - 0ede1de
  - #1078
  - https://errorprone.info/docs/refaster


dmitry-timofeev:
I signed it!

PR #1083, opened 2018-08-07
error-prone doesn't work w/ latest javac version

iignatev:
running error-prone w/ javac from JDK11+ ea build leads to NoSuchMethodErrors b/c error-prone uses a few methods from which have been removed from com/sun/tools/javac/util/AbstractLog by [JDK-8196403](https://bugs.openjdk.java.net/browse/JDK-8196403).

the fix uses JCDiagnostic.Factory::create to create diagnostic objects and Log::report method. to get JCDiagnostic.Factory, I had to propagate an instance of Context in several places.


cushon:
Thanks! There are a couple of pending changes to finish JDK 10 support, and I'll merge this as soon as that lands.

PR #1084, opened 2018-08-07
Add a newline between assertThrows and the following assertions:

dmitry-timofeev:
## Overview
Add a newline after assertThrows with an expression lambda and the following assertions,
so that each statement appears on its own line.

### Note
`com.google.errorprone.BugCheckerRefactoringTestHelper#addOutputLines` does not seem to care about any whitespaces, therefore, I couldn't include a failing test.

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/AbstractExpectedExceptionChecker.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/AbstractExpectedExceptionChecker.java
@@ -274,7 +274,7 @@ public abstract class AbstractExpectedExceptionChecker extends BugChecker
       fix.prefixWith(throwingStatements.get(0), fixPrefix.toString());
       if (useExpressionLambda) {
         fix.postfixWith(((ExpressionStatementTree) throwingStatements.get(0)).getExpression(), ")");
-        fix.postfixWith(getLast(throwingStatements), Joiner.on('\n').join(newAsserts));
+        fix.postfixWith(getLast(throwingStatements), '\n' + Joiner.on('\n').join(newAsserts));
       } else {
         fix.postfixWith(getLast(throwingStatements), "});\n" + Joiner.on('\n').join(newAsserts));
       }


cushon:
We usually run a formatter ([e.g.](https://github.com/google/google-java-format)) on any modified lines after running refactorings, instead of trying to make sure the suggested fixes are well formatted. Is that an option for you?

> com.google.errorprone.BugCheckerRefactoringTestHelper#addOutputLines does not seem to care about any whitespaces, therefore, I couldn't include a failing test.

By default it does an AST diff and ignores whitespace, but there's an example of a whitespace-sensitive test here:

https://github.com/google/error-prone/blob/2623f228cece3a24fd7ccb77be478c4e81f2883d/core/src/test/java/com/google/errorprone/bugpatterns/MultiVariableDeclarationTest.java#L63

dmitry-timofeev:
> We usually run a formatter (e.g.) on any modified lines after running refactorings, instead of trying to make sure the suggested fixes are well formatted. Is that an option for you?

I see, that makes sense! I certainly do not propose to properly indent the code, but having two statements on a single line results in quite a long line in a patch. Auto-formatting might not work if a user has a build configuration that checks that that the formatting is valid, but does not enforce a _canonical_ formatting. In this case running a formatter that cannot operate on changed lines only might result in unrelated changes.

> By default it does an AST diff and ignores whitespace, but there's an example of a whitespace-sensitive test here:

Thank you for the reference! Do you think it is worth including a whitespace-sensitive test in this case or would it be too fragile? I've attached a patch, if it's OK, I would push it to this branch.
<details>
<summary>positiveMatchingExceptionMessage test</summary>

```
Index: core/src/test/java/com/google/errorprone/bugpatterns/ExpectedExceptionRefactoringTest.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- core/src/test/java/com/google/errorprone/bugpatterns/ExpectedExceptionRefactoringTest.java	(date 1533650826000)
+++ core/src/test/java/com/google/errorprone/bugpatterns/ExpectedExceptionRefactoringTest.java	(date 1533655406000)
@@ -16,6 +16,8 @@
 
 package com.google.errorprone.bugpatterns;
 
+import static com.google.errorprone.BugCheckerRefactoringTestHelper.TestMode.TEXT_MATCH;
+
 import com.google.errorprone.BugCheckerRefactoringTestHelper;
 import java.io.IOException;
 import org.junit.Test;
@@ -74,6 +76,51 @@
         .doTest();
   }
 
+  @Test
+  public void positiveMatchingExceptionMessage() throws IOException {
+    testHelper
+        .addInputLines(
+            "in/ExceptionTest.java",
+            "import static org.hamcrest.MatcherAssert.assertThat;",
+            "import static org.hamcrest.Matchers.nullValue;",
+            "import java.util.Arrays;",
+            "",
+            "import org.hamcrest.CoreMatchers;",
+            "import org.junit.Rule;",
+            "import org.junit.Test;",
+            "import org.junit.rules.ExpectedException;",
+            "class ExceptionTest {",
+            "  @Rule ExpectedException thrown = ExpectedException.none();",
+            "  @Test",
+            "  public void test() throws Exception {",
+            "    thrown.expectMessage(nullValue(String.class));",
+            "    thrown.expect(NullPointerException.class);",
+            "    Arrays.asList(null);",
+            "  }",
+            "}")
+        .addOutputLines(
+            "out/ExceptionTest.java",
+            "import static org.hamcrest.MatcherAssert.assertThat;",
+            "import static org.hamcrest.Matchers.nullValue;",
+            "import static org.junit.Assert.assertThrows;",
+            "",
+            "import java.util.Arrays;",
+            "import org.hamcrest.CoreMatchers;",
+            "import org.junit.Rule;",
+            "import org.junit.Test;",
+            "import org.junit.rules.ExpectedException;",
+            "class ExceptionTest {",
+            "",
+            "  @Test",
+            "  public void test() throws Exception {",
+            "",
+            "    NullPointerException thrown = assertThrows(NullPointerException.class, () -> Arrays.asList(null));",
+            "assertThat(thrown.getMessage(), nullValue(String.class));",
+            "  }",
+            "}")
+        .doTest(TEXT_MATCH);
+  }
+
   @Test
   public void noThrowingStatements() throws IOException {
     testHelper
```

</details>

cushon:
> Do you think it is worth including a whitespace-sensitive test in this case or would it be too fragile?

Thanks, I think skipping the test is OK in this case. I'll merge this change momentarily.

PR #1089, opened 2018-08-15
Moe Sync

ronshapiro:

Commits:
=====
<p> Make ProtoFieldNullComparison match checkNotNull (+ similar) and Lite protos behind flags.

We can either add the extra functionality of matching via final variables into ProtoFieldPreconditionsCheckNotNull or combine the checks into one. The latter makes more sense to me; they were duplicating most of their code anyway.

RELNOTES: Make ProtoFieldNullComparison match checkNotNull & lite protos

899169b7ddf440d9b8ae3d2e72213d84c1e994a5

-------

<p> Created a check to avoid using protos as key of map or member of Set
RELNOTES: Added ProtosAsKeyOfSetOrMap check

871843ad14f50cc3cb10b02df152f53123040842

-------

<p> Remove duplicated word.

875c6df17c6d036af8a33d6250c8fea62317195e

-------

<p> Make EqualsBrokenForNull track reassignments (with casting) of the incoming parameter.

RELNOTES: Make EqualsBrokenForNull track reassignments (with casting) of the incoming parameter.

9d03645748ac2dd8713f14e349fba0efa2511f52

-------

<p> Replace more equals matchers with helpers from Matchers.

RELNOTES: N/A

f6f38fbab0033824c589c50efc7b447d1de7e6dc

-------

<p> Disable ProtosAsKeyOfSetOrMap for now while we decide what to do with it.

RELNOTES: n/a

bbd2d2596b23385f5819a7a5b53fd1f1b753915c

-------

<p> TypeParameterNaming will not suggest replacing IDataT -> IDataTT, instead
replacing it with I.

RELNOTES: n/a

5cf714ca00264823a23371cc963b16d11867c41e

-------

<p> Add a bugpattern to catch instanceof checks which are always true or equivalent to a null check.

These are not flagged by the compiler, and JLS specifically calls out instanceof as not being considered a constant expression: so this pattern can lead to completely unreachable code that javac won't flag.

RELNOTES: Add a bugpattern to catch redundant instanceof checks.

d2065f0d2fade6c8b58b4b30d49b36e36cd235a3


cushon:
#1090

Issue #1021, opened 2018-05-04
false positive in OrphanedFormatString check


msridhar:
### What version of Error Prone are you using?

Latest master

### Does this issue reproduce with the latest release?

Yes

### What did you do?

```java
import java.util.Locale;
class OrphanedTest {
  public static void main(String[] args) {
    StringBuilder b = new StringBuilder().append("Hello ").append("%s");
    String formatted = String.format(Locale.getDefault(), b.toString(), "world");
    System.out.println(formatted);
  }
}
```

### What did you expect to see?

No warning

### What did you see instead?

```
OrphanedTest.java:4: warning: [OrphanedFormatString] String literal contains format specifiers, but is not passed to a format method
    StringBuilder b = new StringBuilder().append("Hello ").append("%s");
                                                                  ^
    (see http://errorprone.info/bugpattern/OrphanedFormatString)
1 warning
```

awturner:
Thanks for the report.

This will be fixed in the next release. The check on `StringBuilder.append` has been tightened to only consider invocations which pass parameters (i.e. `append(String, int, int)` and `append(char[], int, int)`).

PR #1117, opened 2018-09-12
Moe Sync

### LINKED ISSUE #35: Check that equals() and hashCode() read the same fields
_[Original issue](https://code.google.com/p/error-prone/issues/detail?id=35) created by **cpovirk@google.com** on 2012-08-06 at 08:48 PM_

---

If I add a new field to a class, I can remember to update equals() but forget to update hashCode().  How would we detect this?  In many cases, both methods should read all fields of a class, but that's probably too strong a check.  (For example, a List.equals() implementation might read no fields directly, preferring to operation on the public size() and get() methods.)  A weaker but probably still useful check is that the two methods read the same set of fields.  False positives are still possible.  For example, the check would flag a field containing a cached hash code, which would likely be read in hashCode() but not equals().  This particular example is avoidable by requiring the hashCode() looks at a subset of the fields that equals() looks at.  I conjecture that it's more common for programmers to update equals() but forget to update hashCode() than vice versa, so this further weakened check would likely still catch most problems.

ronshapiro:

Commits:
=====
<p> Stop passing unused VisitorState to SuppressionHelper.

RELNOTES: None

84f3bbc923504324aaf8e4bd1240cdd373fbc431

-------

<p> Ignore super(...) invocations in MustBeClosedChecker

RELNOTES: Ignore super(...) invocations in MustBeClosedChecker

75b0055324da6c6565115d4548306ad03d6c9101

-------

<p> OrphanedFormatString: Only consider StringBuilder.append a likely error if the append(CharSequence, int, int) or append(char[], int, int) overloads are invoked.

RELNOTES: OrphanedFormatString: only consider StringBuilder.append a likely error for some overloads (https://github.com/google/error-prone/issues/1021)

233651e6ea77903488f0280455d9e517a02c2bec

-------

<p> Fix flagging serialization methods in Unused.

Extract the test so it's a bit clearer.

RELNOTES: N/A

ccdc625a1853a667eeca7fc0fa4dfd3d8adee045

-------

<p> Make ant example's path handling work on Windows

https://github.com/google/error-prone/commit/3f2ca43e53266e84e17e4fb3f34888be5cb29293#commitcomment-30383244

RELNOTES: N/A

2e89884022017fe2139c801e942f7f28da2a627e

-------

<p> Remove ProtoFieldPreconditionsCheckNotNull, as it is subsumed into ProtoFieldNullComparison.

RELNOTES: N/A

45cc93d6fcd3b20bdda4a13e64dba3909b9d14d6

-------

<p> Fix treatment of non-trivial receivers of method selects in Unused.

RELNOTES: Fix a false positive in Unused.

c6c010f489b2e8d327bd7f1c9c021b11cde63046

-------

<p> Provide a marginally better description in BadInstanceof.

It seems slightly friendlier to also include the actual type of the expression to make the issue a bit clearer.

RELNOTES: N/A

65fe799bce61e82d97c64c0c334a6993c5332559

-------

<p> Provide a different description when removing the `default` case in UnnecessaryDefaultInEnumSwitch.

RELNOTES: N/A

5b094ec90504e1352d1fb8c19e215141d2201307

-------

<p> Remove altName of BadNestedImport from BadImport checker.

There is no need to continue to support both names.

RELNOTES: Remove altName of BadNestedImport from BadImport checker.

135c920d0dc85176c3b9d54e95df8985d04b6337

-------

<p> Add support for suppression of FieldCanBeFinal

Fixes #1038

RELNOTES: N/A

e5eb00a9adc53eafd680b3c0e05a8f9f0385f9ab

-------

<p> Add a bug pattern to discourage writing #equals methods by just calling hashCode.

I'm kinda surprised this is a thing. We only match where the hashCode is the last thing compared within a `return, as I think this is a fairly sane pattern, especially with memoized hashCodes:

  return hashCode() == o.hashCode() && <expensive comparison>;

RELNOTES: [EqualsUsingHashCode] add a bug pattern to discourage writing equals using hashCode.

4fc4181fa400657094ad61da577f44523240f7a8

-------

<p> Add a check for obviously inconsistent hashCode implementations.

This will have plenty of false negatives for classes which compare fields via complex getters, but should work fairly well for simple data classes.

Fixes #35

RELNOTES: [InconsistentHashCode] Add a check to flag inconsistent implementations of hashCode.

e8cf0b5b28ff67ca41681755b41d70159bada09f

-------

<p> Make BooleanParameter match constructors too.

And add a couple of fairly common terms to the list of parameter names to ignore.

RELNOTES: Make BooleanParameter match constructors.

36642fa767d361048439db1bb7a01e88a70ba0ba

-------

<p> In BooleanParameter, don't refactor non-boolean parameters, e.g. generic parameters or Object.

RELNOTES: BooleanParameter now skips non-boolean parameters, e.g. generics.

18c54a07b22c402f9d3522d6f7d4ffa96cc8779f

-------

<p> Revise docs to try to make them clearer.

4667101c24d9ceea39bd7d3d37b39bab4165955f

-------

<p> Fix ParcelableCreator check for raw creator

RELNOTES: Fix ParcelableCreator check for raw creator

f8d8e9349b4374f031b0010e24febca9ad53fd38

-------

<p> Create namedAnyOf(String... varargs) convenience MethodMatchers method.

RELNOTES: N/A

3926cdeff1419bb7a60cbd4b730a71435a2169e9

-------

<p> Show more Error Prone diagnostics at once

RELNOTES: Report errors for all files with Error Prone diagnostics, not just the first one.

435c6ba9c22021263871b397250c5f346a9a3e7d

-------

<p> Make TypeParameterShadowing rename Javadoc references to type parameters too.

RELNOTES: Make TypeParameterShadowing fix Javadocs too.

93b2a7e0f8dc875d706d9e5f534ea34ca04f8e41

-------

<p> Make ParameterComment match non-parameter-comments before arguments too.

Previously it wouldn't match the non-standard form `foo(/* bar */ baz)`.

RELNOTES: Make ParameterComment match before comments too.

09b08e9efc0bcb7a209cb78cbe679b356be8fb92

-------

<p> Make insertion coalescing in Replacements#add explicit

Instead of silently concatenating overlapping insertions together, require
callers of Replacements#add to explicitly requires a policy for concatenating
them (either existing-first, or new-first). The default behaviour now throws
an exception for overlapping insertions.

RELNOTES: N/A

f6d7c53eafe69287fb2a12c9e02c0a6a59ed556d

-------

<p> Call out CharSequence in UndefinedEquals docs
RELNOTES: N/A

958c0dfe909c75515a47eca9998ad1bbb28a3a70

-------

<p> Fix breakage caused by f6d7c53eafe69287fb2a12c9e02c0a6a59ed556d

f9c9e4123436c36522259d725292be2e0bd7544d

-------

<p> Add a qualifyType overload which takes a String of the type name.

A common pattern is qualifyType(..., state.getTypeFromString(typeName)) which stands a fair chance of NPEing if typeName isn't available.

RELNOTES: Add String-taking qualifyType overload.

803cf50a454c9ea9a7573619e951bacf5847be4f

-------

<p> Exempt constructors for classes containing the term "boolean" (aimed at AtomicBoolean).

These are an obvious exception to my claim that "constructors accepting a single boolean are rarely self documenting".

(Also include 'defaultValue'.)

RELNOTES: N/A

ebe4d095d2a372ab1f7f17a4da6438a04489cd10


eaftan:
It looks like commit 435c6ba broke the JDK 11 build.  @cushon can you PTAL?

Issue #444, opened 2016-07-30
Generate parsable reportfile


tomasbjerre:
I'd like to be able to generate reportfiles with result of the analysis. In my case I need a parsable report (there are some examples to perhaps resemble [here](https://github.com/tomasbjerre/violations-lib/tree/master/src/test/resources)).

tomasbjerre:
I'm currently supporting it by parsing the output from the build log:
https://github.com/tomasbjerre/violations-lib/blob/master/src/main/java/se/bjurr/violations/lib/parsers/GoogleErrorProneParser.java

But still a format made for parsing would be better!

leinardi:
+1 for this request, this will help with CI since parsing the entire build output for ErrorProne errors isn't that easy and reliable.

Vyom-Yadav:
+1
It would be nice if this is implemented, current at Checkstyle, we are using a [custom groovy script](https://github.com/Vyom-Yadav/checkstyle/blob/enableErrorProne/.ci/error-prone-check.groovy) to parse the output and get the errors. Built in support will be beneficial.

tbroyer:
ErrorProne uses standard javac diagnostics, so IMO this is something that should be supported at the build tool level. This would then automatically include Xlint warnings and other compilation errors as well.

faisal6621:
> I'm currently supporting it by parsing the output from the build log: https://github.com/tomasbjerre/violations-lib/blob/master/src/main/java/se/bjurr/violations/lib/parsers/GoogleErrorProneParser.java
> 
> But still a format made for parsing would be better!

Hi @tomasbjerre 
I was looking at the https://github.com/tomasbjerre/violations-lib/ and https://github.com/tomasbjerre/violations-maven-plugin. However, in your example for the maven plugin you have not explained well what the `<pattern>` should be for various parsers?

Can you explain that or mention what pattern should be used for _GoogleErrorProne_?

Regards,
Faisal

mihnita:
Hi all,

I have more or less the same need like the others.
Basically: run "something" on a project, and generate a report with all the error-prone issues detected.
The project is maven and multi-module.


Where I had problems :

**Running errorprone fails on errors**

Which means I can't generate the report with all the issues detected.

Yes, I can specify `-XepAllErrorsAsWarnings`
But then all errors are "merged" with the warnings.
Which makes the report less useful.

So this is basically the same as https://github.com/google/error-prone/issues/4546

[...]

PR #1132, opened 2018-10-02
Parseable report file

stalar:
This is a first shot at implementing https://github.com/google/error-prone/issues/444. Comments and feedback welcome!

This PR adds a new option called `-XepReportFile:<path>`. If this option is present, an xml document will be created in the location specified. The contents of the document looks like:

    <?xml version="1.0" encoding="utf-8"?>
    <errorprone timestamp="2018-10-02T13:05:37.389+02:00[Europe/Stockholm]">
    <descriptions>
      <description
        type="WARNING"
        source="/Users/sla/dev/example/src/main/java/com/example/Example.java"
        line="6"
        column="8">
        <message>[PrivateConstructorForUtilityClass] Utility classes (only static members) are not designed to be instantiated and should be made noninstantiable with a default constructor.
    (see https://errorprone.info/bugpattern/PrivateConstructorForUtilityClass)
    Did you mean to remove this line?</message>
      </description>
      ...
    </errorprone>

The implementation is done in a new class called `FileReporter` which is created on startup and passed down to the `JavacErrorDescriptionListener`. The class is also a `TaskListener` so that the file can be closed when the compilation is done.


jbnerd:
Any possibilities of this getting merged any time soon?

tomasbjerre:
@stalar Are you aware of the merge conflict? Really hope you get this merged =) Nice job!

davidhart82:
+1 for this, we're having to parse an awfully large build output to find the errors. This should bring it on line with how every other static analysis tool works.

davidhart82:
Though have you thought about reusing one of the other already supported formats available? It would mean that most tools, and in particular, CI systems would already be able to handle this.

check_api/src/main/java/com/google/errorprone/JavacErrorDescriptionListener.java
```diff
@@ -57,11 +58,14 @@
 
   private JavacErrorDescriptionListener(
       Log log,
+      FileReporter fileReporter,
```
davidhart82: Feels like we'd be better having a Reporter interface?

check_api/src/main/java/com/google/errorprone/FileReporter.java
```diff
@@ -0,0 +1,82 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.time.ZonedDateTime;
+import org.apache.commons.text.StringEscapeUtils;
+
+/**
+ * Generates a parseable report file.
+ */
+public class FileReporter implements TaskListener {
```
davidhart82: XMLFileReporter?

check_api/src/main/java/com/google/errorprone/FileReporter.java
```diff
@@ -0,0 +1,82 @@
...
+      }
+    }
+  }
+
+  public void report(final JCDiagnostic.DiagnosticType type, final DiagnosticSource source,
+                     final JCDiagnostic.DiagnosticPosition pos, final String message) {
+    if (pw != null) {
+      pw.println("    <description");
+      pw.println("      type=\"" + type + "\"");
+      pw.println("      source=\"" + source.getFile().getName() + "\"");
```
davidhart82: Probably want to escape this?

check_api/src/main/java/com/google/errorprone/FileReporter.java
```diff
@@ -0,0 +1,82 @@
...
+        pw.close();
+      }
+    }
+  }
+
+  public void report(final JCDiagnostic.DiagnosticType type, final DiagnosticSource source,
+                     final JCDiagnostic.DiagnosticPosition pos, final String message) {
+    if (pw != null) {
+      pw.println("    <description");
+      pw.println("      type=\"" + type + "\"");
```
davidhart82: Probably want to escape this?

check_api/src/main/java/com/google/errorprone/FileReporter.java
tbroyer: How about using an `XMLStreamWriter`? https://hsivonen.fi/producing-xml/

check_api/src/main/java/com/google/errorprone/FileReporter.java
```diff
@@ -0,0 +1,82 @@
...
+      throw new RuntimeException(e);
+    }
+  }
+
+  @Override
+  public void started(final TaskEvent e) {
+    if (e.getKind() == TaskEvent.Kind.COMPILATION) {
+      if (pw != null) {
+        pw.println("<?xml version=\"1.0\" encoding=\"utf-8\"?>");
+        pw.println("<errorprone timestamp=\"" + ZonedDateTime.now() + "\">");
```
davidhart82: This should be formatted, ideally with an ISO standard.

tbroyer: It is already: `ZonedDateTime#toString()` produces an ISO 8601 timestamp.

davidhart82: Relying on a `toString` implementation feels a bit fragile?  I might have used a `DateFormat` with the format specifier `"yyyy-MM-dd'T'HH:mm:ss.SSS"` just to make it more obvious?

davidhart82:
How about having the report configured like so: `-XepReportFile:<formatType>:<path>`?

check_api/src/main/java/com/google/errorprone/JavacErrorDescriptionListener.java
```diff
@@ -106,6 +110,7 @@ public void onDescribed(Description description) {
           break;
       }
       log.report(factory.create(type, log.currentSource(), pos, MESSAGE_BUNDLE_KEY, message));
+      fileReporter.report(type, log.currentSource(), pos, message);
```
davidhart82: I wonder whether we should have a composite reporter, to abstract the use of both of these `report` methods?  Taking a `JCDiagnostic` type as the param to `report` does not seem a terrible idea either.

davidhart82: Though I haven't looked at where `Log` manifests from!

stalar:
I no longer have an interest in this. Feel free to continue working on the feature - use my code as you wish!

Issue #1144, opened 2018-10-17
Raison d'être of EqualsGetClass is fishy


jcayzac:
### What version of Error Prone are you using?
2.3.2

### Does this issue reproduce with the latest release?
Yes

### What did you do?
I wrote an `Foo#equals(Object)` override with the following implementation:

```java
  @Override
  public boolean equals(final Object o) {
    if (this == o) return true;
    if (o == null || getClass() != o.getClass()) return false;

    final Foo that = (Foo) o;
    return field1.equals(that.field1)
        && field2.equals(that.field2)
        && field3.equals(that.field3);
  }
```

### What did you expect to see?
No error prone warning.

### What did you see instead?
An `EqualsGetClass` advocating to replace my perfectly-valid code with a bogus implementation breaking **symmetry**, [a strong requirement of `Object#equals(Object)`](https://download.java.net/java/early_access/jdk11/docs/api/java.base/java/lang/Object.html#equals(java.lang.Object)).

cushon:
Implementations using `instanceof` can be symmetric as long as subtypes don't add additional attributes and include them in their `equals` implementations, which they shouldn't.

The `getClass` approach violates the Liskov substitution principle if it's used on non-final classes.

For more information, see Effective Java Item 8 and [this interview with Josh Bloch](https://www.artima.com/intv/bloch17.html).

jcayzac:
Thanks. I remember reading J.Bloch comment on that part and I agree with his (more nuanced) take on the problems with `getClass()`. I think he also makes it clear there are pros and cons to both approach and that it's a matter of opinion. I think "fishy" was a poor choice of word, sorry about that. Since I can silence the warning with `-Xep:EqualsGetClass:OFF` I'll close the issue.

Thanks for the quick reply!

ejona86:
I'm surprised Error Prone is taking a position on this.

Basically, the problem with that the Liskov substitution principle is that it generally doesn't apply. A subclass isn't equal: you _couldn't_ replace one instance with the other, because then you would lose information. Someone created the subclass instead of the base class for a reason. Collections are the only common case it seems to hold in code that I write.

And this is evident in the examples and Josh's discussion. It says "innocuous". What is an innocuous change to a subclass, such that you _could_ replace one with another? Is there a non-container example, in practice, you could reference?

> as long as subtypes don't add additional attributes and include them in their equals implementations, which they shouldn't.

[...]

PR #1152, opened 2018-10-18
Moe sync

cushon:


jbduncan:
@kevinb9n, thank you very much for writing up an initial explanation for why `instanceof` is preferred over `getClass()` for custom `equals()` implementations, over in commit https://github.com/google/error-prone/pull/1152/commits/b9078a9d77fb0c1c45801a5ce973e5e6dda9939f (which I assume was written in response to issue https://github.com/google/error-prone/issues/1144)!

However, I admit that I'm rather confused by the explanation given, because from what I've understood so far, it goes into why inheritance and custom `equals()` impls should never be mixed, rather than why `instanceof` is preferable to `getClass()` when inheritance and `equals()` _are_ mixed.

Furthermore, it's a bit late in the evening for me to read through my copy of Effective Java to understand why Joshua Bloch thinks that `instanceof` is preferable, so I'm lacking the necessary context.

Thus I wonder if you have any plans to improve the documentation w.r.t. all this? :)

kevinb9n:
Thanks @jbduncan for asking - I think I'll move this discussion to the issue in question.

PR #1154, opened 2018-10-19
Whitelist CompletableFuture#exceptionally in FutureReturnValueIgnored

pkoenig10:


pkoenig10:
I signed it!

cushon:
The function passed to `exceptionally` may be invoked and produce a result which should not be ignored.

Can you provide more context about the situations where you want to ignore the return value?

pkoenig10:
There is no way to asynchronously handle exceptions from a `CompletableFuture` without also returning a new future. It is likely that after some amount of error handling there is nothing else that can be done. For example, eventually you might simply log that error handling has failed and don't really care about handling exceptions thrown when logging.
```
CompletableFuture.runAsync(...)
    .whenComplete(...) // handle result or exception
    .exceptionally(throwable -> {
        log.error("Something went wrong");
        return null;
    });
```

To use a `CompletableFuture` you almost always have suppress this warning, which defeats the purpose of having the warning in the first place.

PR #1163, opened 2018-11-02
Moe Sync

cushon:


annotation/pom.xml
```diff
@@ -21,7 +21,7 @@
   <parent>
     <groupId>com.google.errorprone</groupId>
     <artifactId>error_prone_parent</artifactId>
-    <version>2.3.3-SNAPSHOT</version>
+    <version>2.3.2-SNAPSHOT</version>
```
Stephan202: Accidental version downgrade?

cushon: Yep, thanks for the catch: #1164

PR #1170, opened 2018-11-08
Add missing declaration, to clarify explanation

mernst:


docs/bugpattern/CollectionIncompatibleType.md
```diff
@@ -9,6 +9,7 @@ element type are "incompatible." A typical example:
 
 ```java
 Set<Long> values = ...
+Integer i = ...
```
tbroyer: There's no `i` anywhere in the code snippets. Did you misread the `1` (one) for an `i`?

mernst: Yes, I did.  Oops!  Sorry for the false alarm.

I should increase my font size.
In addition, I think it would also be useful to change that to a "more obviously  integer" value such as 42.
Thanks.

cushon: Good idea, I'll switch it to use `42` instead.

mernst: Thanks!

Issue #1185, opened 2018-11-23
How to use ErrorProne with Lombok: @Value annotations seem disabled?


RichardBradley:
### Description of the problem / feature request:

I am trying to use ErrorProne and Lombok together.
Various open issues in this repos suggest that this is possible but perhaps difficult or unsupported?

I have found that using ErrorProne seems to entirely disable the "`@Value`" and "`@Data`" and similar Lombok annotations, causing compile errors in code that relies on the getters generated by these annoations.

How can I get these two libraries to work together?

### Feature requests: what underlying problem are you trying to solve with this feature?

I am trying to use ErrorProne on a codebase that relies on Lombok code-generation annotations.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

I have prepared two simple examples based on the "`examples/maven`" example in this repos:

The relevant class looks like this:

```
import lombok.Value;

@Value
public class SomeDataClass {

    public String name;

    @Override
    public String toString() {
        return "SomeDataClass[" + this.getName() + "]";
    }
}
```

1) Maven and Lombok together: https://github.com/RichardBradley/error-prone/tree/d3ca8c6716a8cde2b2489f5beafa36837acc54e9/examples/maven-with-lombok

This fails to build with the following message:

```
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR :
[INFO] -------------------------------------------------------------
[ERROR] C:\Work\error-prone\examples\maven-with-lombok\src\main\java\SomeDataClass.java:[11,38] error: cannot find symbol
  symbol: method getName()
[INFO] 1 error
```

2) The same code, but with ErrorProne disabled: https://github.com/RichardBradley/error-prone/tree/ccadedda3/examples/maven-with-lombok

This builds fine

### What version of Error Prone are you using?

Latest `master`, 976ab7c62771 (`2.3.3-SNAPSHOT`)

###  Have you found anything relevant by searching the web?

Related issues:
 * #775 : someone using Lombok `@Data` with ErrorProne, closed as dup
 * #751 "Lombok & error-prone dont play together ?": closed as dup (of which?)
 * #690: Lombok breaks NestedInstanceOfConditions and InstanceOfAndCastMatchWrongType - closed as fixed?
 * #932: seems to be a similar or same question, but was closed without comment

Many thanks,

Rich

tbroyer:
[...]

PR #1186, opened 2018-11-23
#1185 add a note about Lombok to the installation instructions

RichardBradley:
Addresses the question I raised in #1185: how to use Lombok with ErrorProne

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -60,6 +60,7 @@ Edit your `pom.xml` file to add settings to the maven-compiler-plugin:
               <artifactId>error_prone_core</artifactId>
               <version>2.3.3-SNAPSHOT</version>
             </path>
+            <!-- Add other annotation processors and Error Prone custom checks here if needed -->
           </annotationProcessorPaths>
         </configuration>
       </plugin>
@@ -142,7 +143,7 @@ and add the following javac task to your project's `build.xml` file:
     <path id="processorpath.ref">
       <pathelement location="${user.home}/.m2/repository/com/google/errorprone/error_prone_core/2.3.3-SNAPSHOT/error_prone_core-2.3.3-SNAPSHOT-with-dependencies.jar"/>
       <pathelement location="${user.home}/.m2/repository/com/google/code/findbugs/jFormatString/3.0.0/jFormatString-3.0.0.jar"/>
-      <!-- Add annotation processors and Error Prone custom checks here if needed -->
+      <!-- Add other annotation processors and Error Prone custom checks here if needed -->
     </path>
 
     <javac srcdir="src" destdir="build" fork="yes" includeantruntime="no">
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -261,3 +261,12 @@ visible to the classloader which currently locates the javac Main class. Then
 replace the call of `javac.main.Main.main()` with the Error Prone compiler:
 
 `return new ErrorProneCompiler.Builder().build().compile(args) == 0`
+
+## Using ErrorProne in concert with other Annotation Processors
+
+All of the above instructions use the `javac` option `-processorpath` which has
+side-effect of causing `javac` to no longer scan the compile classpath for
+annotation processors. If you are using other annotation processors in addition
+to ErrorProne, such as Lombok, then you will need to add their JARs to your
```
tbroyer: …or Dagger, AutoValue, Mapstruct, Immutables, etc.

docs/installation.md
```diff
@@ -261,3 +261,12 @@ visible to the classloader which currently locates the javac Main class. Then
...
+`-processorpath` argument. The mechanics of this will vary according to the build
+tool you are using.
```
tbroyer: …too bad this does not list them :wink: 

How about adding a comment to the snippets above saying something like "insert your other annotation processors here"?
(though that this applies mainly to Maven; it gained only recently the ability to use the processor path, and almost nobody updated their build, and most importantly their knowledge)

docs/installation.md
RichardBradley: Of course. 
Are you asking me to change the PR to list these other examples?
I don't think that adds a lot, but I am happy to do so if you like.
Are you happy with listing those 5 examples? It seems a bit overkill to me

docs/installation.md
RichardBradley: Good point
I did similar to that in the adjacent PR #1187
I will add the same change to Maven above.
I'll have a go at the others, but Bazel and Gradle look like closed black-boxes to me that don't admit such a change at present

RichardBradley: > Maven gained only recently the ability to use the processor path, and almost nobody updated their build, and most importantly their knowledge

Why use "-processorpath" instead of adding these things to the main classpath in "provided" scope? It seems to me that the latter is more robust and much easier when multiple annotation processors are in use.

tbroyer: > Why use "-processorpath" instead of adding these things to the main classpath in "provided" scope? It seems to me that the latter is more robust and much easier when multiple annotation processors (or custom ErrorProne rules) are in use.

ErrorProne is a plugin (`-Xplugin`), and those are **only** loaded from the processor path, not the compilation classpath.

Wrt other usages of annotation processors, without ErrorProne in loop, having too separate classpaths means you can have separate dependencies, or put differently, that annotation processors won't impose their own dependencies on you. For example, if you use Guava and target Java 7, you won't use a version past 20.0 (or so, from memory); or if you target Android you'll use a special version targeting Android; you don't want your annotation processors (such as Dagger or AutoValue) to impose a different or more recent version of Guava. Putting them in the processor path helps segregate those dependency trees.

RichardBradley: Thanks, that makes sense

cushon:
https://github.com/google/error-prone/commit/4c29d5add584e24cb09c05f7333f365569388d4e

PR #1195, opened 2018-12-05
Fix errors when suggesting replacements to Lombok generated code

RichardBradley:
I am trying to introduce `error-prone` to a large codebase which uses Lombok (see also #1185).

I have found that a few rules attempt to suggest fixes to code that Lombok has auto-generated, particularly "setters" and "withers" which use reference equality in a way that upsets some rules.

This auto-generated bytecode does not have valid positions in the sourcecode files, so fix suggestion throws an exception.

This change ignores that exception and skips the fixes in question.

### DIFF
--- a/check_api/src/main/java/com/google/errorprone/JavacErrorDescriptionListener.java
+++ b/check_api/src/main/java/com/google/errorprone/JavacErrorDescriptionListener.java
@@ -20,6 +20,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.errorprone.fixes.AppliedFix;
 import com.google.errorprone.fixes.Fix;
+import com.google.errorprone.fixes.Replacement;
 import com.google.errorprone.matchers.Description;
 import com.sun.source.tree.Tree.Kind;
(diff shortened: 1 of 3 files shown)


RichardBradley:
Are you waiting for anything from me on this?
Is there anything I can do to get this merged in?

Thanks,


Rich

cushon:
Hi, sorry for not responding to this earlier.

I would prefer to avoid handling these errors at this level and discarding the invalid fixes. While this does work around some of the issues with lombok creating AST nodes with invalid source positions, it would also cause Error Prone to silently skip fixes that were incorrect for other reasons, e.g. if there was a bug in a check.

There's some related discussion about the underlying problem lombok creates for Error Prone in https://github.com/google/error-prone/issues/1250#issuecomment-763097162, and a link to an upstream lombok bug with they're considering fixing up the source position information for the AST nodes lombok edits, which would help prevent this kind of crash. I'm hopeful that approach will result in a solution that allows lombok and Error Prone to run in the same compilation without crashing Error Prone.

PR #1199, opened 2018-12-12
Moe Sync

### LINKED ISSUE #1196: Erroneous compile error due to ZonedDateTime formatted as timezone

The following code generates a build error but the test passes

> [FormatString] illegal format conversion: 'java.time.LocalDateTime' cannot be formatted using '%Z' (see http://errorprone.info/bugpattern/FormatString)

    @Test
    public void testErrorProne() {
        assertEquals("GMT", String.format("%tZ", ZonedDateTime.of(LocalDate.of(2018, 12, 27), LocalTime.of(17, 0), ZoneId.of("Europe/London"))));
    }

2.3.2

ronshapiro:

Commits:
=====
<p> RELNOTES: Disallow @CompileTimeConstant when overriding methods without it.

0edbcef1b60cab0e5792904bed08d6e2fe722a93

-------

<p> Refactoring-only CL: Replace multiple constructors of VisitorState with factory methods, using them as appropriate within the Error Prone codebase.

This will make it more clear when to add statistics collection in VisitorState.

RELNOTES: n/a

3f713f817ea7f74608fb07dcf193d2e646d6c261

-------

<p> Support %tZ format string conversions

Fixes #1196

RELNOTES: N/A

b878fd53b744ef146c08af07276988c6e65b12b6

-------

<p> Fix bug in WellKnownMutability incorrectly identifying proto1 as proto2.

It turns out Suppliers.typeFromString("com.google.io.protocol.ProtocolMessage") returns a com.google.io.protocol.ProtocolMessage<T>, and com.google.io.protocol.ProtocolMessage<SomeOtherType> is not assignable to this, so it always returns false.

It is, however, assignable to the raw type.

RELNOTES: Fix bug in WellKnownMutability incorrectly identifying proto1 as proto2.

ee5426146525f45371c14761c860caed0a50a3ec

-------

<p> Add documentation to CodeTransformer

RELNOTES: n/a

5da7bce1230830eab07ec3743650bfef96136bef

-------

<p> Replace Type#equals with TypeSymbol comparisons.

RELNOTES: N/A

91fd7db5531472b8973b848a6673473116e7683a

-------

<p> Update javadocs in HidingFieldPositiveCases1

RELNOTES: n/a

593dcb918e704e97df4a76443e07d42bd15bf4aa

-------

<p> Skip testonly targets for MissingSummary and non-public elements.

Given Javadocs aren't required at all, it seems a bit harsh to complain for these cases.

RELNOTES: N/A

4b7405544d71f769af6a561b990c8be9ba54622f

-------

<p> Demote ComplexBooleanConstant to a warning

RELNOTES: ComplexBooleanConstant is now a warning instead of an error

32fd6022f8189672d610684acfd88bfa5c5b649a

-------

<p> Use a different internal Mockito class for static method invocation

This class actually does not exist in 1.10.19. Use a different class that
exists in both 1.9.5 and 1.10.19.i

RELNOTES:

274367961871c5d884835f453df2011d54a551b9

-------

<p> Discourage Type#equals

RELNOTES: Check to discourage Type#equals

6a6831458f59bf43fc2884236eecc498605aad67


ronshapiro:
The previously failing test class, now that's empty, has this error:
>   WellKnownMutabilityTest.initializationError » InvalidTestClass Invalid test cl...

PR #1201, opened 2018-12-13
Update installation for pre-jdk8 to include javac dependency

nglorioso:
It looks like we had some issues trying to set up Error Prone on a local project.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -85,12 +85,19 @@ support:
           <plugin>
             <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-compiler-plugin</artifactId>
+            <dependencies>
+              <dependency>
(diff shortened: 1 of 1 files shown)


tbroyer:
You shouldn't need this, as com.google.errorprone:javac is (still) a dependency of com.google.errorprone:error_prone_core.

And ErrorProne requires at least JDK 8.

msridhar:
An explicit dependency is not required, but the Maven installation instructions for creating the jdk8 profile reference a `javac.version` variable that is not defined anywhere.  Creating that var with value `9+181-r4173-1` fixes the problem.

nglorioso:
Closing this as @cushon fixed this with https://github.com/google/error-prone/commit/6bf0676be44661ca440cd6566a2462d9bc4e5ff1

PR #1218, opened 2019-01-23
Moe Sync

ronshapiro:

Commits:
=====
<p> Add a bugpattern to detect methods which return a mixture of immutable and mutable types.

RELNOTES: [MixedMutabilityReturnType] flag methods which return both mutable and immutable types

1e67d9fcef7e932d9975de13ee5ccab3ed0a129b

-------

<p> UnescapedEntity: don't suggest wrapping in {@code } in code/pre tags if the enclosing content contains HTML entities or annotations.

Instead, escape 'em individually. Otherwise, we can produce fixes that just make life worse.

RELNOTES: N/A

9c4c0d5156e28b110debe1b5901f815c60c25641

-------

<p> consistently check for declaration annotations in addition to type annotations in nullness qualifier inference.
Includes recognizing checkerframework's @NonNullDecl
RELNOTES: None.

b39ec304145c33d9db4f2cfcdf04dee3c5ae0a5a

-------

<p> Remove an unnecessary indefinite article.

As it stands, this produces messages like this:
"AutoValue instances should be deeply immutable. Therefore, we recommend returning a ImmutableMap instead." Saying "...return ImmutableMap" reads just as well in addition to being grammatically correct.

eb5ee7cf25cff59c74874e4935e0cd262a8540c1

-------

<p> AssignmentToMock: discourage manual assignment to fields annotated with @Mock.

RELNOTES: AssignmentToMock: discourage manual assignment to fields annotated with @Mock.

5710e010a50a7eac9a4289bde0af1139dfb137bb

-------

<p> Use non-deprecated GuardedBy annotation

b30d92ef7012238ff518dd60b89465741ffc147f

-------

<p> Prevent user-written code from referring to Dagger-generated implementation code

RELNOTES: New check that prevents user-written code from referring to Dagger-generated implementation code

4a0147a79dfcea25a3bbbc923955bd884aeef175

-------

<p> Update MethodCanBeStatic docs

RELNOTES: N/A

e68592b05e41c2659667d66b98eaa3ff15bb81fe

-------

<p> Put the side-effect-removing fix first in ModifiedButNotUsed and Unused.

Also simplify the heuristics used in SideEffectAnalysis; they're very
limited, and wrong anyway, so we might as well just consider any method
invocation as possibly side-effecting.

RELNOTES: N/A

63b685fc2d3f2f70cbfaf89ddd111c3e15d3a61e

### DIFF
--- a/check_api/src/main/java/com/google/errorprone/dataflow/nullnesspropagation/Nullness.java
+++ b/check_api/src/main/java/com/google/errorprone/dataflow/nullnesspropagation/Nullness.java
@@ -127,7 +127,8 @@ public enum Nullness implements AbstractValue<Nullness> {
   // TODO(kmb): Correctly handle JSR 305 @Nonnull(NEVER) etc.
   private static final Predicate<String> ANNOTATION_RELEVANT_TO_NULLNESS =
       Pattern.compile(
-              ".*\\.((Recently)?Nullable(Decl)?|(Recently)?NotNull|Nonnull|NonNull|CheckForNull)$")
+              ".*\\.((Recently)?Nullable(Decl)?|(Recently)?NotNull(Decl)?|NonNull(Decl)?|Nonnull|"
+                  + "CheckForNull)$")
           .asPredicate();
 
   private static final Predicate<String> NULLABLE_ANNOTATION =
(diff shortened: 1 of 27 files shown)


eaftan:
@ronshapiro I think the CI failures were caused by your commit 688eadac46fe97268cc697b705d67de0f0dffe0f.  I think you may need to add a dep to the pom.xml.  Could you PTAL?

ronshapiro:
Looks like my dagger change broke the tests since we don't process the test code with annotation processors.

Should we exclude that test in Maven? Or should I pull in the dagger processor here?

PR #1239, opened 2019-03-02
Have Refaster resolve classes against the appropriate modules

Stephan202:

but fail on JDK 9+.

### DIFF
--- a/core/src/main/java/com/google/errorprone/refaster/Inliner.java
+++ b/core/src/main/java/com/google/errorprone/refaster/Inliner.java
@@ -84,8 +84,7 @@ public final class Inliner {
       throws CouldNotResolveImportException {
     try {
       Symbol symbol =
-          JavaCompiler.instance(context)
-              .resolveIdent(symtab().java_base, qualifiedClass.toString());
+          JavaCompiler.instance(context).resolveBinaryNameOrIdent(qualifiedClass.toString());
       if (symbol.equals(symtab().errSymbol) || !(symbol instanceof ClassSymbol)) {
         throw new CouldNotResolveImportException(qualifiedClass);
       } else {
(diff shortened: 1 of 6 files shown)


Stephan202:
@lowasser would it be possible to have this change merged in time for the next release? Without it a significant portion of Refaster templates simply don't match on JDK 9+.

knutwannheden:
Does this effectively mean that I cannot use Refaster in a Java 11 code base, where I also want to write Refaster rules pertaining to the types in said code base?

Stephan202:
@knutwannheden indeed, many interesting Refaster rules will not match without these changes. I wonder whether inside Google no code has been migrated to Java 9+ yet, because otherwise they should have hit this same issue :thinking: .

Within our company we work around this issue (and a small number of others) by maintaining a [fork](https://github.com/PicnicSupermarket/error-prone). It is available on [Jitpack](https://jitpack.io/#PicnicSupermarket/error-prone). I tagged version `v2.3.4-picnic-2` this weekend; it has a [delta](https://github.com/PicnicSupermarket/error-prone/commits/v2.3.4-picnic-2) of 6 commits commits compared to the official version 2.3.4.

knutwannheden:
@Stephan202 Thanks for the heads up. I am new to Refaster and it took me a while to find this PR. I thought I had hit some corner case with generics.

I will give your fork a try. Thanks for the tip!

knutwannheden:
@Stephan202 Using your forked error-prone build I was indeed able to write and apply Refaster rules against Java 11 code. Very nice! Now it would be nice if this PR could also be integrated...

Xcelled:
This bit us, too. It seemed like refaster just wasn't working due to lack of any obvious error. I had to spend a while spelunking through refaster to find what was wrong, and had created a band-aid fix before seeing this PR. It'd be great to get this merged, as more of the world moves to JDK 9+.

Issue #1234, opened 2019-02-24
ErrorProne crashed: MixedMutabilityReturnType


shevek:
> ATTENTION! Please read and follow:
> - if this is a _question_ about Error Prone, send it to error-prone-discuss@googlegroups.com
> - if this is a _bug_ or _feature request_, fill the form below as best as you can.

### Description of the problem / feature request:

ErrorProne crashed

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

We don't have anything self-contained, we just can't build our code.

### What version of Error Prone are you using?

```
     error-prone version: 2.3.3
     BugPattern: MixedMutabilityReturnType
     Stack Trace:
     java.lang.IllegalArgumentException: Replacement{range=[2200..2253), replaceWith=ImmutableSet.copyOf(new ArrayUnenforcedSet<>(Collections.singleton(edge)))} conflicts with existing replacement Replacement{range=[2200..2253), replaceWith=ImmutableList.copyOf(new ArrayUnenforcedSet<>(Collections.singleton(edge)))}
        at com.google.errorprone.fixes.Replacements.add(Replacements.java:109)
        at com.google.errorprone.fixes.SuggestedFix.getReplacements(SuggestedFix.java:90)
        at com.google.errorprone.fixes.AppliedFix$Applier.apply(AppliedFix.java:71)
        at com.google.errorprone.JavacErrorDescriptionListener.lambda$new$0(JavacErrorDescriptionListener.java:71)
        at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
        at java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:175)
        at java.util.Collections$2.tryAdvance(Collections.java:4717)
        at java.util.Collections$2.forEachRemaining(Collections.java:4725)
        at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
        at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
        at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
        at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:499)
        at com.google.errorprone.JavacErrorDescriptionListener.onDescribed(JavacErrorDescriptionListener.java:84)
[...]

PR #1258, opened 2019-04-12
Moe Sync

ronshapiro:

Commits:
=====
<p> Add another Dagger 1 codegen case to RefersToDaggerCodegen

RELNOTES: n/a

1fcb40ec9d3115ce48cd0637632be50d6c7bba0f

-------

<p> Remove some deprecated matchers

RELNOTES: `Matchers.instanceMethod` is no longer supported; prefer `MethodMatchers.instanceMethod()`
RELNOTES: `Matchers.isDescendantOfMethod(className, methodName)` is no longer supported; prefer `MethodMatchers.instanceMethod().onDescendantOf(className).named(methodName)`

f49391125d41b352d91ce13d992b13b5d2e3eeae

-------

<p> RELNOTES: UnusedVariable: don't warn when nulling out variables. This is often done to allow the garbage collector to collect an expensive object.

e263f354c37d92881f2df2aee6ee5eb437fb752e

-------

<p> RELNOTES: Tests for external issue #1234.

Sadly this didn't turn up the actual bug, but having coverage for these seems good.

97767bbe295dfe3bd5a346ba0b086f043a950259

-------

<p> Preserve non-constant messages in fail() in exception tests
RELNOTES: N/A

b2730be18dbb3a6ae7ae8db32b12ed3dff909785

-------

<p> RELNOTES: In ComparableType, require an exact match, not just that the argument type of Comparable can be cast to the owning class type

38c65b4a7346ee7ff1ed1330442ec0879e8c02f8

-------

<p> RELNOTES: Remove the exemption for GoogleLoggers from UnusedVariable.

784101c4d1e24ed00ee62a2f612ca49a043be361


ronshapiro:
This is just a partial sync, since we've been seeing intermittent MOE issues internally. Those seem to be resolved around ~late March, so I recommend that we consider merging this even if the tests are broken, that way we can hopefully make continual progress to getting internal -> external fully synced. Even if tests are broken, fixing+resyncing may be quite difficult all as one PR.

PR #1259, opened 2019-04-14
Bundle Guava's failureaccess in the with-dependencies JAR

### LINKED ISSUE #1238: java.lang.NoClassDefFoundError: com/google/common/util/concurrent/internal/InternalFutureFailureAccess while using the "with-dependencies.jar"

java.lang.NoClassDefFoundError: com/google/common/util/concurrent/internal/InternalFutureFailureAccess while using the "with-dependencies.jar"  

Run error-prone with the "with-dependencies.jar" (only this is broken)

2.3.3

I could workaround the issue while adding the missing "failureaccess" jar (version 1.0.1) from guava to the path, but this should bundled with the with-dependencies.jar. I guess after splitting guava in one on the latest versions this was not added to the "with-dependencies.jar"  packager.

tbroyer:
Fixes #1238

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -367,6 +367,7 @@
                   <include>com.github.stephenc.jcip:jcip-annotations</include>
                   <include>org.pcollections:pcollections</include>
                   <include>com.google.guava:guava</include>
+                  <include>com.google.guava:failureaccess</include>
                   <include>com.google.auto:auto-common</include>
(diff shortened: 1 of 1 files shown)


eaftan:
@tbroyer could you PTAL at the CI failures?

tbroyer:
AFAICT these are the same as on `master`, probably due to the update to Mockito 2.

PR #1264, opened 2019-04-18
Sync master 2019/04/18

cushon:


cushon:
See https://github.com/google/error-prone/pull/1259 for the CLA.

Issue #1260, opened 2019-04-15
IndexOutOfBoundsException in OptionalMapToOptional


facboy:
> ATTENTION! Please read and follow:
> - if this is a _question_ about Error Prone, send it to error-prone-discuss@googlegroups.com
> - if this is a _bug_ or _feature request_, fill the form below as best as you can.

### Description of the problem / feature request:

```
An unhandled exception was thrown by the Error Prone static analysis plugin.
[ERROR]      Please report this at https://github.com/google/error-prone/issues/new and include the following:

error-prone version: 2.3.3
[ERROR]      BugPattern: OptionalMapToOptional
[ERROR]      Stack Trace:
[ERROR]      java.lang.IndexOutOfBoundsException: Index: 1, Size: 0
[ERROR]         at com.sun.tools.javac.util.List.get(List.java:490)
[ERROR]         at com.google.errorprone.bugpatterns.OptionalMapToOptional.matchMethodInvocation(OptionalMapToOptional.java:92)
[ERROR]         at com.google.errorprone.scanner.ErrorProneScanner.processMatchers(ErrorProneScanner.java:433)
[ERROR]         at com.google.errorprone.scanner.ErrorProneScanner.visitMethodInvocation(ErrorProneScanner.java:732)
[ERROR]         at com.google.errorprone.scanner.ErrorProneScanner.visitMethodInvocation(ErrorProneScanner.java:150)
[ERROR]         at com.sun.tools.javac.tree.JCTree$JCMethodInvocation.accept(JCTree.java:1644)
[ERROR]         at com.sun.source.util.TreePathScanner.scan(TreePathScanner.java:82)
[ERROR]         at com.google.errorprone.scanner.Scanner.scan(Scanner.java:71)
[ERROR]         at com.google.errorprone.scanner.Scanner.scan(Scanner.java:45)
[ERROR]         at com.sun.source.util.TreeScanner.visitReturn(TreeScanner.java:469)
[ERROR]         at com.google.errorprone.scanner.ErrorProneScanner.visitReturn(ErrorProneScanner.java:803)
[ERROR]         at com.google.errorprone.scanner.ErrorProneScanner.visitReturn(ErrorProneScanner.java:150)
[ERROR]         at com.sun.tools.javac.tree.JCTree$JCReturn.accept(JCTree.java:1548)
[ERROR]         at com.sun.source.util.TreePathScanner.scan(TreePathScanner.java:82)
[...]

Issue #1253, opened 2019-03-30
NullableDereference ignores methods that throw if null in call path


bobtiernay-okta:
### Description of the problem / feature request:

The following common [AssertJ](http://joel-costigliola.github.io/assertj/) `isNotNull()` assertion will not pass `NullableDereference`:

```java
Integer value = null;
assertThat(value)
   .isNotNull()
   .isEqualTo(1);
```

Similarly, neither will this:

```java
Integer x = null;
Integer y = Preconditions.checkNotNull(x);
int result = 1 + y;
```

And this:

```java
class Foo {
  Integer value;
  
  String get() {
      Preconditions.checkNotNull(value);
      return value.toString();
  }
}
```

### Feature requests: what underlying problem are you trying to solve with this feature?

In order to prevent false positives, it null analysis should examine the call path for null checks.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

See above.

### What version of Error Prone are you using?

2.3.3

###  Have you found anything relevant by searching the web?

No.

kevin1e100:
Thanks for your report.  Note this checker is experimental and incomplete at this point; please don't rely on it.

There's actually two issues here:
1. Based on checkNotNull's annotations, the checker should warn when passing null to it.
1. Ideally as mentioned above the checker would assume checkNotNull's result to be non-null.

I say "ideally" b/c what's being asked here is ironically somewhat difficult to do once there is an error, which is the case here.

You should be able to side-step this issue by using `Verify.verifyNotNull` instead.  That method is more appropriate for "sanitizing" possibly-null values, as its annotations allow `null` values to be passed in.  

NullableDereference uses available annotations to do its work and the two methods are annotated [subtly](https://github.com/google/guava/blob/master/guava/src/com/google/common/base/Preconditions.java) [differently](https://github.com/google/guava/blob/master/guava/src/com/google/common/base/Verify.java) to allow nullable inputs only for `verifyNotNull`.

cushon:
d73fa691c8a60cf88934b585b4d908c9f07c868b

PR #1267, opened 2019-04-29
Moe Sync

ronshapiro:

Commits:
=====
<p> Ignore @Bind-annotated fields in UnnecessaryLambda

99a530bc0b0046d2dec8674f136d09223729afe5

-------

<p> Add notes about the semantic impact of applying the suggested fix.

9dd6b9da48397271e3cbf01c376ae7d514c782ef

-------

<p> Migrate Truth Subjects from no-arg check() to the overload that accepts a description.

The overload that accepts a description generally produces better failure messages:
- The first line of the message it produces is something like: "value of: myProto.getResponse()" (where "getResponse()" is taken from the provided description)
- The last line of the message it produces is something like: "myProto was: response: query was throttled" (the full value of myProto)
- And the existing text goes in between.

Additional motivation: We have deprecated the no-arg overload in open-source Truth.

Note that the new overload is available externally as of Truth 0.40.

c65852596e6690b1a83c5147f279891a00b5796a

-------

<p> Simplify @Generated handling

cec21ea4d9871baf2e40774351f0bfc6bea9f19c

-------

<p> Recognize bound type parameters when null-checking method call arguments

discovered due to https://github.com/google/error-prone/issues/1253

cac4cc5e441a8138550310e83e82bbdb0c353ae5

-------

<p> Don't flag AutoValue generated code for LiteEnumValueOf check.

RELNOTES: Don't flag AutoValue generated code for LiteEnumValueOf check.

68247869d7d048fbe411c50ac8b21a27c3ebedbf

-------

<p> OptionalMapToOptional: don't crash on raw types.

Somewhat interesting this didn't come up internally.

Fixes external #1260

7313f66d4bd0d927783a2c40363d458a0466784e

-------

<p> Add a check to detect anonymous classes that can be replaced with method references

fa4b1e2c87c957f5e4b2e2a7eadf12430b4f4222

-------

<p> Fix the finding position for unused reassignments.

36ba43a2d21acc7c44b3d4b0251afc5feee5e230

-------

<p> Test cases for lite proto enum value of matches enclosed in lite and full proto

925e1cd2255661e55bfeb543662b8f46f9a05e7b

-------

<p> Update to Truth 0.44.

f4b4364d051da61f88b53df1810f9785262a1fcd

-------

<p> Don't flag lite enums wrapped in GeneratedMessage

RELNOTES: Don't flag lite enums wrapped in GeneratedMessage

480718d659910e13823dace5ca11f9c2f1828e82

-------

<p> Rename SelfEquals test files for Guava and remove custom TEST_RESOURCES setting in BUILD file

GuavaSelfEquals -> SelfEqualsGuava

3d0fe65e7b34ee84bf8da5a9bcf83b7e656b8b9a

-------

<p> SuggestedFixes#renameMethodInvocation: tokenize so as to handle type args.

5886fb2217a924ad40d6fe7480a92739c023b882

### DIFF
--- a/docs/bugpattern/UnnecessaryAnonymousClass.md
+++ b/docs/bugpattern/UnnecessaryAnonymousClass.md
@@ -1,0 +1,54 @@
+Prefer method references to anonymous classes that implement functional
+interfaces.
+
+That is, prefer this:
+
+```java {.good}
+private static Bar getBar(Foo foo) {
+  return BarService.lookupBar(foo, defaultCredentials());
+}
+
+...
+
+return someStream().map(MyClass::getBar)....;
+```
+
+to this:
+
+```java
+private static final Function<Foo, Bar> GET_BAR_FUNCTION =
+    new Function<Foo, Bar>() {
+      @Override
+      public Bar apply(Foo foo) {
+        return BarService.lookupBar(foo, defaultCredentials());
+      }
+    };
+
+...
+
+return someStream().map(GET_BAR_FUNCTION)....;
+```
+
+Advantages of using a method include:
+
+*   It avoids hardcoding a named dependency on the functional interface type
+    (e.g., is it a `com.google.common.base.Function` or a
+    `java.util.function.Function`?)
+*   It is easier to name the method than the constant.
+*   It is more natural to test the method than the constant.
+*   If the behavior is nontrivial, it's more natural to write javadoc for the
+    method.
+
+Be aware that this change is not purely syntactic: it affects the semantics of
+your program in some small ways. In particular, evaluating the same method
+reference twice is not guaranteed to return an identical object.
+
+This means that, first, inlining the reference instead of using a constant may
+cause additional memory allocations - usually this very slight performance cost
+is worth the improved readability, but use your judgment if the performance
+matters to you.
+
+Secondly, if the correctness of your program depends on reference equality,
+inlining the method reference may break you. Ideally, you should *not* depend on
+reference equality, but if you are doing so, consider not making this change.
+
--- a/docs/bugpattern/UnnecessaryLambda.md
+++ b/docs/bugpattern/UnnecessaryLambda.md
@@ -34,3 +34,17 @@ Advantages of using a method include:
 *   If the behavior is nontrivial, it's more natural to write javadoc for the
     method.
 
+Be aware that this change is not purely syntactic: it affects the semantics of
+your program in some small ways. In particular, evaluating the same method
+reference twice is not guaranteed to return an identical object.
+
+This means that, first, inlining the reference instead of storing a lambda may
+cause additional memory allocations - usually this very slight performance cost
+is worth the improved readability, but use your judgment if the performance
+matters to you.
+
+Secondly, if the correctness of your program depends on reference equality of
+your lambda, inlining it may break you. Ideally, you should *not* depend on
+reference equality for a lambda, but if you are doing so, consider not making
+this change.
+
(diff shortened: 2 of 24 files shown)


ronshapiro:
FYI @eaftan

docs/bugpattern/UnnecessaryLambda.md
```diff
@@ -34,3 +34,17 @@ Advantages of using a method include:
...
     method.
 
+Be aware that this change is not purely syntactic: it affects the semantics of
+your program in some small ways. In particular, evaluating the same method
+reference twice is not guaranteed to return an identical object.
+
+This means that, first, inlining the reference instead of storing a lambda may
+cause additional memory allocations - usually this very slight performance cost
+is worth the improved readability, but use your judgment if the performance
+matters to you.
```
jbduncan: @amalloy I'm lost by these docs for a couple of reasons.

Firstly, I can't see why changing a stored lambda to a method reference would in of itself cause a lack of a guarantee to return an identical object, because AFAIU stored lambdas have the exact same problem...

...And secondly, I'm struggling to tell why using a method reference over a stored lambda would cause more memory to be used. 🤔

Would you mind expanding on these in the docs?

amalloy: Consider

```java
private IntSupplier hash = this::hashCode;
public boolean same() {
  return hash == hash;
}
```

The comparison is guaranteed to be true, and we will allocate exactly one lambda object over the lifetime of this object.

Now apply the rewrite rule, yielding:

```java
public boolean same() {
  return this::hashCode == this::hashCode;
}
```

which is not guaranteed to be the same, because two different lambdas may be allocated for each call to `same`. This is the difference the new paragraphs warn about. Perhaps it would be worth adding this example to the documentation.

jbduncan: Ah okay, so am I right to understand that in the second code sample, the two `this::hashCode` lambdas aren't guaranteed to be equal to one another because the JVM could assign them different object identities?

This wasn't clear to me from the docs, because I thought they were talking about the lambdas' _outputs_ rather than the lambdas themselves. So, although I don't know how one could better word the docs, I agree that it would be worth including this example to help prevent confusion. :)

jbduncan: This would also explain to me why the refactoring would increase memory usage slightly, because naturally if both lambdas have different object identities, then we end up not with _one_ object but _two_ of them.

docs/bugpattern/UnnecessaryAnonymousClass.md
```diff
@@ -0,0 +1,54 @@
...
+reference twice is not guaranteed to return an identical object.
+
+This means that, first, inlining the reference instead of using a constant may
+cause additional memory allocations - usually this very slight performance cost
+is worth the improved readability, but use your judgment if the performance
+matters to you.
+
+Secondly, if the correctness of your program depends on reference equality,
+inlining the method reference may break you. Ideally, you should *not* depend on
+reference equality, but if you are doing so, consider not making this change.
```
jbduncan: Hi @cushon - similarly to my discussion with @amalloy over at https://github.com/google/error-prone/pull/1267/files/99a530bc0b0046d2dec8674f136d09223729afe5..9dd6b9da48397271e3cbf01c376ae7d514c782ef#r279655102, I wonder it would be useful to have an example here to help illustrate why the change introduced by this refactoring is not purely syntactic. :)

eaftan:
@ronshapiro Ack, I'm looking at the test failures.

cushon:
re: test failures, we needed to update the version of compile-testing to match truth. I did that internally, the next sync should work.

PR #1295, opened 2019-06-15
Update WORKSPACE in Bazel example

the1derer:
v2.3.4 is not available

### DIFF
--- a/examples/plugin/bazel/WORKSPACE
+++ b/examples/plugin/bazel/WORKSPACE
@@ -1,6 +1,6 @@
 maven_jar(
     name = "error_prone",
-    artifact = "com.google.errorprone:error_prone_ant:2.3.4-SNAPSHOT",
+    artifact = "com.google.errorprone:error_prone_ant:2.3.2-SNAPSHOT",
     repository = "https://oss.sonatype.org/content/repositories/snapshots",
 )
 


tbroyer:
2.3.4 hasn't been released yet, but this is about 2.3.4-SNAPSHOT.
The problem here actually is that error_prone_ant no longer exists. AFAICT,  it's been replaced with error_prone_core's with-dependencies variant (along with a few unbundled dependencies, due to license concerns)

the1derer:
@tbroyer 
Thanks for the reply.
I actually first tried "error_prone_core-with-dependencies-2.3.4-SNAPSHOT"
But I was not able to find artifact-name for the same.
Can you provide the artifact-name that I can use for the "error_prone_core-with-dependencies-2.3.4-SNAPSHOT"?
Thanks.

tbroyer:
Try with `com.google.errorprone:error_prone_core:jar:with-dependencies:2.3.4-SNAPSHOT`, or possibly `com.google.errorprone:error_prone_core:2.3.4-SNAPSHOT:jar:with-dependencies`.

That being said, it looks like this would eventually have to be moved to using the Skylark `jvm_maven_import_external` (see https://docs.bazel.build/versions/master/skylark/backward-compatibility.html#remove-native-maven-jar or https://github.com/bazelbuild/bazel/issues/6799)

PR #1294, opened 2019-06-12
Moe Sync

ronshapiro:
This code has been reviewed and submitted internally. Feel free to discuss on the PR and we can submit follow-up changes as necessary.

Commits:
=====
<p> Replace google.github.io/dagger with dagger.dev

bb00ae194a3d126033eaa3c5d68f16707755d2b2

-------

<p> Flag instances of Duration.of(long, ChronoUnit) where the ChronoUnit is not an exact duration.

#badtime #goodtime

RELNOTES: Flag instances of Duration.of(long, ChronoUnit) where the ChronoUnit is not an exact duration.

221fde7d65d5cdd6ed0bd5a9a1fa7e5f4ab48acd

-------

<p> Fix an incorrect UnnecessaryAnonymousClass refactoring

don't refactor anonymous classes that implement the FI via a superclass.

0faf698a1563222527195638befec3429a4cdf80

-------

<p> Fix NPE on JDK13+ when InconsistentHashCode visits a break statement with a label.

Fixes #1287

RELNOTES: n/a

7bb2a776f215389d23637fe087e9a93ebba7ebe4

-------

<p> Expand SelfAssignment to cover the Precondition-style APIs in c.g.c.time.Durations,c.g.protobuf.util.{Durations,Timestamps}.

5ddc53ff5138ef6777855d1ea51d73b049507340

-------

<p> Allow ChronoUnit.DAYS explicitly in Duration.of(long, TemporalUnit).

(It's explicitly called out in the javadocs, but I missed it the first read.)

Also turn off the checker for desugar'ed java.time.

254f3f29e96da06e2aad0cb5851c5b7c2a237137

-------

<p> Rename DurationOfLongTemporalUnit to DurationTemporalUnit and cover duration.plus(long, TemporalUnit) and duration.minus(long, TemporalUnit).

Also ban calls to Instant APIs where the TemporalUnit is not supported.

beef3d89bc77ba88daddd867f87e154dbebe0dfc

-------

<p> Provide an example of using proto3's primitive wrappers.

e2cb9fed7eced3226fb6131715e4dbfafe3ef992

-------

<p> Migrate org.mockito.Matchers#any* to org.mockito.ArgumentMatchers

The former is deprecated and replaced by the latter in Mockito 2. However, there is a
functional difference: ArgumentMatchers will reject `null` and check the type
if the matcher specified a type (e.g. `any(Class)` or `anyInt()`). `any()` will
remain to accept anything.

[...]

PR #1296, opened 2019-06-18
Moe Sync

ronshapiro:

Commits:
=====
<p> Add PreferDurationOverload which suggests using a Duration-based overload instead of a (long, TimeUnit)-based overload. For example, if an API has both overloads:
  setDeadline(long, TimeUnit)
  setDeadline(Duration)
then this checker will suggest callers invoke the Duration-based overload instead of the (long, TimeUnit)-based overload.

8ca41e650edc2ca7c83531616cab43afe9be21e5

-------

<p> Extend UnnecessaryBoxedVariable to cover parameters of non-overriding/overrideable methods.

96a131dff72f387778feb4fa2ee51226555a87ca

-------

<p> Add support for JodaDuration -> JavaDuration suggestions to PreferDurationOverload.

4cda29d8dcdfed16955b9f924a05f0124ac1e928

-------

<p> Add TimeUnit.MICROSECONDS support to PreferDurationOverload.

7c6405245fd38e998cf96cbc7c03a6a0d337cc8f

-------

<p> Add support for unwrapping Joda Durations that are created inline. For example:

foo(org.joda.time.Duration.standardSeconds(42)) -> foo(java.time.Duration.ofSeconds(42)).

89c5f49c9fbeb510614397ec583f7acaac813131

-------

<p> Make example code look more like a real Subject.

in response to:
https://github.com/google/error-prone/pull/1294#discussion_r293021750

a4e476fe001a1df158024a7c2c37317928ac8366

-------

<p> "Use checked exceptions for recoverable conditions ..." is item 70 in Effective Java 3rd edition.

e772749dd6a6e8502c85ce7d96c71ef2c68a29be

-------

<p> Remove an example annotation

35f33e92ab6f4c7b7771a35f2454b614c945d09c

-------

<p> Fix typo

d580e7eee3672c8aaa4bd5b7d66f9371800678b0

-------

<p> Add @fails as an invalid block tag which will be ignored.

This is pretty popular (generally for Promises) and led to a right flurry of not useful clicks.

19b287335198193f3568e3f117ac64400fd23794

-------

<p> Fix use of array toString.

556c604186277d0c08a85092bb102df3ff26c15c

-------

<p> AbstractToString: match #log calls in Flogger and @FormatMethods.

Invert the operation of the check so it looks for methods that implicitly toString, and then matches the arguments, rather than attempting to do it the other way around.

3c2b2155076f8450dd5e91247029ec1066a6b80c

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/AbstractToString.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/AbstractToString.java
@@ -20,28 +20,32 @@ import static com.google.errorprone.matchers.Description.NO_MATCH;
 import static com.google.errorprone.matchers.Matchers.anyOf;
 import static com.google.errorprone.matchers.Matchers.instanceMethod;
 import static com.google.errorprone.matchers.Matchers.staticMethod;
+import static com.google.errorprone.matchers.Matchers.symbolHasAnnotation;
+import static com.google.errorprone.util.ASTHelpers.getReceiver;
+import static com.google.errorprone.util.ASTHelpers.getSymbol;
+import static com.google.errorprone.util.ASTHelpers.getType;
 
 import com.google.errorprone.VisitorState;
-import com.google.errorprone.bugpatterns.BugChecker.IdentifierTreeMatcher;
-import com.google.errorprone.bugpatterns.BugChecker.MemberSelectTreeMatcher;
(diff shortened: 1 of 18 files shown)


ronshapiro:
One of the tests uses Flogger, but the pom wasn't updated to include a flogger (`google-extensions`) dep. I'm surprised this passed internally...

PR #1302, opened 2019-06-27
Moe Sync

ronshapiro:

Commits:
=====
<p> Fix an instance of TypeParameterQualifier

and make the check unsuppressible.

The behaviour of `P.Builder` and `Message.Builder` is identical at compile
time and at run time.

f1f420bfd778427556e8720efcc9dcd1ce4802a9

-------

<p> Avoid references to internal Error Prone matcher interfaces

54528c250a5f317da06c105ea88dd740466eab65

-------

<p> Don't suggest an infinite recursion when PreferDurationOverload
suggests calling a method inside its implementation.

80c33b4e44cc8397445b30475b7e470c90df3f08

-------

<p> Prepare static analysis for when ProtoSubject no longer extends ProtoFluentAssertion.

That change will bring ProtoSubject in line with IterableOfProtosSubject and the other types in its package.

ed7781cd30de75b09719df305166ad50fdd43f32

-------

<p> Recognize calls of default interface methods

fd15337b802e91a37c90df6872be755509d6faf6

-------

<p> Add a ERROR ErrorProne check when new object/array creation is wrapped in
checkNotNull or verifyNotNull

RELNOTES: new ERROR which flags when new object/array creation is wrapped in
checkNotNull or verifyNotNull

c10a89dc20f86f0df882d97745b5ba8baaf735ae

### DIFF
--- a/check_api/src/main/java/com/google/errorprone/matchers/Matchers.java
+++ b/check_api/src/main/java/com/google/errorprone/matchers/Matchers.java
@@ -162,6 +162,11 @@ public class Matchers {
     return (tree, state) -> tree.getKind() == kind;
   }
 
+  /** Matches an AST node of a given kind, for example, an Annotation or a switch block. */
+  public static <T extends Tree> Matcher<T> kindAnyOf(Set<Kind> kinds) {
+    return (tree, state) -> kinds.contains(tree.getKind());
+  }
+
   /** Matches an AST node which is the same object reference as the given node. */
   public static <T extends Tree> Matcher<T> isSame(Tree t) {
     return (tree, state) -> tree == t;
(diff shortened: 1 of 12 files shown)


jbduncan:
In commit https://github.com/google/error-prone/commit/c10a89dc20f86f0df882d97745b5ba8baaf735ae, I wonder if `Objects.requireNonNull` should also be checked against.

@sumitbhagwani @ronshapiro Do you have any thoughts regarding this?

ronshapiro:
@jbduncan seems reasonable. Send a PR?

jbduncan:
@ronshapiro Sure! I'll see if I can send a PR this weekend. :)

sumitbhagwani:
I already have a change submitted for it internally. It'll be out when we do a new moe sync. Thanks.

jbduncan:
Thanks @sumitbhagwani, much appreciated! :+1:

sumitbhagwani:
@jbduncan https://github.com/google/error-prone/pull/1306 has the changes. 
More specifically : https://github.com/google/error-prone/commit/8e788f7178a84a46474287d883745f11c688e731

jbduncan:
Thank you very much @sumitbhagwani. :)

PR #1313, opened 2019-07-18
Clear diagnostics after testing in CompilationTestHelper

pkoenig10:
When a `CompilationTestHelper` is reused after a positive test case, it is possible for a false positive test case to pass. This is because we do not clear the `DiagnosticTestHelper` after calling `doTest`. So a subsequent false positive test will pass if the diagnostics is that same as the previous positive test case.


cushon:
Thanks for the report. These instances aren't really intended to be use-used, but I agree we should either support re-use or else detect it and report an error if `doTest` is called twice on the same instance.

If you have tests where it would be helpful to support the former, can you expand on your use-case?

cushon:
Tentatively resolving with https://github.com/google/error-prone/commit/8bfc2f894bc73e1107aafd2ab5e3951f0295bc58.

PR #1329, opened 2019-08-07
[core] Add support for ConcurrentHashMap in AbstractAsKeyOfSetOrMap

blemale:
Add support for `ConcurrentHashMap` in `AbstractAsKeyOfSetOrMap`, thus the error patterns `ArrayAsKeyOfSetOrMap` and `ProtosAsKeyOfSetOrMap` are available for `ConcurrentHashMap`.

PR #1331, opened 2019-08-16
Moe Sync

netdpb:

Commits:
=====
<p> [core] Add support for ConcurrentHashMap in
AbstractAsKeyOfSetOrMap

Fixes #1329

24536708ccad497bf91293d2c634d7e9cf15e4ee

-------

<p> LiteProtoToString: provide a separate description and fix for enums.

24b34be34eff5ef79b310adca6c9df8afc3ed140

-------

<p> s/override/overriding method/

Perhaps this is more clear?

c5e4b17c4ce1a6b97f4feaf7616389237d152831

-------

<p> Support lambdas for @CompileTimeConstant

d2b9abae75fc348081f316b43175296a4e0833b1

-------

<p> Update checker to reflect current API name

fbaa109493c6333f1d91373cd2d06c175f3de471

-------

<p> Improve performance of annotation scanning

The focal point here was SuppresionInfo.withExtendedSuppressions. This
method is called for every matcher at every AST node, so any expense
is greatly magnified. This commit caches the Name objects ahead of
time instead of looking them up each time. Additionally, ASTHelpers
has gained a method for testing a set of annotations in one go, so
clients don't have to call hasAnnotation many times.

The net result is that the method is around twice as fast as it was
before, saving us 1.1% of total javac time.

e7e61b354c4df14d017dc296657db3d33f517287

-------

<p> Cache the Name used in Matchers.hasAnnotation

This was a smaller performance win than I expected, only 0.18% of
javac time, but still better than the status quo.

6c2106a52218ce8f998d42652d296977d873973d

-------

<p> Improve next/prev/last statement matches

b3953f5de9a8858d56208e3925da330c68297226

-------

<p> Check for misuses of Collections.disjoint in CollectionIncompatibleType

as well as Sets.{union,disjoint}.

e19c4dd81a4a6f5e88a2ca0d76989f02b3d6add5

-------

<p> Add bugpattern to discourage AtomicReference#compareAndSet with primitive types.

c05e3d31f32515040b36c70a1fb08b75cc6b04b3


netdpb:
@blemale Can you please follow the instructions in https://github.com/google/error-prone/pull/1331#issuecomment-522032118: leave a comment here that contains only `@googlebot I consent.`? Thanks.

blemale:
@googlebot I consent.

cpovirk:
Thanks. Merged in https://github.com/google/error-prone/pull/1332 instead.

Issue #1342, opened 2019-08-30
java.lang.NoSuchFieldError: label


zielint0:
### Description of the problem / feature request:

IDEA 2019.2.1, Windows 10, JDK 12

Error:(123, 21) java: An unhandled exception was thrown by the Error Prone static analysis plugin.
     Please report this at https://github.com/google/error-prone/issues/new and include the following:
  
     error-prone version: 2.3.3
     BugPattern: Finally
     Stack Trace:
     java.lang.NoSuchFieldError: label
  	at com.google.errorprone.bugpatterns.Finally$FinallyJumpMatcher.<init>(Finally.java:171)
  	at com.google.errorprone.bugpatterns.Finally.matchBreak(Finally.java:79)
  	at com.google.errorprone.scanner.ErrorProneScanner.processMatchers(ErrorProneScanner.java:433)
  	at com.google.errorprone.scanner.ErrorProneScanner.visitBreak(ErrorProneScanner.java:513)
  	at com.google.errorprone.scanner.ErrorProneScanner.visitBreak(ErrorProneScanner.java:150)
  	at jdk.compiler/com.sun.tools.javac.tree.JCTree$JCBreak.accept(JCTree.java:1577)
  	at jdk.compiler/com.sun.source.util.TreePathScanner.scan(TreePathScanner.java:82)
  	at com.google.errorprone.scanner.Scanner.scan(Scanner.java:71)
  	at com.google.errorprone.scanner.Scanner.scan(Scanner.java:45)
  	at jdk.compiler/com.sun.source.util.TreeScanner.scanAndReduce(TreeScanner.java:91)
  	at jdk.compiler/com.sun.source.util.TreeScanner.scan(TreeScanner.java:106)
  	at jdk.compiler/com.sun.source.util.TreeScanner.visitBlock(TreeScanner.java:249)
  	at com.google.errorprone.scanner.ErrorProneScanner.visitBlock(ErrorProneScanner.java:507)
  	at com.google.errorprone.scanner.ErrorProneScanner.visitBlock(ErrorProneScanner.java:150)
  	at jdk.compiler/com.sun.tools.javac.tree.JCTree$JCBlock.accept(JCTree.java:1038)
  	at jdk.compiler/com.sun.source.util.TreePathScanner.scan(TreePathScanner.java:82)
  	at com.google.errorprone.scanner.Scanner.scan(Scanner.java:71)
  	at com.google.errorprone.scanner.Scanner.scan(Scanner.java:45)
[...]

Issue #1301, opened 2019-06-22
An unhandled exception was thrown by the Error Prone static analysis plugin.


hollandjake:
### Description of the problem / feature request:

```java
Error:(52, 33) java: An unhandled exception was thrown by the Error Prone static analysis plugin.
     Please report this at https://github.com/google/error-prone/issues/new and include the following:
  
     error-prone version: 2.3.3
     BugPattern: Finally
     Stack Trace:
     java.lang.NoSuchFieldError: label
  	at com.google.errorprone.bugpatterns.Finally$FinallyJumpMatcher.<init>(Finally.java:171)
  	at com.google.errorprone.bugpatterns.Finally.matchBreak(Finally.java:79)
  	at com.google.errorprone.scanner.ErrorProneScanner.processMatchers(ErrorProneScanner.java:433)
  	at com.google.errorprone.scanner.ErrorProneScanner.visitBreak(ErrorProneScanner.java:513)
  	at com.google.errorprone.scanner.ErrorProneScanner.visitBreak(ErrorProneScanner.java:150)
  	at jdk.compiler/com.sun.tools.javac.tree.JCTree$JCBreak.accept(JCTree.java:1577)
  	at jdk.compiler/com.sun.source.util.TreePathScanner.scan(TreePathScanner.java:82)
  	at com.google.errorprone.scanner.Scanner.scan(Scanner.java:71)
  	at com.google.errorprone.scanner.Scanner.scan(Scanner.java:45)
  	at jdk.compiler/com.sun.source.util.TreeScanner.scanAndReduce(TreeScanner.java:91)
  	at jdk.compiler/com.sun.source.util.TreeScanner.scan(TreeScanner.java:106)
  	at jdk.compiler/com.sun.source.util.TreeScanner.scanAndReduce(TreeScanner.java:114)
  	at jdk.compiler/com.sun.source.util.TreeScanner.visitCase(TreeScanner.java:372)
  	at com.google.errorprone.scanner.ErrorProneScanner.visitCase(ErrorProneScanner.java:521)
  	at com.google.errorprone.scanner.ErrorProneScanner.visitCase(ErrorProneScanner.java:150)
  	at jdk.compiler/com.sun.tools.javac.tree.JCTree$JCCase.accept(JCTree.java:1293)
  	at jdk.compiler/com.sun.source.util.TreePathScanner.scan(TreePathScanner.java:82)
  	at com.google.errorprone.scanner.Scanner.scan(Scanner.java:71)
[...]

Issue #1257, opened 2019-04-11
Finally BugPattern breaks on Java 12


shemnon:
### Description of the problem / feature request:

FindBugs is broken with Java 12

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

This repo: https://github.com/PegaSysEng/pantheon/tree/1.0.3 with OpenJDK 12 produces this stack trace on Windows and MacOS

```
<root>/pantheon/util/src/main/java/tech/pegasys/pantheon/util/uint/UInt256Bytes.java:215: error: An unhandled exception was thrown by the Error Prone static analysis plugin.
        if (diff == 0) break;
                       ^
     Please report this at https://github.com/google/error-prone/issues/new and include the following:
  
     error-prone version: 2.3.3
     BugPattern: Finally
     Stack Trace:
     java.lang.NoSuchFieldError: label
        at com.google.errorprone.bugpatterns.Finally$FinallyJumpMatcher.<init>(Finally.java:171)
        at com.google.errorprone.bugpatterns.Finally.matchBreak(Finally.java:79)
        at com.google.errorprone.scanner.ErrorProneScanner.processMatchers(ErrorProneScanner.java:433)
        at com.google.errorprone.scanner.ErrorProneScanner.visitBreak(ErrorProneScanner.java:513)
        at com.google.errorprone.scanner.ErrorProneScanner.visitBreak(ErrorProneScanner.java:150)
        at jdk.compiler/com.sun.tools.javac.tree.JCTree$JCBreak.accept(JCTree.java:1577)
        at jdk.compiler/com.sun.source.util.TreePathScanner.scan(TreePathScanner.java:82)
        at com.google.errorprone.scanner.Scanner.scan(Scanner.java:71)
        at com.google.errorprone.scanner.Scanner.scan(Scanner.java:45)
        at jdk.compiler/com.sun.source.util.TreeScanner.scanAndReduce(TreeScanner.java:91)
        at jdk.compiler/com.sun.source.util.TreeScanner.visitIf(TreeScanner.java:445)
        at com.google.errorprone.scanner.ErrorProneScanner.visitIf(ErrorProneScanner.java:640)
        at com.google.errorprone.scanner.ErrorProneScanner.visitIf(ErrorProneScanner.java:150)
        at jdk.compiler/com.sun.tools.javac.tree.JCTree$JCIf.accept(JCTree.java:1500)
[...]

PR #1343, opened 2019-09-01
Fix for Java12+ compatibility

thomasgalvin:
This commit fixes issues #1342, #1335, #1301, #1257, and #1249

Prior to this change, static analysis of code containing a `break` statement would fail when running on Java version 12+, with an error similar to the following:

```
BugPattern: Finally
Stack Trace:
java.lang.NoSuchFieldError: com/sun/tools/javac/tree/JCTree$JCBreak.label
```

This was due to a change in the class `com.sun.tools.javac.tree.JCTree$JCBreak`. In Java 12+, `label` is no longer a field, and is instead a calculated value. This change simply changes `com.google.errorprone.bugpatterns.Finally` to call `JCBreak.getLabel()` instead of `JCBreak.label`.

This change maintains compatibility with prior versions of Java.

 Changes to be committed:
	modified:   core/src/main/java/com/google/errorprone/bugpatterns/Finally.java

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/Finally.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/Finally.java
@@ -168,7 +168,7 @@ public class Finally extends BugChecker
     }
 
     public FinallyJumpMatcher(JCBreak jcBreak) {
-      this.label = jcBreak.label;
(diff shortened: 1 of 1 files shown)


thomasgalvin:
@googlebot I fixed it.

PR #1354, opened 2019-09-23
Moe Sync

netdpb:

Commits:
=====
<p> Add bugpattern (to JavaCodeClarity) to encourage treating RuntimeException et al as abstract.

b21ed7d5c754160a8865d149f0ee4b3d6f0b018e

-------

<p> SuggestedFixes.compilesWithFix: check that extraOptions is empty before fast-returning, because the extra options may affect compilation.

a4f28dda5c214b6cb2ff3862b818f7876a87120b

-------

<p> ExtendsAutoValue: Makes it illegal to extend an @AutoValue class.

RELNOTES=ExtendsAutoValue: Makes it illegal to extend an @AutoValue class.

8987fa34d8cc9c39b888c8b3339ac8a42668adac

-------

<p> Refactor out "nearby code" method in anticipation of using it for more checks
relating to Period.

RELNOTES: n/a

00381a4590501256d4794233cb1aee66f9a3347a

-------

<p> UnusedNestedClass: obey @UsedReflectively

7738ee25b9b6cd93858751db7081742b8f7f9e26

-------

<p> PreferJavaTimeOverload - avoid suggesting an infinite recursion when the method call is inside an override of a method that accepts the java.time object.

RELNOTES: n/a

73e9456b3771c3e63dafeac01d813796b04cfb7a

-------

<p> Add a check for accidental use of ^

i.e. 2 ^ 16 is not 65536

46a2c8acc1785124423e2926be7be9bfb6148c19

-------

<p> Fix a typo

66b9d831ab0db93f5b45a9d1676de4ad533981f7

-------

<p> DoNotCall: allow static non-final methods to have the annotation.

2b8ab0a9da27c6d2f46490bece4edb6da400a66e

-------

<p> PrivateConstructorForUtilityClass: rephrase description a bit to include main classes.

Drive-by modernise the Java a bit.

b2c39075b445b5fbc303065605e64c62352fb9f3

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/ExtendsAutoValue.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/ExtendsAutoValue.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2019 The Error Prone Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.errorprone.bugpatterns;
+
+import com.google.errorprone.BugPattern;
+import com.google.errorprone.BugPattern.SeverityLevel;
+import com.google.errorprone.VisitorState;
+import com.google.errorprone.bugpatterns.BugChecker.ClassTreeMatcher;
+import com.google.errorprone.matchers.Description;
+import com.google.errorprone.util.ASTHelpers;
+import com.sun.source.tree.ClassTree;
+
+/** Makes sure that you are not extending a class that has @AutoValue as an annotation. */
+@BugPattern(
+    name = "ExtendsAutoValue",
+    summary = "Do not extend an @AutoValue class in non-generated code.",
+    severity = SeverityLevel.WARNING)
+public final class ExtendsAutoValue extends BugChecker implements ClassTreeMatcher {
+
+  @Override
+  public Description matchClass(ClassTree tree, VisitorState state) {
+
+    if (!ASTHelpers.getGeneratedBy(ASTHelpers.getSymbol(tree), state).isEmpty()) {
+      // Skip generated code. Yes, I know we can do this via a flag but we should always ignore
+      // generated code, so to be sure, manually check it.
+      return Description.NO_MATCH;
+    }
+
+    if (tree.getExtendsClause() == null) {
+      // Doesn't extend anything, can't possibly be a violation.
+      return Description.NO_MATCH;
+    }
+    if (ASTHelpers.hasAnnotation(
+        ASTHelpers.getSymbol(tree.getExtendsClause()), "com.google.auto.value.AutoValue", state)) {
+      // Violation: one of its superclasses extends AutoValue.
+      return buildDescription(tree).build();
+    }
+
+    return Description.NO_MATCH;
+  }
+}
--- a/core/src/main/java/com/google/errorprone/bugpatterns/ThrowSpecificExceptions.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/ThrowSpecificExceptions.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright 2019 The Error Prone Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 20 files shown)


core/src/main/java/com/google/errorprone/bugpatterns/ThrowSpecificExceptions.java
```diff
@@ -0,0 +1,78 @@
...
+            + " generic exceptions forces any users of the API that wish to handle the failure"
+            + " mode to catch very non-specific exceptions that convey little information.",
+    providesFix = REQUIRES_HUMAN_ATTENTION,
+    severity = WARNING)
+public final class ThrowSpecificExceptions extends BugChecker implements NewClassTreeMatcher {
+  private static final ImmutableList<AbstractLikeException> ABSTRACT_LIKE_EXCEPTIONS =
+      ImmutableList.of(
+          AbstractLikeException.of(RuntimeException.class, IllegalStateException.class),
+          AbstractLikeException.of(Throwable.class, AssertionError.class),
+          AbstractLikeException.of(Error.class, AssertionError.class));
```
Stephan202: Just curious: was `Exception.class` left out on purpose? It may not have an obvious replacement, but Error Prone could still flag its use. WDYT?

core/src/main/java/com/google/errorprone/bugpatterns/ExtendsAutoValue.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+    if (ASTHelpers.hasAnnotation(
+        ASTHelpers.getSymbol(tree.getExtendsClause()), "com.google.auto.value.AutoValue", state)) {
+      // Violation: one of its superclasses extends AutoValue.
+      return buildDescription(tree).build();
+    }
+
+    return Description.NO_MATCH;
+  }
+}
```
jbduncan: @dplassgit @netdpb I'm not entirely clear on why this check was made as an error-prone check rather than built into AutoValue itself, and I wondered if you could enlighten me?

docs/bugpattern/XorPower.md
```diff
@@ -0,0 +1,4 @@
+The `^` binary XOR operator is sometimes mistaken for a power operator, but e.g.
+`2 ^ 10` evaluates to `0`, not `4`.
```
jbduncan: @cushon I'm not clear on how 2 to the power of 10 evaluates to 4. Did you mean to say that it evaluates to 1024, or am I missing something here? :)

jbduncan: @cushon Oops, sorry, I saw you fixed it in the next commit!

core/src/main/java/com/google/errorprone/bugpatterns/ExtendsAutoValue.java
tbroyer: This check runs on classes that are **not** processed by AutoValue, but extend AutoValue-annotated classes.

It could have been added as an error prone check to the AutoValue project, but then there's the case of extending an AutoValue class from another project while not using AutoValue yourself in your project, so you wouldn't have AutoValue in your annotation processor path, and would therefore not have the check either.

dplassgit: @jbduncan, @tbroyer answered for me. 

(The following is a synopsis of various justifications:)

In general extending an AutoValue class is not what you want. AutoValue classes are simple value classes that can contain simple intrinsic behavior but shouldn't require complex dependencies and shouldn't access static state.
Therefore, you should essentially never need an alternative (subclassed) implementation of an AutoValue class, whether hand-written or generated by a mocking framework. If your behavior has enough complexity (or dependencies) that it actually needs to be mocked or faked, split it into a separate type that is not a value type. Otherwise it permits an instance with "real" behavior and one with "mock/fake" behavior to be equals, which does not make sense.

jbduncan: Thanks for explanation @dplassgit and @tbroyer, but I'm still unclear on why it was decided to make the check an error-prone check rather than a compile-time check by the AutoValue code generator itself.

Is it not possible to compile-check something at code generation time? :thinking:

tbroyer: It is, but only **if** you're generating code.

Examples:

project A has an `@AutoValue public abstract class A {}`; it includes auto-value-annotations in the classpath and auto-value (the processor) in the processor path. If you add `public final class AA extends extends A {}`, technically, yes, the processor could discover it (by looking at **all** the compiled classes, and not just the ones annotated with `@AutoValue`)

Now, introduce project B that has a dependency on project A. Project doesn't use AutoValue, so it doesn't have auto-value in its processor path (though it possibly inherits auto-value-annotations from project A, transitively, in its classpath). The how do you check that you don't have a `public class B extends A {}`?
By using a) an ErrorProne check that b) is shipped as part of ErrorProne (and not an AutoValue-specific plugin), you can guarantee that any project using ErrorProne will have the warning, whether they use AutoValue or not.

jbduncan: @tbroyer Ahh, now that makes more sense to me!

So in other words, we can't guarantee that every project that imports another AutoValue-annotations-using project will use the AutoValue code generator, and thus if we do the can't-extend-AutoValue-class check inside the code generator, we can't guarantee that non-AutoValue-using projects won't violate the check.

Cheers! :)

PR #1355, opened 2019-09-23
Moe Sync

netdpb:

Commits:
=====
<p> Avoid NPEing on symbols which have no enclosing class

See also b/123431414, which is about symbols with no enclosing package.

0f1c1a8b73f7f3e5749a05c7c6a29db281b93543


netdpb:
This is a spurious sync, probably because I've been testing the sync scripts. This commit has already been merged from another run.

Issue #1350, opened 2019-09-15
An unhandled exception was thrown by the Error Prone static analysis plugin


binkley:
The plugin requested I report the following (below).

This corresponds to https://github.com/binkley/spikes/tree/e4a1ef56d4166146ad9a9111a9431609cd5c032a/micronaut-hello-world.

The project which failed is a Gradle project in the `micronaut-hello-world` directory underneath the project root.

```
Please report this at https://github.com/google/error-prone/issues/new and include the following:
  
     error-prone version: 2.3.3
     BugPattern: UnusedVariable
     Stack Trace:
     java.lang.IllegalArgumentException: invalid replacement: [168, -1) ()
        at com.google.common.base.Preconditions.checkArgument(Preconditions.java:458)
        at com.google.errorprone.fixes.Replacement.create(Replacement.java:37)
        at com.google.errorprone.fixes.SuggestedFix$ReplacementFix.getReplacement(SuggestedFix.java:397)
        at com.google.errorprone.fixes.SuggestedFix.getReplacements(SuggestedFix.java:91)
        at com.google.errorprone.fixes.AppliedFix$Applier.apply(AppliedFix.java:71)
        at com.google.errorprone.JavacErrorDescriptionListener.lambda$new$0(JavacErrorDescriptionListener.java:71)
        at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
        at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:177)
        at java.base/java.util.Collections$2.tryAdvance(Collections.java:4747)
        at java.base/java.util.Collections$2.forEachRemaining(Collections.java:4755)
        at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
        at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
        at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:913)
        at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:578)
        at com.google.errorprone.JavacErrorDescriptionListener.onDescribed(JavacErrorDescriptionListener.java:84)
        at com.google.errorprone.ErrorProneAnalyzer.lambda$finished$1(ErrorProneAnalyzer.java:136)
[...]

Issue #1266, opened 2019-04-26
 java.lang.IndexOutOfBoundsException with lombok 1.18.6.0 


meifans:
> ATTENTION! Please read and follow:
> - if this is a _question_ about Error Prone, send it to error-prone-discuss@googlegroups.com
> - if this is a _bug_ or _feature request_, fill the form below as best as you can.

### Description of the problem / feature request:

```
[ERROR]      error-prone version: 2.3.3
[ERROR]      BugPattern: UnusedVariable
[ERROR]      Stack Trace:
[ERROR]      java.lang.IndexOutOfBoundsException
[ERROR]         at java.base/java.nio.HeapCharBuffer.subSequence(HeapCharBuffer.java:628)
[ERROR]         at java.base/java.nio.HeapCharBuffer.subSequence(HeapCharBuffer.java:43)
[ERROR]         at com.google.errorprone.fixes.SuggestedFixes.replaceIncludingComments(SuggestedFixes.java:1021)
[ERROR]         at com.google.errorprone.bugpatterns.UnusedVariable.buildUnusedVarFixes(UnusedVariable.java:387)
[ERROR]         at com.google.errorprone.bugpatterns.UnusedVariable.matchCompilationUnit(UnusedVariable.java:239)
[ERROR]         at com.google.errorprone.scanner.ErrorProneScanner.processMatchers(ErrorProneScanner.java:433)
[ERROR]         at com.google.errorprone.scanner.ErrorProneScanner.visitCompilationUnit(ErrorProneScanner.java:541)
[ERROR]         at com.google.errorprone.scanner.ErrorProneScanner.visitCompilationUnit(ErrorProneScanner.java:150)
[ERROR]         at jdk.compiler/com.sun.tools.javac.tree.JCTree$JCCompilationUnit.accept(JCTree.java:591)
[ERROR]         at jdk.compiler/com.sun.source.util.TreePathScanner.scan(TreePathScanner.java:56)
[ERROR]         at com.google.errorprone.scanner.Scanner.scan(Scanner.java:55)
[ERROR]         at com.google.errorprone.scanner.ErrorProneScannerTransformer.apply(ErrorProneScannerTransformer.java:43)
[ERROR]         at com.google.errorprone.ErrorProneAnalyzer.finished(ErrorProneAnalyzer.java:151)
[ERROR]         at jdk.compiler/com.sun.tools.javac.api.MultiTaskListener.finished(MultiTaskListener.java:120)
[...]

Issue #1250, opened 2019-03-26
lombok causes an IndexOutOfBoundsException in UnusedVariable


mrt181:
### Description of the problem / feature request:

`./gradlew clean compileJava` throws this error

```
Service.java:1: error: An unhandled exception was thrown by the Error Prone static analysis plugin.                                                    
package io.mine.service.grpc;
^
     Please report this at https://github.com/google/error-prone/issues/new and include the following:
	 
     error-prone version: 2.3.3
     BugPattern: UnusedVariable
     Stack Trace:
     java.lang.IndexOutOfBoundsException
        at java.nio.HeapCharBuffer.subSequence(HeapCharBuffer.java:580)
        at java.nio.HeapCharBuffer.subSequence(HeapCharBuffer.java:42)
        at com.google.errorprone.fixes.SuggestedFixes.replaceIncludingComments(SuggestedFixes.java:1021)
        at com.google.errorprone.bugpatterns.UnusedVariable.buildUnusedVarFixes(UnusedVariable.java:387)
        at com.google.errorprone.bugpatterns.UnusedVariable.matchCompilationUnit(UnusedVariable.java:239)
        at com.google.errorprone.scanner.ErrorProneScanner.processMatchers(ErrorProneScanner.java:433)
        at com.google.errorprone.scanner.ErrorProneScanner.visitCompilationUnit(ErrorProneScanner.java:541)
        at com.google.errorprone.scanner.ErrorProneScanner.visitCompilationUnit(ErrorProneScanner.java:150)
```

### Feature requests: what underlying problem are you trying to solve with this feature?

none, its applied by error-prone `2.3.3`

### What version of Error Prone are you using?
`errorproneJavacVersion = "9+181-r4173-1"`
`errorproneVersion = "2.3.2"`
`errorpronePluginVersion = "0.7.1"`

`openjdk version "1.8.0_192"`

###  Have you found anything relevant by searching the web?

reverting back to `2.3.2` fixes the problem.
I tried to exclude the BugPattern but that does not work:

```
gradle.projectsEvaluated {
    tasks.withType(JavaCompile) {
        options.encoding = "UTF-8"
        options.annotationProcessorPath = configurations.errorprone
        options.errorprone.enabled = true
        options.errorprone.disableWarningsInGeneratedCode = true
        options.errorprone.excludedPaths = ".*/build/gen.*/.*"
        options.errorprone.errorproneArgs = ["-Xep:ParameterName:OFF",
[...]

PR #1379, opened 2019-10-17
Fix exception from unused variable bug pattern from Lombok.

chrismgrayftsinc:
This should fix all such exceptions due to variables added by
annotation processors that are unused.

### DIFF
--- a/check_api/src/main/java/com/google/errorprone/InvalidTokenException.java
+++ b/check_api/src/main/java/com/google/errorprone/InvalidTokenException.java
@@ -1,0 +1,23 @@
+/*
+ * Copyright 2019 The Error Prone Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


chrismgrayftsinc:
@googlebot I signed it!

chrismgrayftsinc:
@tbroyer This PR is for #1266 and #1350 (which I noticed you were looking at today).  I am working with my employer to get the CLA correct.

adamlehenbauer:
+1 this change; this bug is preventing me from adopting error-prone

chrismgrayftsinc:
Hi @cushon I was on parental leave when you asked about the CLA.  Let me know if this patch would still be helpful and I can prioritize getting it signed.

cushon:
@chrismgrayftsinc there's been some progress on the underlying issue in #1250 and https://github.com/rzwitserloot/lombok/issues/2691

PR #1391, opened 2019-10-28
maven:  use 'https' instead of 'http'

sullis:


hazendaz:
@Stephan202 All of maven bits can be 'https' but requires this added to release plugin '<addSchema>false</addSchema>'.

See [here](https://github.com/hazendaz/base-parent/blob/master/pom.xml) with project already using https for maven bits for this purpose.

hazendaz:
@Stephan202  Glad to pass on some knowledge :)  As long as the schema is present to start with, what that flag does is tell maven to not delete and re-add during the release process.  Basically maven is trying to fix invalid information and I guess probably since they changed it a long time ago its preferable to be the proper one.  The 'hack', if you will, came directly from the maven team when I was checking with them feasibility of removing all 'http' so global searches on real items that need resolved were much easier to find especially in 100s of poms :)  While there were zero plans at that time to change and I suspect that is still the case due to same reasons you mention, this was a great compromise.  All the IDE's still behave properly as well which was mostly to be expected.

PR #1396, opened 2019-11-01
MOE Sync 2019-11-01

netdpb:

Commits:
=====
<p> maven:  use 'https' instead of 'http'

Fixes #1391

b40559db1bf1d00cf8a4294a9c0a61ec81a48693

-------

<p> Do not report toString calls on lite protos for info logging

46df469afb987d53fae0e60557e962ac02a9ce9e

-------

<p> InterruptedExceptionSwallowed: flag methods which declare they throw Exception, but actually throw an InterruptedException.

ea8406f1fc0e3e640b5303b888aa1198285d79bc

### DIFF
--- a/check_api/pom.xml
+++ b/check_api/pom.xml
@@ -26,7 +26,7 @@
     <repository>
       <id>alt-repo1</id>
       <name>repo1.maven.org</name>
-      <url>http://repo1.maven.org</url>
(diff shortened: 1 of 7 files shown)


netdpb:
Adding cla:yes because b40559db1bf1d00cf8a4294a9c0a61ec81a48693 comes from @sullis's PR #1391.

PR #1398, opened 2019-11-05
Bundle caffeine in the with-dependencies JAR

lgemeinhardt:
Add caffeine in with-dependencies JAR, broken after https://github.com/google/error-prone/commit/bfb1789f5597a30abef367b88113bd61e023d86b
Similar to https://github.com/google/error-prone/pull/1259 and https://github.com/google/error-prone/issues/1238


mernst:
`diffutils-1.3.0.jar` is also needed, to support `-XepPatchChecks`.

cushon:
Thanks for the fix, this PR is subsumed by #1587.

PR #1425, opened 2019-11-27
Rejuvenate log levels.

saledouble:
Here's a reissue of https://github.com/google/error-prone/pull/1268 with a new version of our tool. The tool made better transformations. Again, we'd appreciate any feedback and are willing to make further changes if you wish to incorporate our PR into your project.

Settings
---
We have several analysis settings. We can vary these settings and rerun if you desire. The settings we are using in this pull request are:

Treat CONFIG levels as a category and not a traditional level, i.e., our tool ignores these log levels.
Never lower the logging level of logging statements within catch blocks.
Never lower the logging level of logging statements within if statements.
Never lower the logging level of logging statements containing certain (important) keywords.
Never raise the logging level of logging statements without particular keywords in their messages.
Avoid log wrapping by disregarding logging statements contained in if statements mentioning log levels.
The greatest number of commits from HEAD evaluated: 4000.
The head at the time of analysis was: 2118ba35b0efe2fdfe4b1af35d1967ed935b6988

### DIFF
--- a/check_api/src/main/java/com/google/errorprone/RefactoringCollection.java
+++ b/check_api/src/main/java/com/google/errorprone/RefactoringCollection.java
@@ -194,7 +194,7 @@ class RefactoringCollection implements DescriptionListener.Factory {
         fileDestination.writeFile(file);
       } catch (IOException e) {
         logger.log(
-            Level.WARNING,
+            Level.SEVERE,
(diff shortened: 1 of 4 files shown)


cushon:
Thanks for the suggestion, I think the existing log levels are appropriate here.

PR #1429, opened 2019-12-02
fix #1428 InterruptedExceptionSwallowed handles CatchTree throw statements

### LINKED ISSUE #1428: InterruptedExceptionSwallowed.getThrownExceptions incorrectly handles catch blocks

InterruptedExceptionSwallowed.getThrownExceptions incorrectly handles catch blocks when there are multiple catch trees. As written, the function assumes each catch is capable of handling errors thrown by previous catch blocks, which is not the case.

The bug checker should take into account how TryTrees are executed in java.

I will push a failing test shortly.

df4cdefbf351c76098562c2d83eebf4cafaa8eea

No.

carterkozak:
fix #1428

Previously the check assumed that exceptions could be handled by
subsequent catch blocks, which is incorrect.

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/InterruptedExceptionSwallowed.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/InterruptedExceptionSwallowed.java
@@ -236,17 +236,25 @@ public final class InterruptedExceptionSwallowed extends BugChecker
       scan(tree.getFinallyBlock(), null);
       Set<Type> fromBlock = thrownTypes.pop();
       getThrownTypes().addAll(fromBlock);
+      // The above getCatches scan isn't quite accurate when exceptions are thrown from earlier catch blocks.
+      // It's necessary to prune exceptions thrown by resources and the main block, but doesn't adequately
+      // capture those thrown from catch blocks.
+      for (CatchTree catchTree : tree.getCatches()) {
+        ScanThrownTypes nestedScanner = new ScanThrownTypes(state);
+        catchTree.getBlock().accept(nestedScanner, null);
+        getThrownTypes().addAll(nestedScanner.getThrownTypes());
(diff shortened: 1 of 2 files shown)


carterkozak:
Relatedly, it would be incredibly helpful if the `getThrownExceptions` function was publicly available, perhaps from ASTHelpers.java! Any chance we can make that happen?

graememorgan:
Thanks for the report and the fix! We tend to copy PRs into internal commits, so I've done that. I changed the implementation a bit: I think we can achieve the same thing more simply with two loops over the catch trees:

```
    @Override
    public Void visitTry(TryTree tree, Void unused) {
      thrownTypes.push(new HashSet<>());
      inResources = true;
      scan(tree.getResources(), null);
      inResources = false;
      scan(tree.getBlock(), null);
      // Make two passes over the `catch` blocks: once to remove caught exceptions, and once to
      // add thrown ones. We can't do this in one step as an exception could be caught but later
      // thrown.
      for (CatchTree catchTree : tree.getCatches()) {
        Type type = getType(catchTree.getParameter());
        for (Type unionMember : extractTypes(type)) {
          getThrownTypes().removeIf(thrownType -> types.isAssignable(thrownType, unionMember));
        }
      }
      for (CatchTree catchTree : tree.getCatches()) {
        scan(catchTree.getBlock(), null);
      }
      scan(tree.getFinallyBlock(), null);
      Set<Type> fromBlock = thrownTypes.pop();
      getThrownTypes().addAll(fromBlock);
      return null;
    }
```

> Relatedly, it would be incredibly helpful if the getThrownExceptions function was publicly available, perhaps from ASTHelpers.java! Any chance we can make that happen?

Sure, I'll do that after.

carterkozak:
Fantastic, thanks @graememorgan!

PR #1433, opened 2019-12-04
UnnecessaryCheckNotNull correctly qualifies Preconditions

carterkozak:
Previously other imports or nested classes named Preconditions
caused the suggested fix to fail.


cushon:
Thanks for the report! This was resolved by https://github.com/google/error-prone/commit/b920acfb06376caaa674f3c3af1983d3e42d27d9.

Issue #1377, opened 2019-10-16
PreferJavaTimeOverload does not play well with AssertJ assertThat


lgemeinhardt:
### Description of the problem / feature request:

PreferJavaTimeOverload does not play well with AssertJ assertThat

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

```
package sample;

import static org.assertj.core.api.Assertions.assertThat;

import org.junit.Test;

public class SampleTest {

	@Test
	public void testSomething() {
		assertThat(1 + 2).isEqualTo(3);
	}

}

```
raises

```
sample\SampleTest.java:11: warning: [PreferJavaTimeOverload] If the numeric primitive (1 + 2) represents a Instant, please call assertThat(Instant) instead.
     		assertThat(1 + 2).isEqualTo(3);
```

Could we may exclude AssertJ assertThat from this check to avoid tons of @SuppressWarnings("PreferJavaTimeOverload")?

### What version of Error Prone are you using?

2.3.4 SNAPSHOT (error_prone_core-2.3.4-20191015.160021-47-with-dependencies.jar)

kluever:
We don't use AssertJ inside of Google, so we haven't run into this, but this is definitely reasonable.

tbroyer:
I suppose you'll have the same issue with Truth once you add java.time support to Truth8 :wink:

kluever:
@tbroyer Yup, that's true - we're still [trying to figure out](https://github.com/google/truth/issues/420) if adding `java.time` support to Truth makes sense :-)

kluever:
Perhaps a good heuristic to use would be: "if there's >2 overloads for the method, then don't flag it".

PR #1436, opened 2019-12-05
Don't fire PreferJavaTimeOverload on custom AssertJ assertThat() overloads.

### LINKED ISSUE #1435: PreferJavaTimeOverload fails on AssertJ extensions which handle time

PreferJavaTimeOverload fails eroniously on assertj assertion classes which have time overloads, or extend `org.assertj.core.api.Assertions`.

Consider allowing methods named `assertThat` which return subtypes of `org.assertj.core.api.Assert` to pass validation rather than whitelisting only `org.assertj.core.api.Assertions.assertThat`

Easiest way to reproduce this is by subclassing Assertions (nonfinal utility class, obviously it's bad practice to extend static utility classes in modern java, but this occurs relatively frequently in the wild).

2.3.4

no.

carterkozak:
Fixes #1435
Related to #1377 and f1df86d66b92aeea6e9f5fa9263e366034b68209


carterkozak:
it's unfortunate there is no dependency on assertj, writing a test is difficult.

barorion:
`org.assertj.core.api.BDDAssertions` is a BDD style variant of `org.assertj.core.api.Assertions`. It's assertion method is `then()` (instead of `assertThat()`).
_PreferJavaTimeOverload_ should probably be muted for it too...

PR #1448, opened 2019-12-13
Add missing androidx support where appropriate

ZacSweers:
This adds analogous androidx classes alongside existing support library classes. This also adds an `isAndroidXCompatible` option to state and uses it for appropriate `NullnessFixes`.


core/src/main/java/com/google/errorprone/bugpatterns/nullness/NullnessFixes.java
```diff
@@ -44,7 +45,7 @@ private static String getQualifiedName(VisitorState state, SuggestedFix.Builder
     Symbol sym = FindIdentifiers.findIdent("Nullable", state, KindSelector.VAL_TYP);
     String defaultType =
         state.isAndroidCompatible()
-            ? "android.support.annotation.Nullable"
+            ? (state.isAndroidXCompatible() ? "androidx.annotation.Nullable" : "android.support.annotation.Nullable" )
```
ronshapiro: I'm not sure this is an intended use of these types of flags. The ErrorProne team would be able to state authoritatively, but perhaps switching this to a check for if the type is available in the VisitorState would be more appropriate?

@awturner

ZacSweers: An alternative would just be to accept a "nullableAnnotation" option to the build itself, similar to the way the android option works.

ZacSweers:
I'm unfortunately unable to properly test this, as master test sources seem to be largely broken?

image

image

image

core/src/main/java/com/google/errorprone/bugpatterns/nullness/NullnessFixes.java
eaftan: I like the idea of passing in a flag with the fully qualified name of the nullableAnnotation to use.  I would not do it the way we do for -XDandroidCompatible though, I would use the Error Prong flags functionality: https://errorprone.info/docs/flags#pass-additional-info-to-bugcheckers

How about a flag like `-XepOpt:Nullness:NullableAnnotation`?

ZacSweers: That sounds good to me! I'll rework this PR to use that approach

ZacSweers:
@eaftan updated with a new `-XepOpt:Nullness:NullableAnnotation` flag. Now nullness changes in this PR are:
- Default is `javax.annotation.Nullable`
- Custom fully qualified annotation as specified by the above flag
- Controlled by the `NullnessFixes` class
- The dagger `ProvidesNull` checker is updated to use `NullnessFixes` (CC @ronshapiro, this is the one I was telling you about). This also adds a new overload to accept a spacer suffix to the `makeFix` call, as this checker wants to always put the annotation on a new line.

Behavior changes worth mentioning:
- No longer tries to use an android annotation if the annotation error prone option is set. Must be set via the custom flag.
- `ProvidesNull` will use whatever `NullnessFixes` suggests, rather than always using `javax.annotation.Nullable`.

Let me know if/how you'd like to see tests for the integration and the API changes I made to support it. Similarly, I could also separate the NullnessFixes change to a separate PR since it's a wider scope.

PR #1449, opened 2019-12-15
ConstructorInvokesOverridable: fix some false positives

### LINKED ISSUE #656: Allow more fine-grained suppression of `ConstructorInvokesOverridable`
Consider the following code:
```java
final class Container {
    @SuppressWarnings("ConstructorInvokesOverridable")
    abstract static class Test1 {
        private final Class<?> type = myType();

        abstract Class<?> myType();
    }

    abstract static class Test2 {
        @SuppressWarnings("ConstructorInvokesOverridable")
        private final Class<?> type = myType();

        abstract Class<?> myType();
    }

    abstract static class Test3 {
        @SuppressWarnings("ConstructorInvokesOverridable")
        Test3() {
            myType();
        }

        abstract Class<?> myType();
    }
}
```

Running with `-Xep:ConstructorInvokesOverridable:ERROR` yields:
```
Container.java:11: error: [ConstructorInvokesOverridable] Constructors should not invoke overridable methods.
        private final Class<?> type = myType();
                                            ^
    (see http://errorprone.info/bugpattern/ConstructorInvokesOverridable)
Container.java:19: error: [ConstructorInvokesOverridable] Constructors should not invoke overridable methods.
            myType();
                  ^
    (see http://errorprone.info/bugpattern/ConstructorInvokesOverridable)
2 errors
```

As you can see, no error is emitted for `Test1`. But for the other two classes `@SuppressWarnings("ConstructorInvokesOverridable")` has no effect.

For classes with many fields or much code in initializer blocks/constructors, it'd be nice if `ConstructorInvokesOverridable` could be suppressed at a scope smaller than the whole class. (Refactoring the affected classes would be even better, of course, but alas.)

michaelhixson:
Do not match methods of anonymous classes.  Those methods can't be
overriden because those classes can't be extended.

Do not match methods of certain enum classes.  It is not possible to
override the methods of an enum class except in the (anonymous) class
bodies of its enum constants.  If all the enum constants lack class
bodies, then the enum class's methods can't be overridden.

Fix cases where the bug pattern would become confused about the receiver
of unqualified or this-qualified method invocations, and where it would
wrongly conclude that the method belonged to the instance under
construction and issue a warning.

Support `@SuppressWarnings("ConstructorInvokesOverridable")` on more
elements.  Fixes #656.

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/ConstructorInvokesOverridable.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/ConstructorInvokesOverridable.java
@@ -18,6 +18,8 @@ package com.google.errorprone.bugpatterns;
 
 import static com.google.errorprone.BugPattern.SeverityLevel.WARNING;
 
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Sets;
 import com.google.errorprone.BugPattern;
 import com.google.errorprone.VisitorState;
 import com.google.errorprone.util.ASTHelpers;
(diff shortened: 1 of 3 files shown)


eaftan:
Thanks for the fix!

PR #1452, opened 2019-12-16
Fix #1451: SuggestedFixes.renameMethodInvocation doesn't modify params

### LINKED ISSUE #1451: SuggestedFixes.renameMethodInvocation renames parameters with names matching the method
> ATTENTION! Please read and follow:
> - if this is a _question_ about Error Prone, send it to error-prone-discuss@googlegroups.com
> - if this is a _bug_ or _feature request_, fill the form below as best as you can.

`SuggestedFixes.renameMethodInvocation(tree, "updated", state)` results in:

```diff
- obj.foo(bar, foo);
+ obj.foo(bar, updated);
```

See above, methods are rewritten incorrectly.

See above example.

2.3.4

no.

carterkozak:
`renameMethodInvocation` tracks paren depth in order to avoid
erroneously renaming parameter identifiers instead of the method.


cushon:
This was fixed by https://github.com/google/error-prone/commit/b7f7883b5eb93d1cc8fc881cfce69bac4f46d98c.

PR #1455, opened 2019-12-17
Fix #1454: Matchers.booleanConstant(true) doesn't match Boolean.FALSE

### LINKED ISSUE #1454: Matchers.booleanConstant(true) should _not_ match Boolean.FALSE

`booleanConstant(true)` should not match both `Boolean.TRUE` and `Boolean.FALSE`.

Matchers match code that they're build specifically not to match.

Writing tests.

2.3.4

no

carterkozak:


eaftan:
This was a nice catch. Thanks for the contribution!

carterkozak:
Thanks for the review and merge @eaftan :-)

PR #1466, opened 2020-01-07
MOE Sync 2020-01-07

netdpb:

Commits:
=====
<p> UnnecessaryBoxedVariable: don't consider invocations of checkNotNull/verifyNotNull/requireNonNull as boxed usages.

d6fccbd52a4c13c2aafaaa4a5dc00d46dfa87a7b

-------

<p> InterruptedExceptionSwallowed: Handle classes that inherit implementations of AutoCloseable.close

8d74d2466cb4fff32f04e709d139903fc3a5d51e

-------

<p> Bump maven-compiler-plugin version to 3.8.1.

Fixes #1458

fd3c60a9485231e5bf4cce045a06b3b362596625

-------

<p> Various docs changes.

29789d6cbc69ab190c4c76366c54717ad2874eda

-------

<p> MissingSuperCall: Ignore super calls inside lambdas

follow-up to 96b175973bc741ab9f211f3e75e0b5123aa3f5ce

1850c4ac6f0fb25861b5244e37910b1ae667dd4a

-------

<p> AlmostJavadoc: skip desugared enum subclasses, as these lead to duplicate positions.

17abebd06d20df5ee724a3c6fdaf76df2ba3faeb

-------

<p> UnnecessaryCheckNotNull: be more careful about qualifying the precondition we want to use.

And encourage static importing it; that is the Java practices guidance after all.

This introduces a SuggestedFixes helper to get a name to refer to methods that we want to try to static import, but fall back to qualifying if necessary.

This addresses PR #1433, though not in the same way.

b920acfb06376caaa674f3c3af1983d3e42d27d9

### DIFF
--- a/check_api/src/main/java/com/google/errorprone/fixes/SuggestedFixes.java
+++ b/check_api/src/main/java/com/google/errorprone/fixes/SuggestedFixes.java
@@ -85,6 +85,7 @@ import com.sun.tools.javac.parser.Tokens.TokenKind;
 import com.sun.tools.javac.tree.DCTree;
 import com.sun.tools.javac.tree.JCTree;
 import com.sun.tools.javac.tree.JCTree.JCCompilationUnit;
+import com.sun.tools.javac.tree.JCTree.JCIdent;
 import com.sun.tools.javac.tree.TreeScanner;
 import com.sun.tools.javac.util.Context;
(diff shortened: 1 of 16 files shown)


netdpb:
Adding CLA:yes because fd3c60a comes from @sullis's PR #1458.

Issue #1265, opened 2019-04-24
error-prone fails when maven.compiler.release is set


rovarga:
> ATTENTION! Please read and follow:
> - if this is a _question_ about Error Prone, send it to error-prone-discuss@googlegroups.com
> - if this is a _bug_ or _feature request_, fill the form below as best as you can.

### Description of the problem / feature request:

error-prone fails in FieldCanBeFinal when running with JDK11 and maven.compiler.release set to 8:

[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /home/nite/odl/infrautils/common/util/src/main/java/org/opendaylight/infrautils/utils/concurrent/NamedSimpleReentrantLock.java:[8,1] An unhandled exception was thrown by the Error Prone static analysis plugin.
     Please report this at https://github.com/google/error-prone/issues/new and include the following:
  
     error-prone version: 2.3.3
     BugPattern: FieldCanBeFinal
     Stack Trace:
     java.lang.IllegalArgumentException: error: option -source cannot be used together with --release
        at jdk.compiler/com.sun.tools.javac.main.Arguments.reportDiag(Arguments.java:891)
        at jdk.compiler/com.sun.tools.javac.main.Arguments.lambda$handleReleaseOptions$0(Arguments.java:298)
        at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:183)
        at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:177)
        at java.base/java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:948)
        at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
        at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
        at java.base/java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:150)
        at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:173)
        at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java.base/java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:497)
        at jdk.compiler/com.sun.tools.javac.main.Arguments.checkOptionAllowed(Arguments.java:882)
[...]

PR #1473, opened 2020-01-12
Add `--release` flag support

Stephan202:
The java compiler [does not allow](https://hg.openjdk.java.net/jdk/jdk11/file/1ddf9a99e4ad/src/jdk.compiler/share/classes/com/sun/tools/javac/main/Arguments.java#l297) `-source` and `-target` to be specified explicitly when `--release` is, but [does add them](https://hg.openjdk.java.net/jdk/jdk11/file/1ddf9a99e4ad/src/jdk.compiler/share/classes/com/sun/tools/javac/main/Arguments.java#l315) in response to passing `--release`. As such `SuggestedFixes#compilesWithFix` must compensate for this by removing `-source` and `-target` before triggering another compilation round.

Fixes google/error-prone#1265.

### DIFF
--- a/check_api/src/main/java/com/google/errorprone/fixes/SuggestedFixes.java
+++ b/check_api/src/main/java/com/google/errorprone/fixes/SuggestedFixes.java
@@ -79,6 +79,7 @@ import com.sun.tools.javac.code.Symbol.ClassSymbol;
 import com.sun.tools.javac.code.Type;
 import com.sun.tools.javac.code.Types.DefaultTypeVisitor;
 import com.sun.tools.javac.main.Arguments;
+import com.sun.tools.javac.main.Option;
 import com.sun.tools.javac.parser.Tokens;
 import com.sun.tools.javac.parser.Tokens.Comment;
 import com.sun.tools.javac.parser.Tokens.TokenKind;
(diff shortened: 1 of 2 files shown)


check_api/src/main/java/com/google/errorprone/fixes/SuggestedFixes.java
```diff
@@ -1023,11 +1024,17 @@ public CharSequence getCharContent(boolean ignoreEncodingErrors) throws IOExcept
...
     for (String key : originalOptions.keySet()) {
       String value = originalOptions.get(key);
-      if (key.equals("-Xplugin:") && value.startsWith("ErrorProne")) {
+      if (key.equals(Option.PLUGIN.getPrimaryName()) && value.startsWith("ErrorProne")) {
         // When using the -Xplugin Error Prone integration, disable Error Prone for speculative
         // recompiles to avoid infinite recursion.
         continue;
       }
+      if ((key.equals(Option.SOURCE.getPrimaryName()) || key.equals(Option.TARGET.getPrimaryName()))
+          && originalOptions.isSet(Option.RELEASE)) {
```
Stephan202: This code can be optimized by factoring out `originalOptions.isSet(Option.RELEASE)` and creating a set containing `Option.SOURCE.getPrimaryName()` and `Option.TARGET.getPrimaryName()`. Decided against this because I figured it'd make this already-very-long method even more complex to follow. Let me know if you feel differently and I'll change it.

check_api/src/main/java/com/google/errorprone/fixes/SuggestedFixes.java
```diff
@@ -1023,11 +1024,17 @@ public CharSequence getCharContent(boolean ignoreEncodingErrors) throws IOExcept
     Options originalOptions = Options.instance(javacTask.getContext());
     for (String key : originalOptions.keySet()) {
       String value = originalOptions.get(key);
-      if (key.equals("-Xplugin:") && value.startsWith("ErrorProne")) {
```
cushon, line 1026: I'd rather use the string names of the flags here. `com.sun.tools.javac.main.Optionn` is an unsupported internal API that has changed in the past, and the literal flag names are part of javac's supported API.

Stephan202: Ack! Since you assigned the PR to yourself: want me to update this, or have you already applied those changes internally?

PR #1492, opened 2020-01-31
Remove more {.good} and {.bad} tags

### LINKED ISSUE #1487: Incorrectly rendered markdown in IdentityBinaryExpression

IdentityBinaryExpression markdown still includes `java {.good}` and `java {.bad}`. These tags were supposed to be removed in issue #962 (commit - e43a577723a52de9d000f18b553bc55616d9e5df). This leads to the markdown rendering incorrectly. 

http://errorprone.info/bugpattern/IdentityBinaryExpression

image

Fix the IdentityBinaryExpression markdown so the example is rendered correctly.

IdentityBinaryExpression includes `java {.good}` and `java {.bad}` tags. This causes issues rendering the markdown.

I'm not using Error Prone. This is about the errorprone.info documentation.

Yes. I found the related issue #962 and commit (e43a577723a52de9d000f18b553bc55616d9e5df). The commit (e43a577723a52de9d000f18b553bc55616d9e5df) does not include changes to `IdentityBinaryExpression.md`.

dpursehouse:
Also fix a couple of instances of .java that should not have the
leading dot.

Fixes #1487

### DIFF
--- a/docs/bugpattern/BigDecimalLiteralDouble.md
+++ b/docs/bugpattern/BigDecimalLiteralDouble.md
@@ -1,6 +1,6 @@
 BigDecimal's `double` can lose precision in surprising ways.
 
-```java {bad}
+```java
   // these are the same:
   new BigDecimal(0.1)
   new BigDecimal("0.1000000000000000055511151231257827021181583404541015625")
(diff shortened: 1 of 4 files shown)


nglorioso:
Merged via #1533

Issue #1504, opened 2020-02-06
PreferJavaTimeOverload generates false positive with jOOQ DSL.inline()


tbroyer:
### Description of the problem:

when using jOOQ's `DSL.inline(1)`, ErrorProne emits a warning (which fails our build because we use `-Werror`)
```
warning: [PreferJavaTimeOverload] If the numeric primitive (1) represents a Instant, please call DSL.inline(Instant) instead.
```

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Simply call `DSL.inline(1)`.

### What version of Error Prone are you using?

2.3.4

###  Have you found anything relevant by searching the web?

PreferJavaTimeOverload seems to check the method return type when looking for overloads, but in this case [`DSL.inline(int)`](https://www.jooq.org/javadoc/latest/org.jooq/org/jooq/impl/DSL.html#inline(int)) returns a `Param<Integer>` while [`DSL.inline(Instant)`](https://www.jooq.org/javadoc/latest/org.jooq/org/jooq/impl/DSL.html#inline(java.time.Instant)) returns a `Param<Instant>`.
IMO, they should not be treated as the same return type.

PR #1511, opened 2020-02-11
MOE Sync 2020-02-10

cpovirk:

Commits:
=====
<p> Ignore DSL.inline() in PreferJavaTimeOverload checker.

Fixes https://github.com/google/error-prone/issues/1504

2b1d640cf67155b975b3ce2bf2f8b202c03c8d68

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/time/PreferJavaTimeOverload.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/time/PreferJavaTimeOverload.java
@@ -112,6 +112,7 @@ public final class PreferJavaTimeOverload extends BugChecker
 
   private static final Matcher<ExpressionTree> IGNORED_APIS =
       anyOf(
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/google/errorprone/bugpatterns/time/PreferJavaTimeOverload.java
```diff
@@ -112,6 +112,7 @@
 
   private static final Matcher<ExpressionTree> IGNORED_APIS =
       anyOf(
+          staticMethod().onClass("org.jooq.impl.DSL").named("inline"),
```
tbroyer: Out of curiosity, any specific reason you didn't compare return types taking generics into account? (i.e. in this specific case, treating `Param<Long>` and `Param<Instant>` as different return types, such that `DSL.inline(long)` and `DSL.inline(Instant)` would no longer be considered overloads of each other)

PR #1531, opened 2020-03-04
bump auto-service to 1.0-rc6, the newest version

calvinhkf:
bump versions and unify auto-servide across modules to avoid inconsistency


calvinhkf:
@googlebot I signed it! @cushon Signed :-)

Issue #1211, opened 2019-01-08
RFE: configurable DoNotCall/RestrictedApi/JdkObsolete/DefaultCharset/JavaTimeDefaultTimeZone check


tbroyer:
### Description of the problem / feature request:

I'm currently using [Policeman's Forbidden API Checker](https://github.com/policeman-tools/forbidden-apis) in addition to Error Prone. Some checks are duplicate of what Error Prone provides (default charset, default timezone), but Forbidden APIs provides additional checks (e.g. APIs using the default locale, no `System.out`/`System.err` or `printStackTrace()`), and allows one to add new checks through simple text files.

It occurred to me that those checks could be done at compile-time by Error Prone itself.

### Feature requests: what underlying problem are you trying to solve with this feature?

Remove a tool that duplicates (some) functionality of Error Prone and runs later during the build (also contributing to build time); bringing its features to Error Prone itself:
* more checks: default locale, no `System.out`/`System.err`, or `printStackTrace()`
* configurable "forbidden APIs" checks (à la DoNotCall, RestrictedAPI, JdkObsolete, DefaultCharset, or JavaTimeDefaultTimeZone) through simple text files (rather than writing dedicated Error Prone checks). Alternatively, make it dead-easy to build such checks (possibly using protobuf to store the forbidden signatures, à la ApiDiff).

I might try to build those as custom checks (if I find time), but thought you might be interested in adding this right into Error Prone itself.

### What version of Error Prone are you using?

2.3.2

Stephan202:
W.r.t. point 2 (making it configurable): that could then also trivially replace [`modernizer-maven-plugin`](https://github.com/gaul/modernizer-maven-plugin). (I started working on an Error Prone replacement for this plugin a while back, but didn't yet bring it to a production-ready state. If there is wider interest in this I can see about finalizing and open-sourcing it one of these weeks.)

cushon:
> configurable "forbidden APIs" checks (à la DoNotCall, RestrictedAPI, JdkObsolete, DefaultCharset, or JavaTimeDefaultTimeZone) through simple text files (rather than writing dedicated Error Prone checks)

[...]

PR #1534, opened 2020-03-05
Add SystemOut checker

tbroyer:
This checker checks for uses of System.out and System.err as
possible logs used for debugging purpose.

Inspired by [Policeman's Forbidden API Checker](https://github.com/policeman-tools/forbidden-apis)'s `jdk-system-out` checks.

See also #1211

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/SystemOut.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/SystemOut.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2020 The Error Prone Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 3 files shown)


cushon:
Is the idea for this to be disabled by default, but make it available so projects where using `System.{err,out}` is never desirable (servers, libraries, etc.) can opt in to it? That seems like a good approach. I've seen related bugs in code that's deliberately using `System.out` (e.g. CLIs) where print statements from debugging accidentally get committed to the production version. But that seems like a harder problem to find good heuristics for.

tbroyer:
I would personally always enable it, and `@SuppressWarnings("SystemOut")` where it makes sense.
One heuristic that would be useful (to me) would be similar to `SystemExitOutsideMain`: allowing `System.{out,err}` in classes where a `main()` method is present. That wouldn't work for most CLIs though.

The reason I put it in the disabled-by-default list was mostly that `SystemExitOutsideMain` is there too.

JakeWharton:
I DI System.out and System.err from main but otherwise pass around Print* classes or use a proper logging framework.

cushon:
Thanks for the reminder about `SystemExitOutsideMain`, I'm not sure if something similar had been discussed for `System.out` or if I was just half-remembering that check. The heuristic for `System.out` seems trickier.

I think that requiring `@SuppressWarnings("SystemOut")` or DI'ing `System.{out,err}` are both sensible policies for a codebase to adopt, but might be too opinionated to have in the default EP configuration. We've generally been trying to make the defaults reflect things that are probably bugs, not just things that are required e.g. by Google best practices. (We don't always achieve that, but help staying honest is appreciated.)

I think this makes sense to make available as an opt-in / disabled-by-default check.

cushon (approved):

tbroyer:
I rebased the branch, in case it can unblock copybara.

cushon (approved):

PR #1550, opened 2020-03-27
MOE Sync 2020-03-27

nglorioso:

Commits:
=====
<p> Simplify SuggestedFix creation using the static factories.

76852fc149ccc9601c4adaf6a4c5229c4205d2aa

-------

<p> Have Refaster resolve classes against the appropriate modules

Fixes #1239

cb344d8623c0c6d2db770b607162edef06d76105


nglorioso:
CLA:yes from #1548 1239

nglorioso:
Forgot to merge this, replaced by #1551

PR #1551, opened 2020-03-30
MOE Sync 2020-03-30

nglorioso:

Commits:
=====
<p> Simplify SuggestedFix creation using the static factories.

2be87f55cc07ff6f74f6dd5bf86b0f17b25ec949

-------

<p> Have Refaster resolve classes against the appropriate modules

Fixes #1239

526440dbeb0ace53ad3a2c7f397ec939b71719b1

-------

<p> bump auto-service to 1.0-rc6, the newest version

Fixes #1531

6dcd0ce702729df0058391a0314f1440ef614e71


nglorioso:
CLA ok per agreement in #1239

PR #1110, opened 2018-09-05
GuardedByUtils: return empty set instead of null

don-vip:
See https://github.com/google/error-prone/issues/1106#issuecomment-418546231

This fixes an NPE seen with JDK12 (when #1107 is merged):

```
[javac]     public final WaySegment getNearestWaySegment(Point p, Predicate<OsmPrimitive> predicate,
    [javac]                             ^
    [javac]      Please report this at https://github.com/google/error-prone/issues/new and include the following:
    [javac]   
    [javac]      error-prone version: 2.3.2-SNAPSHOT
    [javac]      BugPattern: GuardedBy
    [javac]      Stack Trace:
    [javac]      java.lang.NullPointerException
    [javac]   	at com.google.errorprone.bugpatterns.threadsafety.HeldLockAnalyzer$LockScanner.checkMatch(HeldLockAnalyzer.java:227)
    [javac]   	at com.google.errorprone.bugpatterns.threadsafety.HeldLockAnalyzer$LockScanner.visitIdentifier(HeldLockAnalyzer.java:196)
    [javac]   	at com.google.errorprone.bugpatterns.threadsafety.HeldLockAnalyzer$LockScanner.visitIdentifier(HeldLockAnalyzer.java:116)
    [javac]   	at jdk.compiler/com.sun.tools.javac.tree.JCTree$JCIdent.accept(JCTree.java:2324)
```

Issue #1106, opened 2018-09-02
JDK 14/15 compatibility


don-vip:
### What version of Error Prone are you using?
2.3.4 (originally 2.3.1)

### Does this issue reproduce with the latest release?
Yes

### What did you do?

Error-prone does not work with JDK12 EA b09 and produces the following exception:
```
    [javac] 1 warning
    [javac] An exception has occurred in the compiler (12-ea). Please file a bug against the Java compiler via the Java bug reporting page (http://bugreport.java.com) after checking the Bug Database (http://bugs.java.com) for duplicates. Include your program and the following diagnostic in your report. Thank you.
    [javac] java.lang.NoSuchMethodError: com.sun.tools.javac.util.Log.error(Lcom/sun/tools/javac/util/JCDiagnostic$DiagnosticPosition;Ljava/lang/String;[Ljava/lang/Object;)V
    [javac] 	at com.google.errorprone.ErrorProneError.logFatalError(ErrorProneError.java:55)
    [javac] 	at com.google.errorprone.ErrorProneAnalyzer.finished(ErrorProneAnalyzer.java:155)
    [javac] 	at jdk.compiler/com.sun.tools.javac.api.MultiTaskListener.finished(MultiTaskListener.java:132)
    [javac] 	at jdk.compiler/com.sun.tools.javac.main.JavaCompiler.flow(JavaCompiler.java:1418)
    [javac] 	at jdk.compiler/com.sun.tools.javac.main.JavaCompiler.flow(JavaCompiler.java:1365)
    [javac] 	at jdk.compiler/com.sun.tools.javac.main.JavaCompiler.compile(JavaCompiler.java:960)
    [javac] 	at jdk.compiler/com.sun.tools.javac.main.Main.compile(Main.java:311)
    [javac] 	at jdk.compiler/com.sun.tools.javac.main.Main.compile(Main.java:170)
    [javac] 	at jdk.compiler/com.sun.tools.javac.Main.compile(Main.java:57)
    [javac] 	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    [javac] 	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    [javac] 	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[...]

PR #1571, opened 2020-04-15
Update build matrix for Java 14

TimvdLippe:
This allows us to check that ErrorProne works on Java 12. It is part of
issue #1106 to make ErrorProne work on newer versions of Java.

It will allow us to verify that #1110 indeed fixes the problem with Java
12.


tbroyer:
12 is EOL since last September, 13 for nearly one month already. Shouldn't this use JDK 14 instead?
(oh, and that method is gone in 14: https://docs.oracle.com/en/java/javase/14/docs/api/jdk.compiler/com/sun/source/tree/BreakTree.html)

FWIW google-java-format doesn't even try to be compatible with 13: https://github.com/google/google-java-format/issues/436#issuecomment-613779574

don-vip:
It should always use the latest version of Java. That means at minimum an update every 6 months.

TimvdLippe:
The compilation error I am currently hitting is a method signature changed in JDK 12 but remains in JDK 14. So we will need to fix that also for JDK 14. We are discussing mitigations in an internal thread.

tbroyer:
You may also want to migrate to Xenial or even Bionic rather than Trusty, and openjdk8 rather than oraclejdk8

core/src/main/java/com/google/errorprone/refaster/UInstanceOf.java
```diff
@@ -32,9 +33,11 @@
 @AutoValue
 abstract class UInstanceOf extends UExpression implements InstanceOfTree {
   public static UInstanceOf create(UExpression expression, UTree<?> type) {
-    return new AutoValue_UInstanceOf(expression, type);
+    return new AutoValue_UInstanceOf(null, expression, type);
   }
 
+  public abstract JCPattern getPattern();
```
TimvdLippe, line 39: This method is problematic, as `JCPattern` has only been introduced in JDK 14. It implements the interface `com.sun.source.tree.PatternTree`, which seems like a marker interface: http://hg.openjdk.java.net/jdk/jdk14/file/6c954123ee8d/src/jdk.compiler/share/classes/com/sun/source/tree/PatternTree.java

I am not sure how we can implement this interface on JDK 14 without breaking JDK 8/11.

core/src/main/java/com/google/errorprone/refaster/Inliner.java
```diff
@@ -224,12 +224,9 @@ public TypeVar inlineAsVar(UTypeVar var) throws CouldNotResolveImportException {
     }
     Name name = asName(var.getName());
     TypeSymbol sym = new TypeVariableSymbol(0, name, null, symtab().noSymbol);
-    typeVar = new TypeVar(sym, /* bound= */ null, /* lower= */ symtab().botType);
+    typeVar = new TypeVar(sym, var.getUpperBound().inline(this), var.getLowerBound().inline(this));
     sym.type = typeVar;
     typeVarCache.put(var.getName(), typeVar);
-    // Any recursive uses of var will point to the same TypeVar object generated above.
-    typeVar.bound = var.getUpperBound().inline(this);
```
TimvdLippe, line 231: These fields are removed and are instead put in the constructor. Looking at the code, it seems to make sense to put them in the constructor in the first place, since we would still override them. This should be 8/11-compatible.

core/src/main/java/com/google/errorprone/refaster/ExpressionTemplate.java
```diff
@@ -282,7 +282,7 @@ private static int getPrecedence(JCTree leaf, Context context) {
       return (typeCast.expr == leaf) ? TreeInfo.prefixPrec : TreeInfo.noPrec;
     } else if (parent instanceof JCInstanceOf) {
       JCInstanceOf instanceOf = (JCInstanceOf) parent;
-      return TreeInfo.ordPrec + ((instanceOf.clazz == leaf) ? 1 : 0);
+      return TreeInfo.ordPrec + ((instanceOf.getType() == leaf) ? 1 : 0);
```
TimvdLippe, line 285: Similar as below, the field has been removed and we should use `getType()` (which is a getter for the field).

check_api/src/main/java/com/google/errorprone/util/FindIdentifiers.java
```diff
@@ -98,17 +99,48 @@ public static Symbol findIdent(String name, VisitorState state, KindSelector kin
...
+      Symbol result = (Symbol) FindIdent.method.invoke(
+              Resolve.instance(state.context),
+              FindIdent.args.of(env, state.getName(name), kind));
       return result.exists() ? result : null;
     } catch (ReflectiveOperationException e) {
       throw new LinkageError(e.getMessage(), e);
     }
   }
 
+  private static final class FindIdent {
```
TimvdLippe, line 111: Pulled all of this in from #1439 to fix a lot of test failures.

core/src/main/java/com/google/errorprone/refaster/PlaceholderUnificationVisitor.java
```diff
@@ -665,7 +665,8 @@ public Boolean visitIdentifier(IdentifierTree node, Unifier unifier) {
     return chooseSubtrees(
         state,
         s -> unifyStatements(node.getStatements(), s),
-        stmts -> maker().Case((JCExpression) node.getExpression(), stmts));
+        s -> unify(node.getBody(), s),
+        (stmts, body) -> maker().Case(CaseTree.CaseKind.STATEMENT, List.<JCExpression>of((JCExpression) node.getExpression()), stmts, body));
```
TimvdLippe, line 669: We have to provide 2 additional arguments on JDK 14. It seems like this should be fixed similarly to the `FindIdentifiers` solution (e.g. reflection).

cushon:
CI is now passing for 12 and 13: https://github.com/google/error-prone/commit/937b3d2d704a7609544161cdef4041e705e2f30a. (14+ is still WIP: https://github.com/google/error-prone/issues/1106#issuecomment-628950617)

PR #1586, opened 2020-04-24
Fix NPE in MutablePublicArray check

Preston4tw:
The MutablePublicArray check throws a NPE when examining
an array that's initialized during a static init block. Add a test case
and null checks.

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/MutablePublicArray.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/MutablePublicArray.java
@@ -67,7 +67,7 @@ public class MutablePublicArray extends BugChecker implements VariableTreeMatche
       return false;
     }
     JCNewArray newArray = (JCNewArray) arrayExpression.getInitializer();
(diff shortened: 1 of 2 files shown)


Preston4tw:
I'm a fairly inexperienced Java developer and it's not obvious if there's a better fix for the new test, but the PR as-is passes all the existing test cases in addition to the newly added one.

core/src/main/java/com/google/errorprone/bugpatterns/MutablePublicArray.java
```diff
@@ -67,7 +67,7 @@ private static boolean nonEmptyArrayMatcher(VariableTree arrayExpression, Visito
       return false;
     }
     JCNewArray newArray = (JCNewArray) arrayExpression.getInitializer();
-    if (!newArray.getDimensions().isEmpty()) {
```
cushon, line 70: Thanks for the fix. I suggest just doing the null-check here, instead of having to repeat it below:

```
    if (newArray == null) {
      return false;
    }
```

(I can make that edit when importing the PR, this is just an FYI.)

PR #1587, opened 2020-04-24
Fix Issue #1536

### LINKED ISSUE #1536: Getting a NoClassDefFoundError for "com/github/benmanes/caffeine/cache/Caffeine" when running Error Prone on ant build
> ATTENTION! Please read and follow:
> - if this is a _question_ about Error Prone, send it to error-prone-discuss@googlegroups.com
> - if this is a _bug_ or _feature request_, fill the form below as best as you can.

Getting a NoClassDefFoundError for "com/github/benmanes/caffeine/cache/Caffeine" when running Error Prone on ant build for Apache Nutch v1.1 project.

After changing "target" and "source" to "1.8" (using version 1.8.0_144 for Java compiler), adding dependencies to locations in .m2 repository and updating build.xml file per instructions in https://errorprone.info/docs/installation#ant, I'm getting the following error below:

compile-core:
    [javac] Compiling 236 source files to C:\Users\Philip\Documents\RIT\Graduate\2019-2020\Spring\SWEN.777.01 Quality Assurance\Assignment 3\apache-nutch-1.1\build\classes
    [javac] error: An unhandled exception was thrown by the Error Prone static analysis plugin.
    [javac]      Please report this at https://github.com/google/error-prone/issues/new and include the following:
    [javac]
    [javac]      error-prone version: 2.3.4
    [javac]      BugPattern: (see stack trace)
    [javac]      Stack Trace:
    [javac]      java.lang.NoClassDefFoundError: com/github/benmanes/caffeine/cache/Caffeine
    [javac]     at com.google.errorprone.util.ASTHelpers.<clinit>(ASTHelpers.java:694)
    [javac]     at com.google.errorprone.SuppressionInfo$1.visitClass(SuppressionInfo.java:117)
    [javac]     at com.google.errorprone.SuppressionInfo$1.visitClass(SuppressionInfo.java:114)
    [javac]     at com.sun.tools.javac.tree.JCTree$JCClassDecl.accept(JCTree.java:808)
    [javac]     at com.sun.source.util.SimpleTreeVisitor.visit(SimpleTreeVisitor.java:80)
    [javac]     at com.sun.source.util.SimpleTreeVisitor.visit(SimpleTreeVisitor.java:94)
    [javac]     at com.google.errorprone.SuppressionInfo.forCompilationUnit(SuppressionInfo.java:121)
    [javac]     at com.google.errorprone.scanner.Scanner.updateSuppressions(Scanner.java:87)
    [javac]     at com.google.errorprone.scanner.Scanner.scan(Scanner.java:54)
    [javac]     at com.google.errorprone.scanner.ErrorProneScannerTransformer.apply(ErrorProneScannerTransformer.java:43)
    [javac]     at com.google.errorprone.ErrorProneAnalyzer.finished(ErrorProneAnalyzer.java:152)
    [javac]     at com.sun.tools.javac.api.MultiTaskListener.finished(MultiTaskListener.java:120)
    [javac]     at com.sun.tools.javac.main.JavaCompiler.flow(JavaCompiler.java:1404)
    [javac]     at com.sun.tools.javac.main.JavaCompiler.flow(JavaCompiler.java:1353)
    [javac]     at com.sun.tools.javac.main.JavaCompiler.compile(JavaCompiler.java:946)
    [javac]     at com.sun.tools.javac.main.Main.compile(Main.java:302)
    [javac]     at com.sun.tools.javac.main.Main.compile(Main.java:162)
    [javac]     at com.sun.tools.javac.Main.compile(Main.java:57)
    [javac]     at com.sun.tools.javac.Main.main(Main.java:43)
    [javac]   Caused by: java.lang.ClassNotFoundException: com.github.benmanes.caffeine.cache.Caffeine
    [javac]     at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
    [javac]     at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
    [javac]     at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
    [javac]     ... 19 more
    [javac] Note: Some input files use or override a deprecated API.
    [javac] Note: Recompile with -Xlint:deprecation for details.
    [javac] Note: Some input files use unchecked or unsafe operations.
    [javac] Note: Recompile with -Xlint:unchecked for details.
    [javac] 1 error

BUILD FAILED
C:\Users\Philip\Documents\RIT\Graduate\2019-2020\Spring\SWEN.777.01 Quality Assurance\Assignment 3\apache-nutch-1.1\build.xml:127: Compile failed; see the compiler error output for details.

Total time: 10 seconds

Even after adding the ben-manes artifact (which contains the Caffeine class) to the project's lib folder, I'm still getting the same error. I'm also getting this error when running the ant example for error-prone, but I'm not getting the error when running the maven example (which is working as expected).

2.3.4

No

Preston4tw:
Remove "provided" scope from com.google.auto.value:auto-value.
Scope provided seems to cause build system consumers to expect that
auto-value is present on the classpath somewhere at runtime, while
nothing is providing it.

I tested this locally with gradle and this seems to be the smallest
change that fixes #1536. Gradle properly handles the auto-value
dependency without requiring it to be shaded.


cpovirk:
It might make sense to have a compile/runtime dependency only on `auto-value-annotations` and then to add `auto-value` (which contains the processor) to `annotationProcessorPaths` instead. See https://github.com/google/auto/blob/master/value/userguide/index.md#with-maven

Preston4tw:
Ok, if I understood you correctly, this next commit is in line with what you proposed, and still seems to work in testing error prone against a gradle project through the net.ltgt.errorprone gradle plugin. It's worth noting that I saw this comment, https://github.com/google/error-prone/pull/1398#issuecomment-609976958 , and I've been testing using `-XepPatchChecks:...` successfully without needing to include diffutils-1.3.0.jar in the shaded jar as well.

cpovirk:
Thanks. I was proposing to put `auto-value` under `annotationProcessorPaths`, rather than keeping it in `dependencies` as `provided`. But I suspect that either way is an improvement over the current setup.

Preston4tw:
I tried this diff (and then removing com.google.auto.value:auto-value as a dependency in the various pom.xml files), but this results in some test failures I don't quite understand:

```
index 5108624da..c819933d0 100644
--- a/pom.xml
+++ b/pom.xml
@@ -131,6 +131,13 @@
           <testExcludes>
             <exclude>**/testdata/**</exclude>
           </testExcludes>
+          <annotationProcessorPaths>
+            <path>
+              <groupId>com.google.auto.value</groupId>
+              <artifactId>auto-value</artifactId>
+              <version>${autovalue.version}</version>
+            </path>
+          </annotationProcessorPaths>
         </configuration>
       </plugin>
       <plugin>
```

```
Results :

Failed tests: 
  ErrorProneJavacPluginTest.applyFixes:151 
expected to be true
  ErrorProneJavacPluginTest.applyToPatchFile:210 
expected to be true
  ErrorProneJavacPluginTest.compilesWithFix:383 expected to contain: [TestCompilesWithFix]
but was            : plug-in not found: ErrorProne
  ErrorProneJavacPluginTest.explicitBadPolicyGiven:259 expected java.lang.RuntimeException to be thrown, but nothing was thrown
  ErrorProneJavacPluginTest.hello:110 expected to contain: [CollectionIncompatibleType]
but was            : plug-in not found: ErrorProne
  ErrorProneJavacPluginTest.noPolicyGiven:236 expected java.lang.RuntimeException to be thrown, but nothing was thrown
  ErrorProneJavacPluginTest.stopOnErrorPolicy:313 value of    : iterable.size()
expected    : 2
but was     : 1
iterable was: [plug-in not found: ErrorProne]

Tests run: 3612, Failures: 7, Errors: 0, Skipped: 49
```

cpovirk:
Sorry about that. I hit a similar problem with another project, so I think I know what's up.

As soon as you add the first entry to `annotationProcessorPaths`, Maven/javac stops looking for _other_ annotation processors (and plugins, like Error Prone) in your _regular_ classpath.

This is probably the more sensible behavior. However, if a project currently depends on finding annotation processors or plugins in the regular classpath, it produces a breakage.

The fix is to change the setup for Error Prone itself to list it in `annotationProcessorPaths`, as in [the instructions](https://errorprone.info/docs/installation#maven).

However, I have _no idea_ what additional challenges that might present, so I completely understand if you want to stick with `provided`. Thanks for trying `annotationProcessorPaths` out.

(You can see where I [unknowingly introduced](https://github.com/google/jimfs/commit/a8489ffa85dfab937dcc9e8d59904cefc4331f31) and [unknowingly fixed](https://github.com/google/jimfs/commit/e8a24aa250bd49d1744753929e49e1cd10f816cf) the problem in another project.)

Preston4tw:
My understanding so far is that adding the annotationProcessorPaths section to pom.xml adds a `-processorpath` arg to javac. Without `-processorpath` specified, annotators are searched for in the classpath.

The test failures are because the tests were relying on being able to find the locally compiled ErrorProne in the classpath, ex. https://github.com/google/error-prone/blob/master/core/src/test/java/com/google/errorprone/ErrorProneJavacPluginTest.java#L302.

As I understand it, the fix here would be to add the target/ directory to `processorpath` or modify the tests somehow to do that themselves. We can't just add ErrorProne to the top pom.xml as an annotationProcessor because that goes through the normal maven external resolution process (if you set the project pom versions to 2.3.6-SNAPSHOT maven quickly fails on resolution for this reason). I can't figure out how to add target/ to processorpath and I'm also unsure how the tests could be modified to do that.

If anyone has any ideas on how to accomplish the above I'd be happy to try them and modify the PR accordingly if any of the ideas pan out, otherwise I think the PR is good as-is.

As a side note I noticed I had committed the diffutils change and I can either leave that in the PR as-is or remove that if you have any comments on that one way or the other. My testing against a gradle project using JDK 11 and using many `-XepPatchChecks:` arguments succeed without this dependency being included in the shaded jar.

tbroyer:
I can't understand why changing how the tests are compiled (looking up annotation processors from `annotationProcessorPaths` rather than the classpath) would change how they behave at runtime. There must be something else that breaks the tests.
I'll have a look if I find time this weekend.

tbroyer:
Ok, understood: same change needs to be done with auto-service: use `auto-service-annotations` as dependency and move `auto-service` to `annotationProcessorPaths`.
The `META-INF/services/com.sun.source.util.Plugin` file no longer is generated, so indeed JavaC can no longer find the plugin when using `-Xplugin:ErrorProne`.
As @cushon said, as soon as you start using `-processorpath` (`annotationProcessorPaths` in Maven), you need to move all processors to it (and ideally out of the classpath); but the problem wasn't ErrorProne itself (which is not used when building ErrorProne), but auto-service.

BTW, should auto-service-annotations and auto-value-annotations be shaded into the `with-dependencies` JAR? They're not required, but if we add them as transitive dependencies to those consuming ErrorProne as a Maven dependency, then maybe we should add them to the `with-dependencies` JAR too, for consistency.

Preston4tw:
Thanks for taking a look and explaining the issue @tbroyer. I've tested moving auto-value and auto-service into annotationProcessorPaths and it works in my local testing and passes the unit tests. I've also modified the shade config to include the auto-value and auto-service annotations.

docgen/pom.xml
```diff
@@ -35,6 +35,21 @@
   </licenses>
 
   <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
+          <annotationProcessorPaths>
```
tbroyer, line 43: Is this missing auto-service? or is the dependency on auto-service-annotations actually a mistake?

docgen_processor/pom.xml
```diff
@@ -47,15 +47,13 @@
     </dependency>
     <dependency>
       <groupId>com.google.auto.service</groupId>
-      <artifactId>auto-service</artifactId>
-      <version>1.0-rc6</version>
-      <optional>true</optional>
+      <artifactId>auto-service-annotations</artifactId>
```
tbroyer, line 50: This submodule is now missing annotation processing

examples/plugin/maven/sample_plugin/pom.xml
```diff
@@ -42,9 +42,8 @@
     </dependency>
     <dependency>
       <groupId>com.google.auto.service</groupId>
-      <artifactId>auto-service</artifactId>
+      <artifactId>auto-service-annotations</artifactId>
```
tbroyer, line 45: Same here: build no longer processes AutoService annotation

core/pom.xml
```diff
@@ -362,6 +362,24 @@
       </extension>
     </extensions>
     <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
+          <annotationProcessorPaths>
```
tbroyer, line 369: I just noticed a `run-annotation-processor` profile that adds `error_prone_docgen_processor` to the dependencies, and would need to be changed to add to the `annotationProcessorPaths` instead (using `combine.children="append"`, see https://maven.apache.org/pom.html#plugins)

----

(docgen_processor is also added as a dependency to docgen, but AFAICT this is for its `BugPatternInstance` class, not for its annotation processor, so no need to add it to the processor path there)

docgen/pom.xml
Preston4tw: I think the inclusion of auto-service for docgen is incorrect. I'm not seeing any references to auto-service in any of the sources. The dependency should probably just be removed.

docgen_processor/pom.xml
Preston4tw: I'm confused. There's the annotation processing for compiling error prone, which seems to be ok because if there were an issue the compile should be failing with an error on a missing symbol for the annotation right?

Are you referring to a downstream effect due to how this piece fits into error prone's design, such that error prone plugin users will have issues with annotations? Like is there a way I can produce a failure somehow to understand the consequence or how this is incorrect?

tbroyer: docgen submodules are used to generate the documentation for the website.
IIUC, they're used when building while enabling the `run-annotation-processor` profile: docgen_processor is then applied to core to generate a txt file, and docgen is executed with that txt file as input to generate the markdown files.
Without the auto-service processor here, the txt file won't be generated.

Preston4tw: This comment was super helpful. I found util/generate-latest-docs.sh and was able to reproduce the issue, and this latest push should fix this and the other comments raised.

core/pom.xml
Preston4tw: As a first pass I ended up just declaring the whole block. I worry that the `combine.children` might later confuse people down the line. If you have a preference for the `combine.children` approach I'm happy to make the change.

Preston4tw:
Anything left to do on this PR before merge?

PR #1592, opened 2020-04-26
RestrictedApi support for Type

yayaa:
Adding support for restricting API for the entire type. RestrictedApi check currently only supports for annotating methods, with this change, it will be possible to annotate the class and make sure all the methods in that class will be restricted with given annotation configuration.

```java
@RestrictedApi(explanation = "type restriction", link = "nothing")
interface EntireClassRestriction {
  void cantTouchThis();
}
```

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/RestrictedApiChecker.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/RestrictedApiChecker.java
@@ -99,19 +99,29 @@ public class RestrictedApiChecker extends BugChecker
   }
 
   private Description checkMethodUse(ExpressionTree tree, VisitorState state) {
-    RestrictedApi annotation = ASTHelpers.getAnnotation(tree, RestrictedApi.class);
+    Symbol symbol = ASTHelpers.getSymbol(tree);
+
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/google/errorprone/bugpatterns/RestrictedApiChecker.java
```diff
@@ -99,19 +99,29 @@ public Description matchNewClass(NewClassTree tree, VisitorState state) {
   }
 
   private Description checkMethodUse(ExpressionTree tree, VisitorState state) {
```
cushon, line 101: I think this is only going to check `@RestrictedApi` on types if a method is called on those types. If we add support for this I think we'd want to disallow all uses of `@RestrictedApi`-annotated types, e.g. if they appear as qualifiers for fields, or as class literals, etc. instead of just when they appear as the qualifier of a method invocation.

cushon:
Thanks for the suggestion. After some more discussion I think we might do something in this area, but there are open questions about what behaviour annotating a type with `@RestrictedApi` should have, and we'd want to answer those questions before proceeding with an implementation. I filed https://github.com/google/error-prone/issues/2087 to track this feature request.

PR #1608, opened 2020-05-10
Avoid `NoClassDefFoundError` in `AutoValueSubclassLeaked`

Stephan202:
This change is relevant for Error Prone configurations that are shared between multiple modules, some of which don't have Auto Value on their classpath.


cushon:
I think this will be fixed by https://github.com/google/error-prone/pull/1587 without needing to use the string name instead of the class literal.

PR #1609, opened 2020-05-10
Prevent "<no match>" reports by `StronglyTypeTime`

Stephan202:
This fix prevents reports such as the following, which are hard to debug for users:
```
[INFO] /path/to/some/File.java: [<no match>] <no match>
    (see <no match>)
```

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/time/StronglyTypeTime.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/time/StronglyTypeTime.java
@@ -133,7 +133,10 @@ public final class StronglyTypeTime extends BugChecker implements CompilationUni
     }.scan(tree, null);
 
     for (Map.Entry<VarSymbol, TreePath> entry : fields.entrySet()) {
-      state.reportMatch(describeMatch(entry.getValue(), usages.get(entry.getKey()), state));
+      Description description = describeMatch(entry.getValue(), usages.get(entry.getKey()), state);
+      if (description != NO_MATCH) {
+        state.reportMatch(description);
+      }
     }
     return NO_MATCH;
   }
--- a/core/src/test/java/com/google/errorprone/bugpatterns/time/StronglyTypeTimeTest.java
+++ b/core/src/test/java/com/google/errorprone/bugpatterns/time/StronglyTypeTimeTest.java
@@ -246,6 +246,19 @@ public final class StronglyTypeTimeTest {
   }
 
   @Test
+  public void unusedField_noMatch() {
+    compilationHelper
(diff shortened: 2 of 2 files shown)


core/src/main/java/com/google/errorprone/bugpatterns/time/StronglyTypeTime.java
```diff
@@ -133,7 +133,10 @@ private void handle(Tree tree) {
     }.scan(tree, null);
 
     for (Map.Entry<VarSymbol, TreePath> entry : fields.entrySet()) {
-      state.reportMatch(describeMatch(entry.getValue(), usages.get(entry.getKey()), state));
```
cushon, line 136: Thanks for the fix. I think this is a more general problem with the handling of `NO_MATCH` that affects other checks, and that we should handle it here:

https://github.com/google/error-prone/blob/3006a975deef491be769ec8c8e257a5160afbb77/check_api/src/main/java/com/google/errorprone/VisitorState.java#L266

instead of here:

https://github.com/google/error-prone/blob/3006a975deef491be769ec8c8e257a5160afbb77/check_api/src/main/java/com/google/errorprone/scanner/Scanner.java#L124

Any concerns with that change? (I can make the edit when importing, I just want to check if you see any reasons not do that.) Thanks!

Stephan202: I think that's a good suggestion, and indeed I wondered about this. But then I noticed that several existing plugins follow the current pattern, so I assumed that there was a reason for not handling `NO_MATCH` at such a "deep" level.

But now that I think about it a bit more, the suggested change makes a lot of sense: `NO_MATCH` is owned by `Description`, so in a sense `NO_MATCH` is to `Description` what `empty` is to `Optional`. Since `VisitorState` explicitly handles `Description` it is thus reasonable that it also handles `NO_MATCH` .

For reference, if you make that change then the following callers can also be simplified:

https://github.com/google/error-prone/blob/704e8a1593e0017d53d5fcd9daabcdcf64c10ab6/core/src/main/java/com/google/errorprone/bugpatterns/ExpectedExceptionChecker.java#L112-L115

https://github.com/google/error-prone/blob/3c0c609c5334f4fecf37830846c8c08d1a503e43/core/src/main/java/com/google/errorprone/bugpatterns/threadsafety/GuardedByChecker.java#L206-L209

https://github.com/google/error-prone/blob/a91042e86aef0d93f5772824076398e0c2da7c29/core/src/main/java/com/google/errorprone/bugpatterns/MockitoCast.java#L180-L183

PR #1614, opened 2020-05-14
MOE Sync 2020-05-14

kluever:

Commits:
=====
<p> fix typo

Fixes #1590

4386471667379840c5f4b50bb68897b805d54131

-------

<p> skip unnecessary boxing

Fixes this warning:

Fixes #1558

90197b0344b29f2f879b41def7c55723bb4f4951

-------

<p> Minor fix to FallThrough to skip generated/mutated AST nodes
(e.g. Lombok)

Fixes #1573

1534bae6b3422be39ac82a65068e3c0bb151e859

-------

<p> Make summary of InjectOnConstructorOfAbstractClass consistent

The first sentence says @Injected and the second one says @Inject'ed.

Change the first one to @Inject'ed so it's consistent.

Change-Id: Id2ab8908a608288739af33cddc902f94707b057a

Fixes #1491

08397375c07ccea59a679ea4a8901413abde020f

-------

<p> Drop `@Nullable` from `RefasterRule#afterTemplates()`

This property is never null and the caller unconditionally dereferences

Fixes #1606

87d4fe5fbe0e001fc3cb79d7aedf6bf5438234c6

-------

<p> Handle `NO_MATCH` in `VisitorState#reportMatch` instead of `Scanner

This fix prevents reports such as the following, which are hard to debug
for users:
```
[INFO] /path/to/some/File.java: [<no match>] <no match>
   (see <no match>)

Fixes #1609

5fdc31e61c2975ac93bbacbba5416cdefa1b9e88

-------

<p> Update junit dependency to 4.13

Fixes #1591

df17a5141f44146deb91a4718aa123b52ffcea25

-------

<p> Inline a helper method

Closes https://github.com/google/error-prone/pull/1325

2946131ec99327a3c7175d5ea282833512f8e94e

-------

<p> Add IntelliJ IDEA ImportOrganizer

Many projects use the default import ordering provided by IntelliJ IDEA.
In order to apply Error Prone without adjusting this ordering, one can

Fixes #1208

deea2cf2d72ec38893106536f8f414c5a579d939

-------

<p> Prevent accidental re-use of CompilationTestHelper and BugCheckerRefactoringTestHelper instances

See e.g. https://github.com/google/error-prone/pull/1313

2c5708c357f9d4f6b4b2a61588fe2f162bcc5b71

-------

<p> Don't handle nulls in VisitorState#reportMatch.

37eeca46f6078a09ec1e0db0b75c0ba95483bbae

-------

<p> Breaks mockito (d'oh)

882fc471b12b67b83588594fecc3fa4b300e8641

### DIFF
--- a/check_api/src/main/java/com/google/errorprone/ImportOrderParser.java
+++ b/check_api/src/main/java/com/google/errorprone/ImportOrderParser.java
@@ -36,6 +36,8 @@ public class ImportOrderParser {
         return ImportOrganizer.ANDROID_STATIC_FIRST_ORGANIZER;
       case "android-static-last":
         return ImportOrganizer.ANDROID_STATIC_LAST_ORGANIZER;
+      case "idea":
+        return ImportOrganizer.IDEA_ORGANIZER;
       default:
         throw new IllegalStateException("Unknown import order: '" + importOrder + "'");
     }
--- a/check_api/src/main/java/com/google/errorprone/VisitorState.java
+++ b/check_api/src/main/java/com/google/errorprone/VisitorState.java
@@ -264,7 +264,7 @@ public class VisitorState {
   }
 
   public void reportMatch(Description description) {
-    if (description == null || description == Description.NO_MATCH) {
+    if (description == Description.NO_MATCH) {
       return;
     }
     // TODO(cushon): creating Descriptions with the default severity and updating them here isn't
--- a/check_api/src/main/java/com/google/errorprone/VisitorState.java
+++ b/check_api/src/main/java/com/google/errorprone/VisitorState.java
@@ -264,7 +264,7 @@ public class VisitorState {
   }
 
   public void reportMatch(Description description) {
-    if (description == Description.NO_MATCH) {
+    if (description == null || description == Description.NO_MATCH) {
       return;
     }
     // TODO(cushon): creating Descriptions with the default severity and updating them here isn't
--- a/check_api/src/main/java/com/google/errorprone/VisitorState.java
+++ b/check_api/src/main/java/com/google/errorprone/VisitorState.java
@@ -264,6 +264,9 @@ public class VisitorState {
   }
 
   public void reportMatch(Description description) {
+    if (description == null || description == Description.NO_MATCH) {
+      return;
+    }
     // TODO(cushon): creating Descriptions with the default severity and updating them here isn't
     // ideal (we could forget to do the update), so consider removing severity from Description.
     // Instead, there could be another method on the listener that took a description and a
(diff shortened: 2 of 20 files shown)


Stephan202:
Could it be that there's a bug in MOE (?) which prevents the last line of the original PR from being mirrored out? Seems to apply to all of my most recent PRs (didn't check further back):
- #1551: b3ee1690754d99319d89210e2aa5fc3815c553d8 vs. 526440dbeb0ace53ad3a2c7f397ec939b71719b1 ("but fail on JDK 9+."l)
- #1473: 80e0b1699fd01522dbe992577532d351fd65d5b6 vs. cd83ca9094ff4e9842c32d0c6d33424e479fa017 (second URL)
- #1609: 0f000f69a45a45c338f2fb6949938480bdbedab1 vs. 5fdc31e61c2975ac93bbacbba5416cdefa1b9e88 (end of markdown block)
- #1614: ef88a48ef499a57601e5891150a7d70c2f8fad75 vs. 87d4fe5fbe0e001fc3cb79d7aedf6bf5438234c6 ("it.")

Stephan202:
(Note that is also the cause of the malformed PR summary: because the monospace markdown block isn't closed it encompasses the description of all commits after it.)

Stephan202:
Just a last example of a PR not by me:
* #1208: a8c2d436cb166a7e222cda1184ec5404aa4617a4 vs. deea2cf2d72ec38893106536f8f414c5a579d939 ("now set `-XepPatchImportOrder:idea`.")

kluever:
Interesting @Stephan202 - hmmm, any ideas @cpovirk  ?

cpovirk:
...huh. Sorry, and thanks for letting me know. I had even seen that at one point but failed to look into it.

For some inexplicable reason, the script I hacked up runs part of the patch through `head -n -2` instead of `head -n -1`, stripping off not just the `---` separator but also the line before. I will fix it.

Stephan202:
Tnx! (Are you perhaps a `vim` user? I find myself sometimes introducing bugs like these after accidentally pressing [`ctrl` + `a`](https://coderwall.com/p/zlz1cw/vim-increment-decrement-a-number-under-cursor), only noticing the issue when reviewing a git diff.)

cpovirk:
I am, and I have definitely hit ctrl+a once or twice in my time :) I also enjoy creating files named `1` by typing `w1` instead of `w!`.

I suspect that what happened here is that I mentally mixed together `tail -n +2`, the way of removing 1 line from the beginning, with `head -n -1`, the way of removing 1 line from the end.

Stephan202:
Ah yes, I always need to do a "dry run" of those, cause I can never seem remember exactly how the parameter to `-n` behaves. Mystery solved :smile:

Issue #1619, opened 2020-05-17
Demote severity of ExtendsAutoValue and RefersToDaggerCodegen to a warning to not break Java cross compilation in Bazel


davido:
Both checks appear to be a breaking changes for Bazel EP integration.

It broke all Bazel users who rely on cross compilation: [1], [2], and was the reason, why the first attempt to bump EP version to 2.3.3 in Bazel was reverted: [3], [4].

The way Bazel cross compilation works is a bit convoluted, but the reason for false positive boils down to: the missing annotation generation is using -source 8 -target 8 with a JDK 11 `-bootclasspath` (because it's targeting the remote JDK 11). So AutoValue processor is trying to use `javax.annotation.Generated` based on the source level, but it isn't available because the JDK 11 bootclasspath only contains `javax.annotation.processing.Generated`.

Demote the severity to warning for the checks that rely on having 
JDK 8's: `javax.annotation.Generated` or JDK 11's `javax.annotation.processing.Generated` on the classpath today, and consider to promote it to error severity again later, when Java 8 is gone.

That would substantially simplify the adoption of new EP releases, because it wouldn't require the addition of `javax.annotation` library to the build tool chain.

[1] https://github.com/bazelbuild/rules_closure/issues/483
[2] https://github.com/bazelbuild/rules_kotlin/issues/321
[3] https://github.com/bazelbuild/bazel/commit/d990746
[4] https://github.com/bazelbuild/bazel/issues/10023

cushon:
#1348

PR #1620, opened 2020-05-17
Demote severity of ExtendsAutoValue and RefersToDaggerCodegen to a wa…

davido:
…rning

Fixes: #1619.

Both checks appear to be a breaking changes for Bazel EP integration.

The way Bazel cross compilation works is a bit convoluted, but the
reason for false positive boils down to: the missing annotation
generation is using -source 8 -target 8 with a JDK 11 -bootclasspath
(because it's targeting the remote JDK 11). So AutoValue processor is
trying to use javax.annotation.Generated based on the source level, but
it isn't available because the JDK 11 bootclasspath only contains
javax.annotation.processing.Generated.

Demote the severity to warning for the checks that rely on having JDK
8's: javax.annotation.Generated or JDK 11's
javax.annotation.processing.Generated on the classpath today, and
consider to promote it to error again later, when Java 8 is gone.

That would substantially simplify the adoption of new EP releases,
because it wouldn't require the addition of javax.annotation library
to the build tool chain.


cushon:
https://github.com/google/error-prone/commit/230ad26ec1d308f6387e7549612d2b4baf244cf2

PR #1621, opened 2020-05-17
Document -XepDisableAllWarnings

msridhar:
Now that #785 has landed, briefly document the existence of the `-XepDisableAllWarnings` flag.  Also describe briefly how the blanket flags can be combined with flags to enable/disable individual checks (slightly redundant but I think users will find it helpful).


msridhar:
@googlebot I signed it!

PR #1626, opened 2020-05-19
Enable JDK 12 and 13 on travis

cushon:
https://github.com/google/error-prone/issues/1106


gliptak:
@cushon will there be a near term ```error-prone``` release with ```12/13``` support?

https://github.com/GoogleContainerTools/jib/pull/2351

PR #776, opened 2017-10-14
Replace public constructor on an abstract class with protected

bangarharshit:
Fixes - https://github.com/google/error-prone/issues/134. Added a suggestion to modify the modifier of constructor in an abstract class to protected. The constructor in abstract class doesn't need to be public since it can't be instantiated and it leads to better code semantics.

PR #1628, opened 2020-05-20
MOE Sync 2020-05-20

cgdecker:

Commits:
=====
<p> Demote a couple of checks that rely on @Generated annotations to warnings

to avoid break builds that don't have an `@Generated` annotation
available, see discussion in:

https://github.com/google/error-prone/issues/1348

0564272889c0acd265de7b78a7805c9bc7a05c74

-------

<p> Fix double objection typo

Fixes #1624

81143259ef7f80ce4337795aa35645c2a989f6bb

-------

<p> Turn down MockitoCastTest

This was a temporary check to support upgrading to the JDK 9 javac.

The upstream mockito bug is fixed:
https://github.com/mockito/mockito/issues/357

Closes https://github.com/google/error-prone/pull/458

5fef0ab8f69c61130372f2c0cf617996cbe304de

-------

<p> Add FieldCanBeStatic checker

This finds final fields which are initialized with a literal.  Adding
static reduces memory overhead from per-instance to per-class and
allows initialization at compile-time instead of run-time.

Fixes #930

46f2ddd3ca9cea1e74d81480f32298052b9e7d8f

-------

<p> Replace public constructor on an abstract class with
protected

Fixes #776

83f68dc4707e18823fa4724a4c15790fc0c7a76b

-------

<p> Discourage calling .equals on java.util.Date

Closes https://github.com/google/error-prone/pull/769

4dfe1e40996af32d761932d0091fa7b03c39e6f8

-------

<p> Discourage empty catch blocks

https://github.com/google/error-prone/pull/372

f067128733b3b60a2339d7ffe0688f2e7a6819d6

-------

<p> Scanner: reject null `Description`s.

fef9c87a6ae2051a8e0f1b7f8cc4007cd25e4a16


cgdecker:
Commits by other authors come from PRs by those authors.

PR #1656, opened 2020-06-01
extend CompareToZero doc

vorburger:
Thank You Error Prone for not just helping a project "conform to contract", but actually avoid real bugs we've found when enforcing this bug pattern... it seems worth to me spell this out more clearly in this doc.

see https://github.com/apache/fineract/pull/951 for https://issues.apache.org/jira/browse/FINERACT-1014

@xurror @thesmallstar @ptuomola FYI

### DIFF
--- a/docs/bugpattern/CompareToZero.md
+++ b/docs/bugpattern/CompareToZero.md
@@ -32,3 +32,17 @@ code if turned into a comparison to `0`, e.g.:
     return comparator.compare(a, b) > 0;
   }
 ```
+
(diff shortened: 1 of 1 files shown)


vorburger:
@percyashu FYI & great job for having found the related bug in our code using Error Prone - Thank You for your participation in Google Summer of Code!

eaftan:
Thank you for the kind note!  I'm sure the author of this check, @graememorgan, would be excited to hear this. 😄 

Our source of truth is the Google-internal repo, so we'll review this internally and then push it back out.

eaftan:
We discussed this internally and decided to stick with the current docs.  But thank you for the positive feedback!

PR #1683, opened 2020-06-14
StringSplitter: also report Pattern#split(String)

simon04:


cushon:
Thanks! Merged with a few modifications as https://github.com/google/error-prone/commit/ea7b968f79229d30bea6546f20644f3641ac0e1d

PR #1685, opened 2020-06-15
Automatic code cleanup.

eaftan:
PiperOrigin-RevId: 316207121


eaftan:
The missing commit author is "Error Prone Team," which is just me running a script.  

@googlebot I fixed it.

PR #1771, opened 2020-08-05
Update to use dataflow framework 3.6.0.

xingweitian:
Adapt to the API change in version 3.5.0 and use the latest dataflow framework version 3.6.0:

The Dataflow Framework supports backward analysis.

Dataflow Framework: Analysis is now an interface.  Added AbstractAnalysis,
ForwardAnalysis, ForwardTransferFunction, ForwardAnalysisImpl,
BackwardAnalysis, BackwardTransferFunction, and BackwardAnalysisImpl.
To adapt existing code:
 * `extends Analysis<V, S, T>` => `extends ForwardAnalysisImpl<V, S, T>`
 * `implements TransferFunction<V, S>` => `implements ForwardTransferFunction<V, S>`

In addition, the method signature of `CFGLambda` changed from
`public CFGLambda(LambdaExpressionTree lambda)` to
`public CFGLambda(LambdaExpressionTree lambda, ClassTree classTree, MethodTree method)`

### DIFF
--- a/check_api/src/main/java/com/google/errorprone/dataflow/DataFlow.java
+++ b/check_api/src/main/java/com/google/errorprone/dataflow/DataFlow.java
@@ -38,11 +38,14 @@ import javax.annotation.Nullable;
 import javax.annotation.processing.ProcessingEnvironment;
 import org.checkerframework.shaded.dataflow.analysis.AbstractValue;
 import org.checkerframework.shaded.dataflow.analysis.Analysis;
+import org.checkerframework.shaded.dataflow.analysis.ForwardAnalysisImpl;
+import org.checkerframework.shaded.dataflow.analysis.ForwardTransferFunction;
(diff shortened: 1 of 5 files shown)


eaftan:
An update: we are still using version 3.4.0 of the Checker Framework inside Google, so this conflicts with the version of the dataflow library we have in our monorepo.  We'll have to update our internal version of the Checker Framework before this can be committed.

xingweitian:
> An update: we are still using version 3.4.0 of the Checker Framework inside Google, so this conflicts with the version of the dataflow library we have in our monorepo. We'll have to update our internal version of the Checker Framework before this can be committed.

Thanks for letting me know! Sure, no problem.

cushon:
https://github.com/google/error-prone/commit/971e945f79593a4e1053fec2636c8a0eda09b5d8

PR #1791, opened 2020-08-23
PreferJavaTimeOverload: don't flag methods on Reactor's Flux type

Stephan202:
For example, one can either a buffer up to a [certain number of values](https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Flux.html#buffer-int-) or for a [certain amount of time](https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Flux.html#buffer-java.time.Duration-); the latter operation is not always a suitable alternative to the former. Similar arguments apply to the cache, replay, skip, take, and window operations.

I'm not sure whether [Reactor](https://projectreactor.io/) is used inside Google, but I hope that this library is deemed sufficiently popular to meet the threshold for inclusion.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -332,6 +332,13 @@
     </dependency>
     <dependency>
       <!-- Apache 2.0 -->
+      <groupId>io.projectreactor</groupId>
+      <artifactId>reactor-core</artifactId>
+      <version>3.3.9.RELEASE</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <!-- Apache 2.0 -->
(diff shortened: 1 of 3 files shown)


Stephan202:
Force-pushed and updated the commit message.

kluever:
Thanks - I've got a CL out to add this internally.

Stephan202:
Tnx for the quick response! Just for my curiosity: the tests were left out to avoid accruing too large a set of odd test dependencies over time?

kluever:
> to avoid accruing too large a set of odd test dependencies

Indeed :-)

PR #1851, opened 2020-09-30
Updated ReadMe

akshit-dev101:


eaftan:
Thanks for the contribution, but I think the text in the README is clear enough as it currently is.

PR #2009, opened 2020-12-13
Handle `SuggestedFixes#qualifyStaticImport` method declaration clashes

Stephan202:
If a class declares a method `m` then statically importing another method `m` does not have the desired effect.

### DIFF
--- a/core/src/test/java/com/google/errorprone/fixes/SuggestedFixesTest.java
+++ b/core/src/test/java/com/google/errorprone/fixes/SuggestedFixesTest.java
@@ -846,14 +846,47 @@ public class SuggestedFixesTest {
   }
 
   @Test
-  public void qualifyStaticImport_whenNamesClash_usesQualifiedName() {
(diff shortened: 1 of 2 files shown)


core/src/test/java/com/google/errorprone/fixes/SuggestedFixesTest.java
```diff
@@ -846,27 +846,59 @@ public void qualifyStaticImport_whenAlreadyImported_usesUnqualifiedName() {
...
             "  void test() {",
             "    verifyNotNull(2);",
             "    Verify.verifyNotNull(1);",
             "  }",
+            "}")
+        .doTest();
+  }
+
+  @Test
+  public void qualifyStaticImport_whenMethodNamesClash_usesQualifiedName() {
```
Stephan202, line 883: Prior to these changes this test would fail, because `verifyNotNull` is declared but not referenced.

cushon (approved):

Issue #1415, opened 2019-11-18
error-prone release schedule


carterkozak:
### Description of the problem / feature request:

Is there an error-prone release schedule? The documentation at https://errorprone.info/bugpatterns references patterns from master that do not exist in any published release.

In particular I'm keen to try `InterruptedExceptionSwallowed`.

### Feature requests: what underlying problem are you trying to solve with this feature?

I'd really appreciate a structured release schedule, perhaps once a month?

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Open https://github.com/google/error-prone/releases or check maven central.

### What version of Error Prone are you using?

2.3.3.

###  Have you found anything relevant by searching the web?

No, neither https://groups.google.com/forum/#!forum/error-prone-discuss nor this project document a release schedule or roadmap.

Thanks for your help!

floscher:
On the same day that you opened this issue, a commit was [tagged as `2.3.4`](https://github.com/google/error-prone/releases/tag/2.3.4) (note the missing leading `v`, hopefully this doesn't break anything in the [release process](https://github.com/google/error-prone/wiki/Releasing#releasing)).
So the next release seems to be right around the corner. Unfortunately I have no idea how long it will take for the maintainers to release. For `2.3.3` they needed almost one month from pushing the git-tag to publishing the release, for `2.3.2` they needed less than 48 hours.

cushon:
Sorry for not responding here sooner. (There have at least been releases since this was filed, so it it should be possible to try out `InterruptedExceptionSwallowed` :)

We had been trying to do releases roughly every two months, but we obviouslybviously haven't been keeping to that schedule. I'm not sure we'll be able to promise a monthly cadence, but I am going to try to make more releases happen this year than last year.

PR #2026, opened 2020-12-18
MissingCasesInEnumSwitch handles case labels with multiple expressions

pkoenig10:
`MissingCasesInEnumSwitch` does not correctly handle multi-expression case labels in switch expressions. It uses the deprecated `CaseTree#getExpression` method, which only returns the first expression in the case label.

A simple example is shown below:
```
enum Case { ONE, TWO }
void m(Case c) {
  switch (c) {
    case ONE, TWO -> {}
  }
}
```

This code will produce the following error:
```
[MissingCasesInEnumSwitch] Non-exhaustive switch; either add a default or handle the remaining cases: TWO
```

This PR updates `MissingCasesInEnumSwitch` to instead use the non-deprecated `CaseTree#getExpressions` method, which returns all expressions in the case label.

Unfortunately this requires using reflection because the `CaseTree#getExpressions` method was not added until Java 12.

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/MissingCasesInEnumSwitch.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/MissingCasesInEnumSwitch.java
@@ -26,12 +26,17 @@ import com.google.errorprone.VisitorState;
 import com.google.errorprone.bugpatterns.BugChecker.SwitchTreeMatcher;
 import com.google.errorprone.matchers.Description;
 import com.google.errorprone.util.ASTHelpers;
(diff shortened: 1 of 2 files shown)


cushon (approved):

pkoenig10:
@cushon Any plans to cut a new release soon? I'm eager to pick up this fix.

See also https://github.com/google/error-prone/issues/1415.

cushon:
@pkoenig10 I'm planning to do one soon (like in the next week or so): https://github.com/google/error-prone/issues/2061

PR #2232, opened 2021-03-11
Introduce `MissingRefasterAnnotation` checker

rickie:
See the `MissingRefasterAnnotation.md` for more details.

Feedback is welcome!

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/MissingRefasterAnnotation.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/MissingRefasterAnnotation.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2021 The Error Prone Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.errorprone.bugpatterns;
+
+import static com.google.errorprone.matchers.ChildMultiMatcher.MatchType.AT_LEAST_ONE;
+import static com.google.errorprone.matchers.Matchers.annotations;
+import static com.google.errorprone.matchers.Matchers.anyOf;
+import static com.google.errorprone.matchers.Matchers.isType;
+
+import com.google.auto.service.AutoService;
+import com.google.errorprone.BugPattern;
+import com.google.errorprone.BugPattern.LinkType;
+import com.google.errorprone.BugPattern.SeverityLevel;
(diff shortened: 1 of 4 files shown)


Stephan202 (approved):

core/src/main/java/com/google/errorprone/bugpatterns/MissingRefasterAnnotation.java
```diff
@@ -0,0 +1,67 @@
...
+    long methodTypes =
+        tree.getMembers().stream()
+            .filter(member -> member.getKind() == Tree.Kind.METHOD)
+            .map(MethodTree.class::cast)
+            .filter(method -> !ASTHelpers.isGeneratedConstructor(method))
+            .map(method -> HAS_REFASTER_ANNOTATION.matches(method, state))
+            .distinct()
+            .count();
+
+    return methodTypes < 2 ? Description.NO_MATCH : buildDescription(tree).build();
```
cushon, line 65: Why only match if there are at least two methods with refaster annotations? I know we'd usually expect to see at least two (e.g. both a `@BeforeTemplate` and `@AfterTemplate`) but would only matching on one result in false positives?

Maybe there are cases where there's an abstract class that's used to share a placeholder or something?

core/src/main/java/com/google/errorprone/bugpatterns/MissingRefasterAnnotation.java
```diff
@@ -0,0 +1,67 @@
...
+      annotations(
+          AT_LEAST_ONE,
+          anyOf(
+              isType("com.google.errorprone.refaster.annotation.Placeholder"),
+              isType("com.google.errorprone.refaster.annotation.BeforeTemplate"),
+              isType("com.google.errorprone.refaster.annotation.AfterTemplate")));
+
+  @Override
+  public Description matchClass(ClassTree tree, VisitorState state) {
+    long methodTypes =
```
cushon, line 56: I'd consider implementing this is a `MethodTreeMatcher` instead of a `ClassTreeMatcher` that scans its members for methods, but that depends on whether the check for `< 2` methods below is necessary.

cushon (changes requested):
Thanks for the check!

core/src/main/java/com/google/errorprone/bugpatterns/MissingRefasterAnnotation.java
rickie: We want to partition the methods into two groups: those with a Refaster annotation and those without.
The checker should only flag classes that have methods of both types. 
So far, we have not seen a scenario in which a Refaster template class contains an unannotated method. 
If those do exist, can you share a concrete example, perhaps from Google's vast internal collection of Refaster templates?

core/src/main/java/com/google/errorprone/bugpatterns/MissingRefasterAnnotation.java
rickie: Perhaps a better name would be `methodTypeCount`. 
Since we want to know about all the methods in the class, we use a `ClassTreeMatcher`.

rickie:
Rebased and added a commit. What do you think of the check?

rickie:
Rebased and added a small commit. Still think this check can be useful :).

rickie:
Rebased.

rickie:
Rebased.

cushon (approved):

PR #2246, opened 2021-03-22
Introduce `Refaster.concatLiterals`

### LINKED ISSUE #1496: Possibility for concatenation of string literals in Refaster rules
### Description of the feature request:

I would like to be able to concatenate string literals in the suggested replacement of a Refaster rule, e.g., an existing string literal of the source code (from a parameter of the rule) with a literal that is specified in the rule itself, such that I have `"foobar"` instead of `"foo" + "bar"` as result, where `"foo"` might come from the matched source code.

I have a Refaster rule that replaces `checkArgument(cond, "message " + o);` with `checkArgument(cond, "message " + "%s", o);` (with an arbitrary message). I would like to get `"message %s"` in the suggested replacement.
<details>
<summary>(click here for full rule)</summary>

```
  @BeforeTemplate
  void eager(boolean b, @Matches(CompileTimeConstantExpressionMatcher.class) String msg1, Object o1) {
    Preconditions.checkArgument(b, msg1 + o1);
  }

  @AfterTemplate
  void lazy(boolean b, String msg1, Object o1) {
    Preconditions.checkArgument(b, msg1 + "%s", o1);
  }
```
</details>

### Suggested solution

Another magic method `Refaster.stringConcat(String... parts)` that could be used as `Refaster.stringConcat(msg, "%s")` and would create a string literal with the concatenation of `msg` and `"%s"` if `msg` is a literal.

2.3.4

No.

rickie:
Introduce `Refaster.concatLiterals`

This new Refaster operator can be used in `@AfterTemplate` methods to
request that its operands are concatenated. Literals will be merged
while other operands are joined using the `+` operator.

The PR seems big, but it consists mainly of tests.
The actual code is concise.

Fixes #1496.

CC: @PhilippWendler

### DIFF
--- a/core/src/main/java/com/google/errorprone/refaster/Refaster.java
+++ b/core/src/main/java/com/google/errorprone/refaster/Refaster.java
@@ -227,4 +227,66 @@ public class Refaster {
   public static void emitComment(String literal) {
     throw new UnsupportedOperationException();
   }
+
+  /**
+   * Indicates that Refaster should attempt to concatenate the given expressions, joining literals
+   * if possible.
+   *
+   * <p>For example, instead of writing
+   *
+   * <pre><code>
+   * {@literal @}AfterTemplate void exampleConcatLiterals(boolean b, String msg, Object o) {
+   *     Preconditions.checkArgument(b, msg + "%s", o);
+   *  }
+   * </code></pre>
+   *
+   * <p>you could write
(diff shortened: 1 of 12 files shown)


Stephan202 (approved):
:tada: We learned a lot about the Refaster internals.

PR #2282, opened 2021-04-13
Add more Optional methods to ReturnValueIgnored

pkoenig10:
The return values of almost all `Optional` instance methods, excluding `ifPresent` and `ifPresentOrElse`, should have their return values used.

For cases where you may don't care about the return value, it's better to use `isPresent` and `isEmpty`. For example, instead of:
```java
optional.orElseThrow(() -> ...);
```
Do:
```java
if (optional.isEmpty()) {
    throw ...;
}
```

---

As a motivating example, I had some code that was intended to look like:
```java
if (condition) {
    return getOptional().orElseThrow(...);
}

return ...;
```

But I forgot the `return`, so the code actually looked like:
```java
if (condition) {
    getOptional().orElseThrow(...);
}

return ...;
```

Detecting the unused `Optional` value here would have been useful.

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/ReturnValueIgnored.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/ReturnValueIgnored.java
@@ -30,6 +30,7 @@ import static com.google.errorprone.util.ASTHelpers.isSameType;
 
 import com.google.common.collect.ImmutableSet;
 import com.google.errorprone.BugPattern;
+import com.google.errorprone.ErrorProneFlags;
 import com.google.errorprone.VisitorState;
 import com.google.errorprone.matchers.Matcher;
 import com.google.errorprone.util.ASTHelpers;
@@ -204,16 +205,27 @@ public class ReturnValueIgnored extends AbstractReturnValueIgnored {
   private static final Matcher<ExpressionTree> OPTIONAL_METHODS =
       anyOf(
           staticMethod().onClass("java.util.Optional"),
-          instanceMethod().onExactClass("java.util.Optional").named("filter"),
-          instanceMethod().onExactClass("java.util.Optional").named("flatMap"),
-          instanceMethod().onExactClass("java.util.Optional").named("get"),
-          instanceMethod().onExactClass("java.util.Optional").named("isEmpty"),
-          instanceMethod().onExactClass("java.util.Optional").named("isPresent"),
-          instanceMethod().onExactClass("java.util.Optional").named("map"),
-          instanceMethod().onExactClass("java.util.Optional").named("or"),
-          instanceMethod().onExactClass("java.util.Optional").named("orElse"),
-          instanceMethod().onExactClass("java.util.Optional").named("orElseGet"),
-          instanceMethod().onExactClass("java.util.Optional").named("orElseThrow"));
+          instanceMethod().onExactClass("java.util.Optional")
+              .namedAnyOf(
+                  "isEmpty",
+                  "isPresent"));
+
+  /**
+   * The return values of {@link java.util.Optional} static methods and some instance methods should
+   * always be checked.
+   */
+  private static final Matcher<ExpressionTree> MORE_OPTIONAL_METHODS =
+          anyOf(
+                  instanceMethod().onExactClass("java.util.Optional")
+                          .namedAnyOf(
+                                  "filter",
+                                  "flatMap",
+                                  "get",
+                                  "map",
+                                  "or",
+                                  "orElse",
+                                  "orElseGet",
+                                  "orElseThrow"));
 
   /**
    * The return values of {@link java.util.concurrent.TimeUnit} methods should always be checked.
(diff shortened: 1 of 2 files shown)


Stephan202 (approved):

core/src/main/java/com/google/errorprone/bugpatterns/ReturnValueIgnored.java
```diff
@@ -204,8 +204,16 @@ private static boolean functionalMethod(ExpressionTree tree, VisitorState state)
   private static final Matcher<ExpressionTree> OPTIONAL_METHODS =
       anyOf(
           staticMethod().onClass("java.util.Optional"),
+          instanceMethod().onExactClass("java.util.Optional").named("filter"),
+          instanceMethod().onExactClass("java.util.Optional").named("flatMap"),
+          instanceMethod().onExactClass("java.util.Optional").named("get"),
+          instanceMethod().onExactClass("java.util.Optional").named("isEmpty"),
           instanceMethod().onExactClass("java.util.Optional").named("isPresent"),
-          instanceMethod().onExactClass("java.util.Optional").named("isEmpty"));
+          instanceMethod().onExactClass("java.util.Optional").named("map"),
```
cushon, line 212: FYI we have a separate check just for this one, with a fix that migrates to `ifPresent`: https://github.com/google/error-prone/blob/master/core/src/main/java/com/google/errorprone/bugpatterns/OptionalMapUnusedValue.java. We were planning to add `map` here after that cleanup was done.

core/src/main/java/com/google/errorprone/bugpatterns/ReturnValueIgnored.java
```diff
@@ -204,8 +204,16 @@ private static boolean functionalMethod(ExpressionTree tree, VisitorState state)
   private static final Matcher<ExpressionTree> OPTIONAL_METHODS =
       anyOf(
           staticMethod().onClass("java.util.Optional"),
+          instanceMethod().onExactClass("java.util.Optional").named("filter"),
```
cushon, line 207: This seems like a good use-case for `namedAnyOf`

core/src/main/java/com/google/errorprone/bugpatterns/ReturnValueIgnored.java
```diff
@@ -204,8 +204,16 @@ private static boolean functionalMethod(ExpressionTree tree, VisitorState state)
...
+          instanceMethod().onExactClass("java.util.Optional").named("flatMap"),
+          instanceMethod().onExactClass("java.util.Optional").named("get"),
+          instanceMethod().onExactClass("java.util.Optional").named("isEmpty"),
           instanceMethod().onExactClass("java.util.Optional").named("isPresent"),
-          instanceMethod().onExactClass("java.util.Optional").named("isEmpty"));
+          instanceMethod().onExactClass("java.util.Optional").named("map"),
+          instanceMethod().onExactClass("java.util.Optional").named("or"),
+          instanceMethod().onExactClass("java.util.Optional").named("orElse"),
+          instanceMethod().onExactClass("java.util.Optional").named("orElseGet"),
+          instanceMethod().onExactClass("java.util.Optional").named("orElseThrow"));
```
cushon, line 216: Calling `orElseThrow` to assert an `Optional` is present and then discarding the return result appears to be more common than the other examples here. I'm not really defending that as a pattern, just saying that making this one an error maybe not be lower priority for us. I'm curious if you've run into examples of that?

cushon (commented):
Thanks for this!

We'll probably need to put this behind a flag initially to avoid making an atomic breaking change to an existing error. If you don't mind doing that there's an example in 86f8e239fca1db4e966369902f7a6b9e63580d1b, otherwise I can take care of it when I import the PR.

core/src/main/java/com/google/errorprone/bugpatterns/ReturnValueIgnored.java
pkoenig10: Do you want me to remove `map` from here for now?

core/src/main/java/com/google/errorprone/bugpatterns/ReturnValueIgnored.java
pkoenig10: Occasionally, but rarely.

pkoenig10: As a reference, SonarLint requires that return values are used for all `Optional` methods that return non-void values.
https://github.com/SonarSource/sonar-java/blob/d59d157c852ac924f3029ddb8c0aa366501f9fda/java-checks/src/main/java/org/sonar/java/checks/IgnoredReturnValueCheck.java#L68

cushon (approved):

pkoenig10:
What do you think about matching all non-void instance methods like we do for `Stream`? This would ensure we match against any methods added in the future.

cushon:
> What do you think about matching all non-void instance methods like we do for `Stream`? This would ensure we match against any methods added in the future.

That SGTM, maybe as a follow-up once the flag can be removed?

cushon (approved):

PR #2301, opened 2021-04-23
ImmutableMemberCollection: support field-level `@SuppressWarnings`

Stephan202:
Currently `@SuppressWarnings("ImmutableMemberCollection")` works at the class level but not the field level, which isn't very intuitive; this change resolves that.

### DIFF
--- a/core/src/test/java/com/google/errorprone/bugpatterns/ImmutableMemberCollectionTest.java
+++ b/core/src/test/java/com/google/errorprone/bugpatterns/ImmutableMemberCollectionTest.java
@@ -48,6 +48,23 @@ public final class ImmutableMemberCollectionTest {
   }
 
   @Test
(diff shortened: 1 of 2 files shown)


core/src/test/java/com/google/errorprone/bugpatterns/ImmutableMemberCollectionTest.java
```diff
@@ -47,6 +47,23 @@ public void listInitInline_notMutated_replacesTypeWithImmutableList() {
         .doTest();
   }
 
+  @Test
+  public void setInitConstructor_notMutatedButSuppressed_doesNothing() {
```
Stephan202, line 51: (Tried to follow the test naming convention of other tests in this class.)

cushon (approved):

PR #2376, opened 2021-06-02
Use shaded dataflow artifact that is specific to Error Prone

mernst:
Currently, Error Prone uses the "shaded dataflow" Maven artifact:
https://mvnrepository.com/artifact/org.checkerframework/dataflow-shaded

We have had complaints from users that Error Prone conflicts with other tools that also use the `dataflow-shaded` artifact.  Therefore, we have created a new shaded dataflow artifact just for Error Prone to use:
https://mvnrepository.com/artifact/org.checkerframework/dataflow-errorprone

This pull request updates Error Prone to use that artifact.


cushon:
Thanks for this! I'll take a look at getting this merged. It'll require coordinating with some uses of the existing shaded library that aren't open source, so I'll probably end up submitting a similar change rather than just merging this as-is.

cushon:
https://github.com/google/error-prone/commit/74ca71603612ef14f535594f142f6f5c5be93114

mernst:
Thanks!

PR #2388, opened 2021-06-08
Make installation instructions refer to Error Prone 2.7.1

mernst:
The Ant and command-line instructions referred to Error Prone 2.5.1.

(Maybe the release process is missing a step to update the documentation.)


cushon:
Thanks! There's a step for this in the release process, but we missed a few version numbers at some point and then searching-and-replacing the previous version number stopped working. This should be better-automated.

Issue #2393, opened 2021-06-12
The `WrongOneOf` check also fails with the new arrow-type switch expressions


jsannemo:
> ATTENTION! Please read and follow:
> - if this is a _question_ about Error Prone, send it to error-prone-discuss@googlegroups.com
> - if this is a _bug_ or _feature request_, fill the form below as best as you can.

### Description of the problem / feature request:

The `WrongOneOf` check also fails with the new arrow-type switch expressions.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

`switch (proto.oneOfCase()) {
  case BLAH -> {...};
  default -> {...};
}

### What version of Error Prone are you using?

2.6.0

###  Have you found anything relevant by searching the web?

Same fix required as https://github.com/google/error-prone/pull/2371 and https://github.com/google/error-prone/pull/2124 here

PR #2470, opened 2021-07-27
Fix Issue 2393

openamiguel:
Fix for Issue #2393 (WrongOneof mishandles new style switch statement) based on PR #2371 and PR #2124.


openamiguel:
@googlebot I signed it!

cushon (changes requested):
Thanks for the fix! Can you also add a test?

openamiguel:
Sure. There isn't any WrongOneofTest currently. Can I just focus on testing that this bug pattern works on new-style switch statements?

cushon:
> There isn't any WrongOneofTest currently. Can I just focus on testing that this bug pattern works on new-style switch statements?

Sorry, we have an internal test for this, but it isn't available in open source to avoid the proto dependency in the maven build. Since there isn't already an existing test, this is fine.

cushon (approved):

openamiguel:
@cushon thanks for the approval! Are we waiting for additional review?

cushon:
> Are we waiting for additional review?

Yes, there's a separate internal review step before the PR is merged

PR #2501, opened 2021-08-17
Improve Travis CI build Performance

YunLemon:
According to the official document [Fast Finishing](https://docs.travis-ci.com/user/build-matrix/#fast-finishing), if some rows in the build matrix are allowed to fail, we can add fast_finish: true to the .travis.yml to get faster feedbacks.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ jdk:
   - openjdk-ea
 
 matrix:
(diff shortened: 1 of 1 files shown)


ChenZhangg:
@googlebot I signed it!

PR #2506, opened 2021-08-19
account for switch expressions in UnnecessaryParentheses check

### LINKED ISSUE #1647: False positive: UnnecessaryParentheses in selector expression of switch expressions

A false positive for `UnnecessaryParentheses` is reported when the selector expression of a switch expression is, for example, a method invocation.

````java
class Foo {
    boolean bar() {
        return switch(baz()) {
            case 42 -> true;
            default -> false;
        };
    }

    int baz() {
        return 42;
    }
}
````

2.4.0

No.

### PR DESCRIPTION
**Pull request by codylerum**
Fixes #1647

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/UnnecessaryParentheses.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/UnnecessaryParentheses.java
@@ -30,6 +30,8 @@ import com.google.errorprone.util.ASTHelpers;
 import com.sun.source.tree.ExpressionTree;
 import com.sun.source.tree.ParenthesizedTree;
 import com.sun.source.tree.StatementTree;
+import com.sun.source.tree.SwitchExpressionTree;
+import com.sun.source.tree.Tree;
 
 /** A {@link BugChecker}; see the associated {@link BugPattern} annotation for details. */
 @BugPattern(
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: core/src/main/java/com/google/errorprone/bugpatterns/UnnecessaryParentheses.java [hunk 1]
```diff
@@ -44,7 +46,10 @@
   @Override
   public Description matchParenthesized(ParenthesizedTree tree, VisitorState state) {
     ExpressionTree expression = tree.getExpression();
-    if (state.getPath().getParentPath().getLeaf() instanceof StatementTree) {
+    Tree parent = state.getPath().getParentPath().getLeaf();
+    if (parent instanceof StatementTree) {
+      return NO_MATCH;
+    }else if (parent instanceof SwitchExpressionTree){
```
cushon, line 52: I'm going to change this a bit when I merge the PR, because we want to avoid referencing `SwitchExpressionTree` because that class is only available in JDK >= 12, and we want to remain compatible with JDK 8 for now.

cushon (commented):
Thanks for the fix!

core/src/main/java/com/google/errorprone/bugpatterns/UnnecessaryParentheses.java
codylerum: Sounds Good. As long as you don't need anything else from me I'll let you take it from here.

Issue #1648, opened 2020-05-30
False positive: UnusedVariable for record components that are not read directly


anthonyvdotbe:
### Description of the problem / feature request:

A false positive for `UnusedVariable` is reported for all record components that are only read through methods.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

````java
class Foo {

    record Bar(int v) {}

    int baz() {
        return new Bar(42).v();
    }

}
````

### What version of Error Prone are you using?

2.4.0

###  Have you found anything relevant by searching the web?

No.

**Comment by ferozco** [comment]
We built a [stricter](https://github.com/palantir/gradle-baseline/blob/develop/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StrictUnusedVariable.java) version of `UnusedVariable` in [gradle-baseline](https://github.com/palantir/gradle-baseline/) and we were able to make the rule compatible with records with this [change](https://github.com/palantir/gradle-baseline/pull/1412) which should be directly applicable to `UnusedVariable`

**Comment by codylerum** [comment]
Same hit on this 

Just the most simple record will fail

```
public record AgentAndCount(String agent, long count) {
}
```

```
Warning:  /home/runner/work/oss/oss/src/main/java/co/foo/oss/agent/AgentAndCount.java:[3,34] [UnusedVariable] The field 'agent' is never read.
    (see https://errorprone.info/bugpattern/UnusedVariable)
  Did you mean 'public record AgentAndCount(, long count) {'?
Warning:  /home/runner/work/oss/oss/src/main/java/co/foo/oss/agent/AgentAndCount.java:[3,46] [UnusedVariable] The field 'count' is never read.
    (see https://errorprone.info/bugpattern/UnusedVariable)
  Did you mean 'public record AgentAndCount(String agent) {'?
```

tschniewind:
Still failes if record is private as in
``
class Foo {

  private record Bar(int v) {
  }

  int baz() {
    return new Bar(42).v();
  }

}
``

nvervelle:
> Please reopen, cause it still failes if record is private as in

This is why #2713 was opened :-)

PR #2515, opened 2021-08-23
account for java records when checking for unused variables

codylerum:


codylerum:
Fixed #2515

codylerum:
Fixes #1648

codylerum:
@cushon Should be good for another pass. Test added to cover missing case.

codylerum:
@cushon Just checking too see if this is good from your end? I would imagine you have some behind the scenes steps, but wanted to make sure it wasn't waiting on me.

Same with https://github.com/google/error-prone/pull/2506

core/src/main/java/com/google/errorprone/bugpatterns/UnusedVariable.java
```diff
@@ -223,6 +223,12 @@ public Description matchCompilationUnit(CompilationUnitTree tree, VisitorState s
       if (onlyCheckForReassignments.contains(unusedSymbol) && specs.size() <= 1) {
         continue;
       }
+
+      // Don't complain if the parent of this variable is a Record
+      if(allUsageSites.stream().anyMatch(e -> e.getParentPath().getLeaf().getKind()== Tree.Kind.RECORD)){
+        continue;
+      }
```
cushon, line 230: I'm wondering if this could miss actual unused variables inside record declarations. I'm going to try skipping the synthetic fields for record components with something like the following, but please let me know if you see any issues with that alternative:

```diff
diff --git a/core/src/main/java/com/google/errorprone/bugpatterns/UnusedVariable.java b/core/src/main/java/com/google/errorprone/bugpatterns/UnusedVariable.java
index 3c347ccc9..26a13ced2 100644
--- a/core/src/main/java/com/google/errorprone/bugpatterns/UnusedVariable.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/UnusedVariable.java
@@ -85,6 +85,7 @@ import com.sun.source.util.SimpleTreeVisitor;
 import com.sun.source.util.TreePath;
 import com.sun.source.util.TreePathScanner;
 import com.sun.source.util.TreeScanner;
+import com.sun.tools.javac.code.Flags;
 import com.sun.tools.javac.code.Symbol;
 import com.sun.tools.javac.code.Symbol.MethodSymbol;
 import com.sun.tools.javac.code.Symbol.TypeSymbol;
@@ -610,10 +611,15 @@ public final class UnusedVariable extends BugChecker implements CompilationUnitT
           && ASTHelpers.hasDirectAnnotationWithSimpleName(variableTree, "Inject")) {
         return true;
       }
+      if ((symbol.flags() & RECORD_FLAG) == RECORD_FLAG) {
+        return false;
+      }
       return variableTree.getModifiers().getFlags().contains(Modifier.PRIVATE)
           && !SPECIAL_FIELDS.contains(symbol.getSimpleName().toString());
     }
 
+    private static final long RECORD_FLAG = 1L << 61;
+
     /** Returns whether {@code sym} can be removed without updating call sites in other files. */
     private boolean isParameterSubjectToAnalysis(Symbol sym) {
       checkArgument(sym.getKind() == ElementKind.PARAMETER);
```

cushon (commented):
Thanks for the fix, and the reminder. I'll get this merged soon

core/src/main/java/com/google/errorprone/bugpatterns/UnusedVariable.java
codylerum: They can't really be unused though right? Since they automatically get a public accessor method.

cushon: Record components can't be unused, but I wanted to make it clearer that stuff like the following was still handled:

```java
  public record R (Integer foo, Long bar) {
    void f() {
      int x = foo; // unused
    }
  }
```

codylerum: Makes sense. good catch.

PR #2706, opened 2021-11-23
Refaster: support method invocation type argument inlining

rickie:


core/src/test/java/com/google/errorprone/refaster/UMethodInvocationTest.java
```diff
@@ -41,7 +41,8 @@ public void match() {
     ULiteral oneLit = ULiteral.intLit(1);
     ULiteral barLit = ULiteral.stringLit("bar");
     UMethodInvocation invocation =
-        UMethodInvocation.create(fooIdent, ImmutableList.<UExpression>of(oneLit, barLit));
+        UMethodInvocation.create(
```
rickie, line 44: Should we extend these tests or leave it as is?

core/src/test/java/com/google/errorprone/refaster/testdata/template/TypeArgumentsMethodInvocationTemplate.java
```diff
@@ -0,0 +1,33 @@
...
+/** Sample template for introducing type arguments on a method invocation in an @AfterTemplate. */
+public class TypeArgumentsMethodInvocationTemplate {
+  @BeforeTemplate
+  boolean before(String string) {
+    return string.isEmpty();
+  }
+
+  @AfterTemplate
+  boolean after(String string) {
+    return string.<String>isEmpty();
```
Stephan202, line 31: Discussed offline: surprisingly this is valid code (TIL!), but let's use a realistic example, one with a method that actually accepts a generic type parameter. (For example, `ExecutorService#submit`.)

core/src/test/java/com/google/errorprone/refaster/UMethodInvocationTest.java
Stephan202: The unification logic does not consider the type parameters (should it?), so I guess right now there's nothing to test. Maybe we can use the varargs overload instead:
```java
UMethodInvocation.create(fooIdent, oneLit, barLit);
```
(But not sure that makes the test easier to understand :thinking:.)

That said, perhaps we could add an extra inline test below, e.g. one that handles two type arguments (since the Refaster test further down already covers the one-type argument case).

Stephan202 (commented):
Maybe we can provide a motivating example for this feature. It really helped during your thesis work to ensure that Refaster-patched code still compiled. An example where this was useful, was the wrapping of expressions that implemented a functional interface. In some cases these expressions were method references or lambda expressions whose exact signature was determined by their context. Wrapping such expression without specifying explicit type parameters (inferred from context) would lead to non-compilable code.

core/src/test/java/com/google/errorprone/refaster/UMethodInvocationTest.java
rickie: I tried adding the `typeArguments` to the unification in `UMethodInvocation`. However, the [ImplicitTypesInlinedTemplate](https://github.com/google/error-prone/blob/master/core/src/test/java/com/google/errorprone/refaster/testdata/template/ImplicitTypesInlinedTemplate.java) fails after doing so, specifically on [this example](https://github.com/google/error-prone/blob/master/core/src/test/java/com/google/errorprone/refaster/testdata/input/ImplicitTypesInlinedTemplateExample.java#L32).
The unification fails simply because the expression in the `@BeforeTemplate` does not have type arguments, but the matched code does.

We could try to define the semantics in such a way that the unification process also accounts for implicit type arguments, but it feels like a bridge too far for this PR. So I would suggest to leave it as is. Let me know what you think.

core/src/test/java/com/google/errorprone/refaster/testdata/template/TypeArgumentsMethodInvocationTemplate.java
rickie: Yes, just pushed the update, now it uses a realistic example.

Stephan202 (approved):
Rebased. The test isn't "exciting", but it works :smile:

Suggested commit message:
```
Refaster: support method invocation type argument inlining
```

I think this PR is ready for review by the Error Prone team; it'd be nice to bring support for method invocations on par with constructor invocations when it comes to type arguments.

rickie:
Within de Picnic codebase we found more situations where "losing" type arguments during a rewrite can lead to non-compilable code. In the example below the explicit type arguments in `ImmutableMap.<K, V>builder()` currently don't have any effect, as Refaster currently does not support inlining such type arguments for method invocations.

```java
static final class ImmutableMapBuilder<K, V> {
  @BeforeTemplate
  ImmutableMap.Builder<K, V> before() {
    return new ImmutableMap.Builder<>();
  }

  @AfterTemplate
  ImmutableMap.Builder<K, V> after() {
    return ImmutableMap.<K, V>builder();
  }
}
```
The changes in this PR resolve this issue.
If desired we can update the test in this PR to use the above example instead of the existing `ExecutorService#submit` example. @cushon if any other changes are required let us know :).

Stephan202:
Rebased and squashed the two commits, as I'll cherry-pick the changes of this PR to our fork. That said: still interested in seeing this included in mainline Error Prone :upside_down_face:.

rickie:
Rebased. Still curious, could this change still get into mainline Error Prone 😄?

rickie:
Rebased. Is it possible to get this change into mainline Error Prone :smile:?

rickie:
Rebased.

cushon (approved):
Sorry for not getting to this sooner. I did some testing with our internal refactor templates, and it looks good. Thanks for the PR!

rickie:
Oh nice, good to hear! 😄 

I added a commit to update the year numbers in the license headers and rebased the PR.

PR #2776, opened 2021-12-23
UnnecessarilyFullyQualified: handle `java.lang` and inconsistent import usages

rickie:
This introduces two improvements:
- Where possible `java.lang.SomeType` references are simplified to `SomeType`
  without introducing an associated `import` statement.
- References to `some.package.SomeType` are now also replaced with `SomeType`
  if said type is already imported and referenced by its simple name elsewhere
  in the compilation unit.


Stephan202 (approved):
Tested against the Picnic code base with positive results :+1:

cushon (approved):

Stephan202:
Rebased the branch so that the JDK 18-ea build now passes. (IIUC @cushon's approval already caused a Google-internal PR to be opened, so this is mostly for bookkeeping purposes.)

PR #2798, opened 2022-01-04
implement FormatString for Java 15 String#formatted

### LINKED ISSUE #2726: Support String.formatted with FormatString Bug Pattern
Added in JDK 15 is the ability for call`formatted` on a String.

https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/String.html#formatted(java.lang.Object...)

`"foo %s".formatted("bar");`

It would be great if the https://errorprone.info/bugpattern/FormatString check was enforced on this.

codylerum:
Here is an initial pass at an implementation for java 15 and up String.formatted

What I'm doing here is detecting early that this is a formatted call on a string instance and just grabbing the value and all of the arguments and sending them directly to a new validate method.

The old validate method is renamed prepareAndValidate and does the work of extracting the pattern and arguments from the argument this for your traditional String#format calls

fixes #2726

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/formatstring/FormatString.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/formatstring/FormatString.java
@@ -17,6 +17,7 @@
 package com.google.errorprone.bugpatterns.formatstring;
 
 import static com.google.errorprone.BugPattern.SeverityLevel.ERROR;
+import static com.google.errorprone.matchers.method.MethodMatchers.instanceMethod;
 
 import com.google.common.collect.ImmutableList;
 import com.google.errorprone.BugPattern;
@@ -24,6 +25,7 @@ import com.google.errorprone.VisitorState;
 import com.google.errorprone.bugpatterns.BugChecker;
 import com.google.errorprone.bugpatterns.BugChecker.MethodInvocationTreeMatcher;
 import com.google.errorprone.matchers.Description;
+import com.google.errorprone.matchers.Matcher;
 import com.google.errorprone.util.ASTHelpers;
 import com.sun.source.tree.ExpressionTree;
 import com.sun.source.tree.MethodInvocationTree;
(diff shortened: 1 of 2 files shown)


core/src/test/java/com/google/errorprone/bugpatterns/formatstring/FormatStringTest.java
```diff
@@ -365,4 +365,33 @@ public void invalidIndex() {
...
+        .addSourceLines(
+            "Test.java",
+            "class Test {",
+            "  void f() {",
+            "    // BUG: Diagnostic contains: missing argument for format specifier",
+            "    \"%s %s %s\".formatted(\"foo\", \"baz\");",
+            "  }",
+            "}")
+        .doTest();
+  }
```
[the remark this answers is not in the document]
cushon: Thanks!

core/src/test/java/com/google/errorprone/bugpatterns/formatstring/FormatStringTest.java
```diff
@@ -16,14 +16,14 @@
...
 
-import static org.junit.Assume.assumeTrue;
-
 import com.google.errorprone.CompilationTestHelper;
 import com.google.errorprone.util.RuntimeVersion;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
 
+import static org.junit.Assume.assumeTrue;
```
cushon, line 25: nit: can you leave the original import order?

core/src/main/java/com/google/errorprone/bugpatterns/formatstring/FormatString.java
```diff
@@ -33,17 +36,41 @@
 @BugPattern(name = "FormatString", summary = "Invalid printf-style format string", severity = ERROR)
 public class FormatString extends BugChecker implements MethodInvocationTreeMatcher {
 
+  private static final Matcher<ExpressionTree> FORMATTED_METHOD = instanceMethod().onExactClass("java.lang.String").named("formatted");
+
   @Override
   public Description matchMethodInvocation(MethodInvocationTree tree, final VisitorState state) {
-    ImmutableList<ExpressionTree> args = FormatStringUtils.formatMethodArguments(tree, state);
-    if (args.isEmpty()) {
+    final ImmutableList<ExpressionTree> args;
```
cushon, line 43: nit: house style is to usually leave `final` off variables, can you leave these out?

(I realize there's a `final VisitorState state` right above this line...)

codylerum:
@cushon That worked well and is a much better way of getting the pattern. I'm just not familiar enough with all the tools available when parsing the source code.

I squashed everything into one commit and updated the formatting and import order.

cushon (approved):

PR #2799, opened 2022-01-04
intellij project files update

codylerum:
Current versions of Intellij are failing to parse the current vcs.xml file as it has a comment before the optional xml declaration.

image

I've updated that to not complain as well as added another entry for the .gitignore for a common intellij plugin. This should make it easier for contributors to open the project. Currently on a fresh clone you get a dirty vcs.xml file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 
 # intellij
 .idea/ant.xml
+/.idea/codestream.xml
 .idea/codeStyleSettings.xml
 .idea/compiler.xml
 .idea/copyright
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,7 @@
 
 # intellij
 .idea/ant.xml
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -8,6 +8,7 @@
 
 # intellij
 .idea/ant.xml
+/.idea/codestream.xml
```
Stephan202, line 11: For consistency I guess this one should also be declared relative?
```suggestion
.idea/codestream.xml
```

Stephan202 (approved):
Thanks for opening this PR; @rickie and I worked around this every time, but it is a bit annoying indeed :)

.gitignore
codylerum: works for me. updated.

cushon (approved):

cushon (approved):

PR #2805, opened 2022-01-05
Refaster: fix expression matching in absence of additional syntactic context

rickie:
`@BeforeTemplate` implementations of the form `return someParam;` are valid
constructs. With these changes such templates behave as expected, such that
`someParam` only matches expressions of the appropriate type, not references to
the type itself.

### DIFF
--- a/core/src/main/java/com/google/errorprone/refaster/RefasterScanner.java
+++ b/core/src/main/java/com/google/errorprone/refaster/RefasterScanner.java
@@ -28,6 +28,7 @@ import com.sun.source.tree.ClassTree;
 import com.sun.source.tree.DoWhileLoopTree;
 import com.sun.source.tree.IfTree;
 import com.sun.source.tree.MethodTree;
+import com.sun.source.tree.PackageTree;
 import com.sun.source.tree.ParenthesizedTree;
 import com.sun.source.tree.SynchronizedTree;
 import com.sun.source.tree.Tree;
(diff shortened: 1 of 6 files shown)


Stephan202 (approved):
This approach was tested extensively in the context of @rickie's Msc thesis :+1:

rickie:
Rebased and added a commit to be compatible with the latest changes.

rickie:
Rebased and resolved conflict. This still is a nice improvement :).

Stephan202:
Some context, since one might ask "why would you want to define such a `@BeforeTemplate`?": by supporting this case, it becomes possible for Refaster to "wrap" all expressions of a given type. This proved useful in the context of Rick's [MSc thesis](https://zenodo.org/record/5801922/files/thesis_library_migration_rick_ossendrijver_final_publication_zenodo.pdf), in which he used Error Prone and Refaster to implement an automated migration from RxJava to Project Reactor. ([Shorter paper on the same topic](https://dl.acm.org/doi/pdf/10.1145/3477314.3507153).)

In case there are questions about this, I'm sure that @rickie will gladly answer them. ;)

rickie:
Rebased.

rickie:
Rebased. 

@cushon I'm curious to hear your thoughts on this PR. Do you have some time to look into this?

This is one of the PRs that we worked on in the context of my thesis as @Stephan202 mentioned in his previous comment.
We now even have a [second paper published](https://dl.acm.org/doi/10.1145/3594264.3594265) about this subject. If you have any questions about it, let me know!

PR #2820, opened 2022-01-08
Make `SuggestedFixes.updateAnnotationArgumentValues` JDK 17-compatible

Stephan202:
When running Java 17, `AnnotationTree#toString()` omits parentheses if
the annotation tree does not contain any arguments. With this change
`#updateAnnotationArgumentValues` no longer relies on the unconditional
presence of such parentheses.

### DIFF
--- a/check_api/src/main/java/com/google/errorprone/fixes/SuggestedFixes.java
+++ b/check_api/src/main/java/com/google/errorprone/fixes/SuggestedFixes.java
@@ -1143,9 +1143,12 @@ public final class SuggestedFixes {
       return SuggestedFix.builder()
           .replace(
               annotation,
-              annotation
-                  .toString()
-                  .replaceFirst("\\(\\)", "(" + parameterPrefix + newArgument(newValues) + ")"));
+              '@'
+                  + annotation.getAnnotationType().toString()
+                  + '('
+                  + parameterPrefix
+                  + newArgument(newValues)
+                  + ')');
     }
     Optional<ExpressionTree> maybeExistingArgument = findArgument(annotation, parameterName);
     if (!maybeExistingArgument.isPresent()) {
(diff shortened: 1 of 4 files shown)


check_api/src/main/java/com/google/errorprone/fixes/SuggestedFixes.java
```diff
@@ -1143,9 +1143,12 @@ private static Tree suppressibleNode(TreePath path) {
       return SuggestedFix.builder()
           .replace(
               annotation,
-              annotation
-                  .toString()
-                  .replaceFirst("\\(\\)", "(" + parameterPrefix + newArgument(newValues) + ")"));
+              '@'
+                  + annotation.getAnnotationType().toString()
```
Stephan202, line 1147: More correct would be to do `state.getSourceForNode(annotation.getAnnotationType())`, but this method currently does not accept a `VisitorState`. Let me know if you'd like me to make this change; it appears that all callers do have a `VisitorState` in scope. (Well, in the open source repository, at least.)

Stephan202:
Tests indicate that this issue applies to JDK 12+.

check_api/src/main/java/com/google/errorprone/fixes/SuggestedFixes.java
cushon: Thanks for the fix!

Adding passing in the `VisitorState` as a parameter and using `state.getSourceForNode(annotation.getAnnotationType())` SGTM if you want to make that change, but this is a clear improvement as-is, so either way is OK with me.

cushon (approved):

check_api/src/main/java/com/google/errorprone/fixes/SuggestedFixes.java
Stephan202: Alright, I added one more commit, though arguably the improvement is minimal. (But I can imagine that at some point [`TreeToString`](https://github.com/google/error-prone/blob/master/core/src/main/java/com/google/errorprone/bugpatterns/TreeToString.java) is extended to cover non-`BugChecker` classes, in which case this change would have to be made anyway.)

cushon:
Sorry, I started merging the PR before your last commit, I'll import that separately

PR #2873, opened 2022-01-18
Check for case where docs have not been updated

msridhar:
This will fix the issue of the doc upload job showing as failed when there are no doc changes.


cushon (approved):
Thanks!

cushon (approved):

Issue #2881, opened 2022-01-21
UnusedMethod is reported for JUnit @BeforeEach and for Java validation annotated methods


sergeykad:
The problem observed with error-prone 2.10.0

The problem is reported for methods with well-known annotations like `org.junit.jupiter.api.BeforeEach`,  `javax.validation.constraints.AssertTrue` and `com.fasterxml.jackson.annotation.JsonProperty`

The reported methods are indeed not used directly by our code, but by a dependency, like JUnit 5 or Jackson. It seems crude to annotate every such method with `@SuppressWarnings("UnusedMethod")`

There is a similar problem with "Unused variable" reported for class members annotated with  `org.springframework.boot.test.mock.mockito.MockBean`.

cushon:
Both of those checks have heuristics with lists of annotations they recognize as meaning the variable or method isn't actually unused:

* https://github.com/google/error-prone/blob/master/core/src/main/java/com/google/errorprone/bugpatterns/UnusedMethod.java#L93
* https://github.com/google/error-prone/blob/master/core/src/main/java/com/google/errorprone/bugpatterns/UnusedVariable.java#L129

I'm happy to take PRs adding annotations to those lists.

Also FYI the next release of Error Prone will have a `@Keep` annotation that can be applied to other annotations, which will suppress those warnings on annotated elements. Not all projects are going to want to or be able to depend on Error Prone's annotations, but it provides an extensible way to add this kind of suppression.

PR #2947, opened 2022-02-01
Added exempting variable annotations

sergeykad:
Partially fixes https://github.com/google/error-prone/issues/2881

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/UnusedVariable.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/UnusedVariable.java
@@ -136,7 +136,10 @@ public final class UnusedVariable extends BugChecker implements CompilationUnitT
           "org.openqa.selenium.support.FindBy",
           "org.openqa.selenium.support.FindBys",
           "org.apache.beam.sdk.transforms.DoFn.TimerId",
-          "org.apache.beam.sdk.transforms.DoFn.StateId");
+          "org.apache.beam.sdk.transforms.DoFn.StateId",
+          "org.mockito.Mock",
+          "org.springframework.boot.test.mock.mockito.MockBean"
+      );
 
   private final ImmutableSet<String> methodAnnotationsExemptingParameters;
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/google/errorprone/bugpatterns/UnusedVariable.java
```diff
@@ -136,7 +136,10 @@
           "org.openqa.selenium.support.FindBy",
           "org.openqa.selenium.support.FindBys",
           "org.apache.beam.sdk.transforms.DoFn.TimerId",
-          "org.apache.beam.sdk.transforms.DoFn.StateId");
+          "org.apache.beam.sdk.transforms.DoFn.StateId",
+          "org.mockito.Mock",
```
cushon, line 140: Why should `@Mock` variables be exempted, isn't it generally unnecessary to have unused mocks?

cushon (changes requested):

core/src/main/java/com/google/errorprone/bugpatterns/UnusedVariable.java
sergeykad: It can be used indirectly, for example by `@InjectMocks` annotation.

cushon: What do you think about adding `@InjectMocks` to the list instead? I'd like to keep the ability to recognize unused fields that are directly annotated with `@Mock`

sergeykad: `@InjectMocks` is added to another field, so it does not really help. Only Mockito itself can know for sure if it is used and complain about it, which it does.

If you are feeling that it is not good enough I will remove the `@Mock` annotation altogether.

sergeykad: So, what should I do?

cushon: (Sorry for losing track of this.)

I think for the uses of UnusedVariable in my employer's codebase, we'd probably rather have the false positive than the false negatives.

An alternative to enabling `@Mock` by default could be to add a flag (similar to the existing `-XepOpt:Unused:methodAnnotationsExemptingParameters` option supported in this checker) to allow configuring it to skip fields annotated with `@Mock`. What do you think?

cushon (changes requested):

PR #3008, opened 2022-02-24
Close verbatim in 3rd column header to fix table rendering

pzygielo:
The key change is in l. 14 only. Lines 15-20 only add single space.

Change [StringSplitter](https://errorprone.info/bugpattern/StringSplitter)

ep-before

to

ep-after


pzygielo:
@cushon, @graememorgan - how could I have it reviewed?

cushon (approved):

PR #3012, opened 2022-02-24
Add UUID.fromString() check to AlwaysThrows

### LINKED ISSUE #3006: Add UUID check to AlwaysThrows?
It looks like the AlwaysThrows check could check UUID.fromString() too.

sixcorners:
Fixes #3006

### DIFF
--- a/core/src/test/java/com/google/errorprone/bugpatterns/AlwaysThrowsTest.java
+++ b/core/src/test/java/com/google/errorprone/bugpatterns/AlwaysThrowsTest.java
@@ -43,6 +43,21 @@ public class AlwaysThrowsTest {
   }
 
   @Test
+  public void positiveUUID() {
+    testHelper
+        .addSourceLines(
+            "Test.java",
+            "import java.util.UUID;",
+            "class T { ",
+            "  void f() {",
+            "    // BUG: Diagnostic contains: will fail at runtime with a IllegalArgumentException",
+            "    UUID.fromString(\"677927337fc742eab14f7ee44c924739\");",
+            "  }",
+            "}")
+        .doTest();
+  }
+
+  @Test
   public void immutableMapThrows() {
     testHelper
         .addSourceLines(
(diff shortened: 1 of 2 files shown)


core/src/test/java/com/google/errorprone/bugpatterns/AlwaysThrowsTest.java
```diff
@@ -42,6 +42,21 @@ public void positive() {
...
 
+  @Test
+  public void positiveUUID() {
+    testHelper
+        .addSourceLines(
+            "Test.java",
+            "import java.util.UUID;",
+            "class T { ",
+            "  void f() {",
+            "    // BUG: Diagnostic contains: will fail at runtime with a IllegalArgumentException",
```
sixcorners, line 53: should this be "an IllegalArgumentException"?

kluever:
I'm confused...

`UUID.fromString(...)` only [throws](https://docs.oracle.com/javase/8/docs/api/java/util/UUID.html#fromString-java.lang.String-) "...If name does not conform to the string representation as described in [toString()](https://docs.oracle.com/javase/8/docs/api/java/util/UUID.html#toString--)"

cushon (approved):

cushon:
I think this addition LGTM

> I'm confused...
> 
> `UUID.fromString(...)` only [throws](https://docs.oracle.com/javase/8/docs/api/java/util/UUID.html#fromString-java.lang.String-) "...If name does not conform to the string representation as described in [toString()](https://docs.oracle.com/javase/8/docs/api/java/util/UUID.html#toString--)"

Maybe the check could be better named, go/AlwaysThrows reports calls of APIs where the arguments are available at compile-time, and we can tell that the call will throw, e.g. it would report an error for `UUID.fromString("nope")`.

core/src/test/java/com/google/errorprone/bugpatterns/AlwaysThrowsTest.java
cushon: I suppose it should, we could consider updating [the logic that constructs the diagnostic](https://github.com/google/error-prone/pull/3012/files#diff-870a751c3cdcc95fa86d5c4bcceba95518f3d400c7752641785735b03a74bbb1R189) to check for vowels as a follow-up

PR #3032, opened 2022-03-06
Create mods.toml

rtfmoz2:
Mod loader information for Forge FTB Minecraft


cushon (changes requested):
I don't know much about Minecraft, can you share some context on what Forge FTB Minecraft is and why we should add this?

PR #3036, opened 2022-03-08
Disallowing extension of Throwable

### LINKED ISSUE #167: Disallow extending Throwable
_[Original issue](https://code.google.com/p/error-prone/issues/detail?id=167) created by **eaftan@google.com** on 2013-08-01 at 12:19 AM_

---

Java allows you to extend Throwable, but it's generally a bad idea.  Most error handling code assumes that the only direct subtypes of Throwable are Exception and Error, and do not deal correctly with other subtypes.

samuelsoder:
A small check warning users when extending throwable. Should close #167.


cushon:
Hi @samuelsoder,

Thanks for this PR, and sorry for the very slow response!

We can't accept this contribution without a Contributor License Agreement (CLA), see: https://cla.developers.google.com/

PR #3037, opened 2022-03-08
'this' escapes constructor bugpattern

### LINKED ISSUE #184: Warning: 'this' escapes constructor
_[Original issue](https://code.google.com/p/error-prone/issues/detail?id=184) created by **kevinb@google.com** on 2013-09-11 at 05:28 PM_

---

A constructor body should never pass 'this' ... anywhere?  In general APIs do not expect to receive partly-constructed instances!  I suppose there might be exceptions to this rule.  We'll have to see what the violations look like.

tjog:
A simple checker for the issue #184 . Does not use more advanced features such as dataflow to detect the pattern more accurately (that is if everything has been initialized at that point/expression).

Closes #184.


cushon:
I added a note to https://github.com/google/error-prone/issues/184 with some more history.

Based on our experience with the previous `ConstructorLeaksThis` check, this isn't a good fit for core Error Prone unless we can find additional heuristics to reduce the false-positive rate. (Also as I noted in the bug, this may be unnecessary to add to Error Prone now that javac has a related built-in diagnostic.)

PR #2788, opened 2021-12-28
Update maven docs to recommend .mvn/jvm.config

mattnelson:
Doc updates from #2786

Issue #2786, opened 2021-12-28
Maven jvm.config instead of fork/compilerArgs


mattnelson:
The error prone maven instructions for jdk16+[^1] recommend using `compilerArgs` and `fork`. I was able to supply those same args to maven itself via `.mvn/jvm.config` file[^2] and get a 4x improvement of build times by no longer forking the compiler.

Anecdotally, with my sample size of 1, using a maven reactor with 80 modules the build time for `mvn clean compile` was 11s using `.mvn/jvm.config` and 44s when forking the compiler.

The maven-compiler-plugin docs[^3] leads me to believe that `fork` was designed for use cases where a different JDK was wanting to be used, rather than for configuration isolation while using the default javac. Thats not to say it can't also work when config isolation is desired, like in this case.

Does errorprone need that level of isolation on the exports/opens configuration? Would the maintainers be open to updating the docs to include the `.mvn/jvm.config` option with the pros/con between the two approaches? or if the config isolation is not needed and the performance impact is impactful enough, change the recommendation to the non-forking option.

[^1]: https://errorprone.info/docs/installation#jdk-16
[^2]: https://maven.apache.org/configure.html#mvn-jvm-config-file
[^3]: https://maven.apache.org/plugins/maven-compiler-plugin/examples/compile-using-different-jdk.html

cushon:
Thanks for the suggestion! Using `.mvn/jvm.config` sounds like a good approach. We aren't relying on `fork` for isolation, it's really just about having a way to pass the JVM flags to javac, and passing `-J` flags to javac only works with `fork` enabled.

I'm trying this out on a couple of projects and so far everything looks good.

Any interest in sending a PR to update the docs to suggest `.mvn/jvm.config` instead of `fork`?

mattnelson:
> Any interest in sending a PR to update the docs to suggest `.mvn/jvm.config` instead of `fork`?

Sure, PR issued at #2788

hisener:
> get a 4x improvement of build times by no longer forking the compiler.

@mattnelson Funny enough, I applied the exact change on Tuesday, and the build times went from ~20 minutes to 8-9 minutes. However, the downside is that the Metaspace usage went up because the number of classes loaded increased significantly. Did you experience the same?

---

> it's really just about having a way to pass the JVM flags to javac, and passing `-J` flags to javac only works with `fork` enabled.

[...]

PR #3045, opened 2022-03-13
Clarify Maven configuration

anthonyvdotbe:
The current way of using `.mvn/jvm.config` only works if the `maven-compiler-plugin` doesn't use an external executable.
The [previous](https://github.com/google/error-prone/pull/2788/files#diff-8b042e3f94ca5c59a7cd990b950aec0073ea84fe811e7b22be51158d7b180d56) way of using `compilerArgs` only works if the `maven-compiler-plugin` *does* use an external executable.

So both ways should be documented, such that users can select the way that works for their Maven setup.
Also note that setting `fork` to `true` is just one way to get the `maven-compiler-plugin` to use an external executable. Enabling the `maven-toolchains-plugin` is another.

PS: not sure whether I should create an issue for this, or if this can go under https://github.com/google/error-prone/issues/2786?

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -46,7 +46,7 @@ Edit your `pom.xml` file to add settings to the maven-compiler-plugin:
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
-        <version>3.8.0</version>
+        <version>3.10.1</version>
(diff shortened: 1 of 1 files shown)


cushon:
Thanks! Merged as https://github.com/google/error-prone/commit/175b9db5c87dc8f7f7f51bd97cbf8bd6eed78d1a

Issue #2518, opened 2021-08-25
Invalid `@UnnecessaryLambda` patch suggestion


Stephan202:
Consider the following code:
```java
import java.util.stream.IntStream;

class Example {
  void someLoopyCode() {
    for (int i : someIterable()) {
      // Do something.
    }
  }

  private Iterable<Integer> someIterable() {
    return () -> IntStream.range(0, 42).boxed().iterator();
  }
}
```

Compile the code as follows:
```sh
javac \
  -XDcompilePolicy=simple \
  -processorpath error_prone_core-2.9.0-with-dependencies.jar \
  '-Xplugin:ErrorProne -XepDisableAllChecks -Xep:UnnecessaryLambda:WARN' \
  Example.java
```

This yields the following warning:
```
Example.java:10: warning: [UnnecessaryLambda] Returning a lambda from a helper method or saving it in a constant is unnecessary; prefer to implement the functional interface method directly and use a method reference instead.
  private Iterable<Integer> someIterable() {
                            ^
    (see https://errorprone.info/bugpattern/UnnecessaryLambda)
  Did you mean 'for (int i : this::someIterable) {'?
1 warning
```

If this suggestion were accepted using `-XepPatchChecks:UnnecessaryLambda` the resulting code would not compile, producing a "_method reference not expected here_" error. Possible solutions:
1. Don't flag enhanced for-loop method calls.
2. Introduce an explicit cast (i.e., suggest `for (int i : (Iterable<Integer>) this::someIterable) {` instead).

cushon:
https://github.com/google/error-prone/commit/93ca62e71a836684d2fe103883f37e7976b69f0d

PR #3234, opened 2022-05-29
fix issue#2518

jingjing-0919:
add a fix to  [issue#2518](https://github.com/google/error-prone/issues/2518)


jingjing-0919:
add a fix to issue#2518

cushon (commented):
Thanks for the PR!

This has been fixed a slightly different way in https://github.com/google/error-prone/commit/93ca62e71a836684d2fe103883f37e7976b69f0d

PR #3249, opened 2022-06-07
Add flag for extra annotations to FieldCanBeFinal

jtraglia:
I would like to use `FieldCanBeFinal` with my project but there are lots of false positives with [`picocli`](https://picocli.info) and [`fasterxml.jackson`](https://github.com/FasterXML/jackson). I think having the ability to configure extra non-constant variable annotations would be useful.


jtraglia:
It's been a month. Closing.

Issue #3225, opened 2022-05-26
NPE analysing new style switch statement (2.14.0)


rob-donovan:
Hello, 

Since upgrading to version 2.14.0, seeing the following NPE being thrown. It appears this is occurring whilst analysing a method with a new style switch statement where the result is a Supplier of an Optional. For example:

```
Supplier<Optional<Record>> = switch(enumValue) {
    case VALUE1 -> () -> ....
    case VALUE2 -> () -> .....
};
```

Changing this to an old-style switch statement prevents the NPE.

```
14:49:32          error-prone version: 2.14.0
14:49:32          BugPattern: Immutable
14:49:32          Stack Trace:
14:49:32          java.lang.NullPointerException: Cannot invoke "com.google.errorprone.util.ASTHelpers$TargetType.type()" because the return value of "com.google.errorprone.util.ASTHelpers.targetType(com.google.errorprone.VisitorState)" is null
14:49:32         at com.google.errorprone.bugpatterns.threadsafety.ImmutableChecker.matchMemberReference(ImmutableChecker.java:150)
14:49:32         at com.google.errorprone.scanner.ErrorProneScanner.processMatchers(ErrorProneScanner.java:449)
14:49:32         at com.google.errorprone.scanner.ErrorProneScanner.visitMemberReference(ErrorProneScanner.java:715)
14:49:32         at com.google.errorprone.scanner.ErrorProneScanner.visitMemberReference(ErrorProneScanner.java:150)
14:49:32         at jdk.compiler/com.sun.tools.javac.tree.JCTree$JCMemberReference.accept(JCTree.java:2546)
14:49:32         at jdk.compiler/com.sun.source.util.TreePathScanner.scan(TreePathScanner.java:86)
14:49:32         at com.google.errorprone.scanner.Scanner.scan(Scanner.java:74)
14:49:32         at com.google.errorprone.scanner.Scanner.scan(Scanner.java:48)
14:49:32         at jdk.compiler/com.sun.source.util.TreeScanner.scanAndReduce(TreeScanner.java:96)
14:49:32         at jdk.compiler/com.sun.source.util.TreeScanner.visitCase(TreeScanner.java:401)
14:49:32         at com.google.errorprone.scanner.ErrorProneScanner.visitCase(ErrorProneScanner.java:534)
14:49:32         at com.google.errorprone.scanner.ErrorProneScanner.visitCase(ErrorProneScanner.java:150)
[...]

PR #3267, opened 2022-06-15
ImmutableChecker handles null types

### LINKED ISSUE #3220: NullPointerException thrown during analysis
After Upgrading from 2.12.1 I get this error during analysis:

```
C:\Users\axel\IdeaProjects\utility\utility\build\generated-sources\cabe\com\dua3\utility\math\MathUtil.java:478: error: An unhandled exception was thrown by the Error Prone static analysis plugin.
                x -> x >= 0 ? java.lang.Math.ceil(x - 0.5) : java.lang.Math.floor(x + 0.5);             case HALF_EVEN ->                 java.lang.Math::rint;
                                                                                                                                          ^

     error-prone version: 2.14.0
     BugPattern: Immutable
     Stack Trace:
     java.lang.NullPointerException: Cannot invoke "com.google.errorprone.util.ASTHelpers$TargetType.type()" because the return value of "com.google.errorprone.util.ASTHelpers.targetType(com.google.errorprone.VisitorState)" is null   
        at com.google.errorprone.bugpatterns.threadsafety.ImmutableChecker.matchMemberReference(ImmutableChecker.java:150)
        at com.google.errorprone.scanner.ErrorProneScanner.processMatchers(ErrorProneScanner.java:449)
        at com.google.errorprone.scanner.ErrorProneScanner.visitMemberReference(ErrorProneScanner.java:715)
        at com.google.errorprone.scanner.ErrorProneScanner.visitMemberReference(ErrorProneScanner.java:150)
        at jdk.compiler/com.sun.tools.javac.tree.JCTree$JCMemberReference.accept(JCTree.java:2546)
        at jdk.compiler/com.sun.source.util.TreePathScanner.scan(TreePathScanner.java:86)
        at com.google.errorprone.scanner.Scanner.scan(Scanner.java:74)
        at com.google.errorprone.scanner.Scanner.scan(Scanner.java:48)
```

schlosna:
`ASTHelpers#getType(ClassTree)` and `ASTHelpers#targetType(VisitorState)` both may return null, which ImmutableChecker did not previously handle safely, which could lead to `NullPointerException` in cases such as switch expressions. This PR handles these potential `null` results more safely.

Fixes #3220 & #3225

### DIFF
--- a/check_api/src/main/java/com/google/errorprone/util/ASTHelpers.java
+++ b/check_api/src/main/java/com/google/errorprone/util/ASTHelpers.java
@@ -1754,14 +1754,30 @@ public class ASTHelpers {
 
     @Override
     public Type visitCase(CaseTree tree, Void unused) {
-      Tree t = parent.getParentPath().getLeaf();
-      // JDK 12+, t can be SwitchExpressionTree
-      if (t instanceof SwitchTree) {
-        SwitchTree switchTree = (SwitchTree) t;
-        return getType(switchTree.getExpression());
+      Tree switchTree = parent.getParentPath().getLeaf();
+      return getType(getSwitchExpression(switchTree));
+    }
+
+    @Nullable
+    private static ExpressionTree getSwitchExpression(Tree tree) {
+      if (tree instanceof SwitchTree) {
+        return ((SwitchTree) tree).getExpression();
       }
-      // TODO(b/176098078): When the ErrorProne project switches to JDK 12, we should check
-      // for SwitchExpressionTree.
+      // Reflection is required for JDK < 12
+      try {
+        Class<?> switchExpression = Class.forName("com.sun.source.tree.SwitchExpressionTree");
+        Class<?> clazz = tree.getClass();
+        if (switchExpression.isAssignableFrom(clazz)) {
+          try {
+            Method method = clazz.getMethod("getExpression");
+            return (ExpressionTree) method.invoke(tree);
+          } catch (ReflectiveOperationException e) {
+            throw new LinkageError(e.getMessage(), e);
+          }
+        }
+      } catch (ClassNotFoundException e) {
+        // continue below
+      }
       return null;
     }
 
--- a/core/src/main/java/com/google/errorprone/bugpatterns/threadsafety/ImmutableChecker.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/threadsafety/ImmutableChecker.java
@@ -120,15 +120,11 @@ public class ImmutableChecker extends BugChecker
 
   @Override
   public Description matchLambdaExpression(LambdaExpressionTree tree, VisitorState state) {
-    Type type = getType(tree);
-    if (type == null) {
-      return NO_MATCH;
-    }
-    TypeSymbol lambdaType = type.tsym;
+    TypeSymbol lambdaType = getType(tree).tsym;
     ImmutableAnalysis analysis = createImmutableAnalysis(state);
     Violation info =
         analysis.checkInstantiation(
-            lambdaType.getTypeParameters(), type.getTypeArguments());
+            lambdaType.getTypeParameters(), getType(tree).getTypeArguments());
 
     if (info.isPresent()) {
       state.reportMatch(buildDescription(tree).setMessage(info.message()).build());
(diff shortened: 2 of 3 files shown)


schlosna:
I wanted to check in to see if there is anything I can help with to move this PR along. Thanks for error-prone, it is very useful!

cushon:
Thanks for the fix!

Does the included test case reproduce the crash for you? I'm having trouble seeing the failure when running that test on JDK 17 without the fix applied.

Adding the defensive null-checks and returning is definitely better than crashing, but I was curious if there's another problem here: e.g. maybe `ASTHelpers.targetType` should be updated to handle new-style switches, or there's an AST node that javac should be attaching type information to but isn't.

cushon:
> Does the included test case reproduce the crash for you?

Sorry, I just saw the CI failure you demonstrated in #3268, I'll take a closer look.

cushon:
> maybe ASTHelpers.targetType should be updated to handle new-style switches

I think this is the culprit: https://github.com/google/error-prone/blob/0da434efc8adca8682bea159c97f7fab65745c83/check_api/src/main/java/com/google/errorprone/util/ASTHelpers.java#L1764-L1765

What do you think about something like this?

```diff
diff --git a/check_api/src/main/java/com/google/errorprone/util/ASTHelpers.java b/check_api/src/main/java/com/google/errorprone/util/ASTHelpers.java
index d56da359a..0cdd625f7 100644
--- a/check_api/src/main/java/com/google/errorprone/util/ASTHelpers.java
+++ b/check_api/src/main/java/com/google/errorprone/util/ASTHelpers.java
@@ -1755,14 +1755,29 @@ public class ASTHelpers {
     @Nullable
     @Override
     public Type visitCase(CaseTree tree, Void unused) {
-      Tree t = parent.getParentPath().getLeaf();
-      // JDK 12+, t can be SwitchExpressionTree
-      if (t instanceof SwitchTree) {
-        SwitchTree switchTree = (SwitchTree) t;
-        return getType(switchTree.getExpression());
-      }
-      // TODO(b/176098078): When the ErrorProne project switches to JDK 12, we should check
-      // for SwitchExpressionTree.
+      Tree switchTree = parent.getParentPath().getLeaf();
+      return getType(getSwitchExpression(switchTree));
+    }
+
+    private static ExpressionTree getSwitchExpression(Tree tree) {
+      if (tree instanceof SwitchTree) {
+        return ((SwitchTree) tree).getExpression();
+      }
+      // Reflection is required for JDK < 12
+      try {
+        Class<?> switchExpression = Class.forName("com.sun.source.tree.SwitchExpressionTree");
+        Class<?> clazz = tree.getClass();
+        if (switchExpression.isAssignableFrom(clazz)) {
+          try {
+            Method method = clazz.getMethod("getExpression");
+            return (ExpressionTree) method.invoke(tree);
+          } catch (ReflectiveOperationException e) {
+            throw new LinkageError(e.getMessage(), e);
+          }
+        }
+      } catch (ClassNotFoundException e) {
+        // continue below
+      }
       return null;
     }
 
```

schlosna:
Hey @cushon thanks for the review and suggestion -- that looks correct to me and I added that commit to the PR.

schlosna:
Please let me know if there's anything else I can do to help get this merged. Thanks!

cushon:
Are the null-checks on the result of `getTarget` still necessary? The attached test case passes for me with just with change to `ASTHelpers`, without the other null checks.

schlosna:
@cushon I can revert the changes to ImmutableChecker, as I believe the ASTHelper changes fix the issue at hand. 

Would you prefer any asserts or just let things NPE for any future changes (e.g. if something like switch records pattern matching https://openjdk.org/jeps/405 were to  break current assumptions)?

schlosna:
Updated to revert 84aabb566f78505828dee93483273c1c2544784d

cushon:
> Would you prefer any asserts or just let things NPE for any future changes

I'm learning towards just letting it crash on unsupported features. It's a bit of a trade-off, and crashing is not great, but the alternative of defensive null-checks and `NO_MATCH` means it might just silently not check new features, which is potentially worse.

Anyway, thanks for the PR! I'll look at getting this merged.

cushon (approved):

schlosna:
Excellent, thanks!

PR #3301, opened 2022-06-29
Introduce command-line flag `-XepAllSuggestionsAsWarnings`

rickie:
This enables users that build with `-Werror` to disallow violations of custom
`SUGGESTION`-level bug patterns.

### DIFF
--- a/check_api/src/main/java/com/google/errorprone/ErrorProneOptions.java
+++ b/check_api/src/main/java/com/google/errorprone/ErrorProneOptions.java
@@ -52,8 +52,8 @@ public class ErrorProneOptions {
   private static final String PATCH_IMPORT_ORDER_PREFIX = "-XepPatchImportOrder:";
   private static final String EXCLUDED_PATHS_PREFIX = "-XepExcludedPaths:";
   private static final String IGNORE_LARGE_CODE_GENERATORS = "-XepIgnoreLargeCodeGenerators:";
-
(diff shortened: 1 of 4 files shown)


Stephan202 (approved):
We have a bunch of `SUGGESTION`-level bug checkers internally (to be open sourced "soon" :tm:), flagging things that aren't objectively warnings in all contexts. So this feature would be very helpful to us. :)

I verified against some internal code that the new flag works as intended.

hisener (approved):

Stephan202:
> We have a bunch of `SUGGESTION`-level bug checkers internally (to be open sourced "soon" tm)

For completeness: those bug checkers are now open sourced as part of [Error Prone Support](https://github.com/PicnicSupermarket/error-prone-support). We're already [using](https://github.com/PicnicSupermarket/error-prone-support/blob/e00aba12c33751304d07d8c8adc2b63d0fbff7ff/pom.xml#L1400) `-XepAllSuggestionsAsWarnings` as part of the project's self-check build, but it'd be nice if the self-check build did not rely on a fork of Error Prone. All that to say: it would be nice if this PR were accepted :)

rickie:
Rebased and resolved conflict.

rickie:
Rebased and resolved conflict. 

Would still be great if this could be added in Error Prone 😄.

cushon (approved):

Issue #3297, opened 2022-06-28
`UnusedMethod`: Add more JPA lifecycle annotations or make annotations configurable


jnehlmeier:
Currently `UnusedMethod` only ignores JPA lifecycle annotation `javax.persistence.PostLoad`.

However there are quite a few more:
- PostLoad
- PostPersist
- PostRemove
- PostUpdate
- PrePersist
- PreRemove
- PreUpdate

Also with the move to Jakarta namespace, all these annotations should be added with `javax.persistence` as well as `jakarta.persistence` package.

reitzmichnicht:
Now with spring boot 3 available this issue is a bit annoying. Any estimate if this will be included in the next error prone release?

Stephan202:
@rickie opened PR #3428 for this; I see that @cushon and @graememorgan worked on `UnusedMethod` before; maybe one of them can give it a push. :)

Agree this would be nice to see improved!

rickie:
The PR https://github.com/google/error-prone/pull/3428 is now merged, so we can close this ticket 😄.

PR #3428, opened 2022-09-02
UnusedMethod: Support additional exempting method annotations

### LINKED ISSUE #3144: `UnusedMethod` flags `@JsonValue` methods as unused
We have a couple of private static nested classes and enums with `@JsonValue` methods, mostly in tests. The `UnusedMethod` bug pattern flags those as unused, unlike the `@JsonCreator` methods.

#### JsonValueExample.java
```java
import java.util.Locale;
import com.fasterxml.jackson.annotation.JsonCreator;
import com.fasterxml.jackson.annotation.JsonValue;

public class JsonValueExample {
  private JsonValueExample() {}

  private enum FooBar {
    FOO,
    BAR;

    @JsonValue
    String value() {
      return name().toLowerCase(Locale.ROOT);
    }

    @JsonCreator
    static FooBar of(String value) {
      return valueOf(value.toUpperCase(Locale.ROOT));
    }
  }
}
```

```
curl -LO https://repo1.maven.org/maven2/com/google/errorprone/error_prone_core/2.13.1/error_prone_core-2.13.1-with-dependencies.jar
curl -LO https://repo1.maven.org/maven2/com/fasterxml/jackson/core/jackson-annotations/2.13.1/jackson-annotations-2.13.1.jar

javac \
  -J--add-exports=jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED \
  -J--add-exports=jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED \
  -J--add-exports=jdk.compiler/com.sun.tools.javac.main=ALL-UNNAMED \
  -J--add-exports=jdk.compiler/com.sun.tools.javac.model=ALL-UNNAMED \
  -J--add-exports=jdk.compiler/com.sun.tools.javac.parser=ALL-UNNAMED \
  -J--add-exports=jdk.compiler/com.sun.tools.javac.processing=ALL-UNNAMED \
  -J--add-exports=jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED \
  -J--add-exports=jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED \
  -J--add-opens=jdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED \
  -J--add-opens=jdk.compiler/com.sun.tools.javac.comp=ALL-UNNAMED \
  -cp jackson-annotations-2.13.1.jar \
  -XDcompilePolicy=simple \
  -processorpath error_prone_core-2.13.1-with-dependencies.jar \
  '-Xplugin:ErrorProne -XepDisableAllChecks -Xep:UnusedMethod:ERROR' \
  JsonValueExample.java
```

```
JsonValueExample.java:13: error: [UnusedMethod] Method 'value' is never used.
    String value() {
           ^
    (see https://errorprone.info/bugpattern/UnusedMethod)
  Did you mean 'BAR'?
1 error
```

A potential solution would be adding `com.fasterxml.jackson.annotation.JsonValue` to exempted annotations similar to `JsonCreator`.
https://github.com/google/error-prone/blob/f48903534c4897f80a7780c2564413307e6022fa/core/src/main/java/com/google/errorprone/bugpatterns/UnusedMethod.java#L97

Let me know if that makes sense; I can create a PR, as well.

rickie:
Fixes #3144 & #3297.


core/src/main/java/com/google/errorprone/bugpatterns/UnusedMethod.java
```diff
@@ -97,30 +98,68 @@ public final class UnusedMethod extends BugChecker implements CompilationUnitTre
...
           "org.aspectj.lang.annotation.Pointcut",
+          "org.aspectj.lang.annotation.After",
           "org.aspectj.lang.annotation.Before",
           "org.springframework.context.annotation.Bean",
           "org.testng.annotations.AfterClass",
           "org.testng.annotations.AfterMethod",
           "org.testng.annotations.BeforeClass",
           "org.testng.annotations.BeforeMethod",
           "org.testng.annotations.DataProvider",
+          "org.junit.After",
```
graememorgan, line 137: Could you explain why these junit methods need to be exempted? I'd expect that they would always be public, so wouldn't be flagged.

rickie: The reason for that is that `org.junit.AfterClass` and `org.junit.BeforeClass` were already in the list, so we wanted to make the list complete and consistent. 

_However_, based on the Javadoc and a local test, it _doesn't_ make sense to have them as they are indeed required to be `public`. 

Therefore we can omit the following annotations from the list:
- `org.junit.After`
- `org.junit.AfterClass`
- `org.junit.Before`
- `org.junit.BeforeClass`

Will push a commit to remove those from the list.

rickie: Rebased and added a commit @graememorgan.

graememorgan (approved):
Thanks!

PR #3437, opened 2022-09-08
Have `InvalidParam` support records

### LINKED ISSUE #2321: InvalidParam warning on Javadoc for Java record components

Javadoc on Java record components are `@param` tags on the record itself. It's also possible to override the accessor methods, but that shouldn't be necessary. error-prone produces an `InvalidParam` warning here though, as it doesn't seem to understand the Java record syntax.

```java
/**
 * Example record.
 * @param name Name of the example record.
 */
public record ExampleRecord(String name) {
}
```

`2.6.0`

rickie:
Fixes #2321.

### DIFF
--- a/check_api/src/main/java/com/google/errorprone/util/ASTHelpers.java
+++ b/check_api/src/main/java/com/google/errorprone/util/ASTHelpers.java
@@ -808,6 +808,29 @@ public class ASTHelpers {
   }
 
   /**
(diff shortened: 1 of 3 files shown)


rickie:
Rebased.

rickie:
Rebased 😄.

cushon (approved):

cushon:
Thanks for this!

PR #3504, opened 2022-10-23
Optimize `VisitorState#getSymbolFromName`

Stephan202:
By delegating module lookup to `Symtab#inferModule`.

### DIFF
--- a/check_api/src/main/java/com/google/errorprone/VisitorState.java
+++ b/check_api/src/main/java/com/google/errorprone/VisitorState.java
@@ -44,13 +44,13 @@ import com.sun.tools.javac.code.Type;
 import com.sun.tools.javac.code.Type.ArrayType;
 import com.sun.tools.javac.code.Type.ClassType;
 import com.sun.tools.javac.code.Types;
-import com.sun.tools.javac.comp.Modules;
 import com.sun.tools.javac.model.JavacElements;
 import com.sun.tools.javac.parser.Tokens.Token;
 import com.sun.tools.javac.tree.JCTree;
 import com.sun.tools.javac.tree.JCTree.JCCompilationUnit;
 import com.sun.tools.javac.tree.TreeMaker;
 import com.sun.tools.javac.util.Context;
+import com.sun.tools.javac.util.Convert;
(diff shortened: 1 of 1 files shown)


check_api/src/main/java/com/google/errorprone/VisitorState.java
```diff
@@ -406,20 +406,8 @@ public Name binaryNameFromClassname(String className) {
...
-      if (result != null) {
-        // TODO(cushon): the path where we iterate over all modules is probably slow.
-        // Try to learn some lessons from JDK-8189747, and consider disallowing this case and
-        // requiring users to call the getSymbolFromString(ModuleSymbol, Name) overload instead.
-        return result;
-      }
-    }
-    return null;
+    ModuleSymbol msym = getSymtab().inferModule(Convert.packagePart(name));
+    return msym != null ? getSymbolFromString(msym, name) : null;
```
Stephan202, line 410: [`Symtab#inferModule`](https://github.com/openjdk/jdk/blob/7cbf6721dcae6b0c22a57838d8dd769d87df9127/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Symtab.java#L852-L871) performs a map lookup rather than a linear scan. This approach is derived from [this JDK logic](https://github.com/openjdk/jdk/blob/7cbf6721dcae6b0c22a57838d8dd769d87df9127/src/jdk.javadoc/share/classes/jdk/javadoc/internal/tool/ToolEnvironment.java#L158-L171). With it, the build still passes, as does the [Error Prone Support](https://github.com/PicnicSupermarket/error-prone-support/) build. That said, neither uses custom modularization, so I can't fully rule out that this logic behaves subtly different.

Stephan202 (commented):
CC @cushon; this resolves a TODO with your name on it :)

cushon (approved):

Stephan202:
Rebased. @cushon I see that you approved; I take it that this PR is internally queued for review or something like that?

cushon:
@Stephan202 thanks for the ping, the import had gotten lost. I will restart it and get this into the next release.

PR #3646, opened 2022-12-31
StatementSwitchToExpressionSwitch: only trigger on compatible target versions

Stephan202:
This change introduces a `SourceVersion` utility class, analogous to the 
existing `RuntimeVersion` class. The new class tells whether selected language
features are available for the source version used by the compiler.

The new utility class is used to make sure that
`StatementSwitchToExpressionSwitch` flags statement switches only if the source
version supports expression switches, irrespective of whether the current
runtime supports such switch types. Concretely this makes the check compatible
with projects that target JDK 11, yet support building with JDK 17.


check_api/src/main/java/com/google/errorprone/util/TargetVersion.java
```diff
@@ -0,0 +1,88 @@
+/*
+ * Copyright 2023 The Error Prone Authors.
```
Stephan202, line 2: Seems safe to assume that this PR won't be merged before the fireworks ;)

pom.xml
```diff
@@ -161,10 +161,10 @@
           <compilerArgs>
             <arg>--add-exports=java.base/jdk.internal.javac=ALL-UNNAMED</arg>
             <arg>--add-exports=jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED</arg>
-            <arg>--add-exports=jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED</arg>
             <arg>--add-exports=jdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED</arg>
-            <arg>--add-exports=jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED</arg>
             <arg>--add-exports=jdk.compiler/com.sun.tools.javac.comp=ALL-UNNAMED</arg>
+            <arg>--add-exports=jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED</arg>
+            <arg>--add-exports=jdk.compiler/com.sun.tools.javac.jvm=ALL-UNNAMED</arg>
             <arg>--add-exports=jdk.compiler/com.sun.tools.javac.main=ALL-UNNAMED</arg>
```
Stephan202, line 168: Summary of changes: 
- Sorted.
- Removed duplicate `com.sun.tools.javac.util` entry.
- Added `com.sun.tools.javac.jvm`.

cushon:
Thanks, SGTM overall. 

I appreciate making `TargetVersion` match the style of the existing `RuntimeVersion`, but in hindsight I'm not sure all of the `isAtLeastN` overloads was a great API, and for this example it might be nice to express it in terms of the feature we want to detect instead of the corresponding version number.

javac has a built-in API for detecting if features are available at the current source version, what do you think about using that instead, e.g.? `Feature.SWITCH_EXPRESSION.allowedInSource(Source.instance(context))`

That's using the source version instead of the target version, which wouldn't enable the check for configurations like `-source 8 -target 19`, but I think that's fine.

cushon (changes requested):
[(See the comment on theP](https://github.com/google/error-prone/pull/3646#issuecomment-1368225237)

pom.xml
```diff
@@ -213,6 +213,7 @@
             -Xmx1g
             --add-exports=jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED
             --add-exports=jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED
+            --add-exports=jdk.compiler/com.sun.tools.javac.jvm=ALL-UNNAMED
```
Stephan202, line 216: If this PR is accepted, then the [documentation](https://errorprone.info/docs/installation) should also be updated to include this flag.

Stephan202 (commented):
Added some notes.

Stephan202:
> what do you think about using that instead, e.g.?

Interesting idea! Let me have a look a this and report back here :+1:

check_api/src/main/java/com/google/errorprone/util/SourceVersion.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.Arrays;
+
+/**
+ * JDK source version utilities.
+ *
+ * @see RuntimeVersion
+ */
+public final class SourceVersion {
+  private static final ImmutableMap<String, Feature> KNOWN_FEATURES =
+      Maps.uniqueIndex(Arrays.asList(Feature.values()), Enum::name);
```
Stephan202, line 33: The string-based lookup is necessary because older runtimes won't recognize newer enum values. A bit fragile, but no more fragile than what's usual for Error Prone.

check_api/src/main/java/com/google/errorprone/util/SourceVersion.java
```diff
@@ -0,0 +1,58 @@
...
+import com.sun.tools.javac.code.Source.Feature;
+import com.sun.tools.javac.util.Context;
+import java.util.Arrays;
+
+/**
+ * JDK source version utilities.
+ *
+ * @see RuntimeVersion
+ */
+public final class SourceVersion {
```
Stephan202, line 31: Could also name the class something like `SourceFeature(s)`, though that might perhaps also call for slightly different method names.  :thinking:

core/src/main/java/com/google/errorprone/bugpatterns/StatementSwitchToExpressionSwitch.java
```diff
@@ -85,8 +86,7 @@ public StatementSwitchToExpressionSwitch(ErrorProneFlags flags) {
 
   @Override
   public Description matchSwitch(SwitchTree switchTree, VisitorState state) {
-    // Expression switches finalized in Java 14
-    if (!RuntimeVersion.isAtLeast14()) {
+    if (!SourceVersion.supportsSwitchExpressions(state.context)) {
```
Stephan202, line 89: Dropped the comment because the code is now self-documenting.

Stephan202 (commented):
Will check whether this can be nicely tested without too much hassle.

check_api/src/main/java/com/google/errorprone/util/SourceVersion.java
```diff
@@ -0,0 +1,58 @@
...
+
+  /** Returns true if the compiler source version level supports switch expressions. */
+  public static boolean supportsSwitchExpressions(Context context) {
+    return supportsFeature("SWITCH_EXPRESSION", context);
+  }
+
+  /** Returns true if the compiler source version level supports text blocks. */
+  public static boolean supportsTextBlocks(Context context) {
+    return supportsFeature("TEXT_BLOCKS", context);
+  }
```
Stephan202, line 43: The reason for adding this method is twofold:
1. With two examples it should be clear for future devs how other methods may be added.
2. I have a use case for this particular feature in PicnicSupermarket/error-prone-support#198.

Stephan202 (commented):
Added some unit tests.

check_api/src/main/java/com/google/errorprone/util/SourceVersion.java
cushon: Thanks, I didn't consider that when making the suggestion. This looks like a good way to deal with that.

One more suggestion would be to keep the current API, but avoid relying on the `Feature` enum, e.g.:

```java
public static boolean supportsSwitchExpressions(Context context) {
  return Source.instance(state.context).compareTo(Source.lookup("14")) >= 0;
}
```

That means we can't use javac's knowledge about which features were used in which versions, but it also avoids depending on the string names of the `Feature` constants, and they're an internal API that could change (e.g. constants will probably be removed as javac drops support for old version features).

check_api/src/main/java/com/google/errorprone/util/SourceVersion.java
cushon: 👍 

There is at least one other use of `Source` in Error Prone that can also be cleaned up to use this approach. (That's just an FYI, it doesn't need to happen in this PR.)

https://cs.github.com/google/error-prone/blob/721096fd16b9f9af7ae7dd657bd9d4ae75db5b46/core/src/main/java/com/google/errorprone/bugpatterns/VarChecker.java?q=repo%3Agoogle%2Ferror-prone++Source.instance#L102

Stephan202 (commented):
Applied :)

cushon (approved):

Stephan202:
(Updated the PR description to be in line with the latest revision.)

Tnx for the rapid review iterations @cushon!

cushon (approved):

PR #3654, opened 2023-01-02
ClassCanBeStatic: Exclude JUnit `@Nested` classes

### LINKED ISSUE #956: Exclude inner classes annotated with @Nested from ClassCanBeStatic rule
Please answer these questions before submitting your issue. Thanks!

2.2.0 (with maven.plugin 2.8.2)

### Does this issue reproduce with the latest release?

Yes

### What did you do?

JUnit 5 provides nested tests using an inner class (not static): See Example at  https://junit.org/junit5/docs/current/user-guide/#writing-tests-nested 

If I use the rule **ClassCanBeStatic** on all my classes, then all those tests are flagged. 

### What did you expect to see?

An inner class with the annotation `@org.junit.jupiter.api.Nested` should not be flagged as warning.

### What did you see instead?

[WARNING] /Users/hafner/Development/jenkins/workspace/Pipeline/src/test/java/edu/hm/hafner/analysis/parser/dry/CodeDuplicationTest.java:[64,5] [ClassCanBeStatic] Inner class is non-static but does not reference enclosing class

ljrmorgan:
Exclude inner classes annotated with JUnit's `@Nested` annotation from the `ClassCanBeStatic` check. JUnit requires them to not be `static`.

From the [JUnit docs](https://junit.org/junit5/docs/current/user-guide/#writing-tests-nested):

> Only non-static nested classes (i.e. inner classes) can serve as `@Nested` test classes.

Fixes #956.

### DIFF
--- a/check_api/src/main/java/com/google/errorprone/util/ASTHelpers.java
+++ b/check_api/src/main/java/com/google/errorprone/util/ASTHelpers.java
@@ -1041,7 +1041,9 @@ public class ASTHelpers {
    * though they are.
    */
   private static final ImmutableSet<String> ANNOTATIONS_CONSIDERED_KEEP =
-      ImmutableSet.of("org.apache.beam.sdk.transforms.DoFn.ProcessElement");
+      ImmutableSet.of(
+          "org.apache.beam.sdk.transforms.DoFn.ProcessElement",
+          "org.junit.jupiter.api.Nested");
 
   private static final String USED_REFLECTIVELY = "UsedReflectively";
 
--- a/core/src/main/java/com/google/errorprone/bugpatterns/ClassCanBeStatic.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/ClassCanBeStatic.java
@@ -43,6 +43,7 @@ import javax.lang.model.element.NestingKind;
     tags = {StandardTags.STYLE, StandardTags.PERFORMANCE})
 public class ClassCanBeStatic extends BugChecker implements ClassTreeMatcher {
 
+  private static final String JUNIT_NESTED_ANNOTATION = "org.junit.jupiter.api.Nested";
   private static final String REFASTER_ANNOTATION =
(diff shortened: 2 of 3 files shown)


Stephan202:
Nice one. In our code base I find quite a number of these:
```java
@Nested
@SuppressWarnings("ClassCanBeStatic" /* @Nested classes cannot be static. */)
```

cushon:
Thanks, LGTM overall.

We've been starting to consolidate more of the heuristics for situations where Error Prone shouldn't change modifiers etc.  by using `@Keep`, or treating annotations as if they were annotated with `@Keep`.

https://github.com/google/error-prone/blob/bb9ede9f250660199a718b36407cffec36b2be46/annotations/src/main/java/com/google/errorprone/annotations/Keep.java#L30-L31

What do you think about using `ASTHelpers.shoudlKeep` here, and adding `@Nested` to this list: https://github.com/google/error-prone/blob/bb9ede9f250660199a718b36407cffec36b2be46/check_api/src/main/java/com/google/errorprone/util/ASTHelpers.java#L1043-L1044

ljrmorgan:
@cushon thank you for taking a look at the PR, and for the suggestion!

I think that makes sense - the JavaDoc for `@Keep` made me think initially of checks like `UnusedMethod` etc., but I suppose the only reason that `@Nested` classes can't be static is because of how they are used via reflection, which fits the description of `@Keep`. I guess my only concern is that we might lose some valid `ClassCanBeStatic` warnings on types that happen to be annotated with `@Keep`, but you are much better placed than I am to say whether that is a legitimate concern 😁 

I've pushed that change as a separate commit, I'm happy to squash the two commits into one if you'd prefer, just let me know.

Thanks!

ljrmorgan:
Hi @cushon, I wondered if you'd had a chance to take a look at the latest changes? Thanks!

cushon (approved):

PR #3645, opened 2022-12-31
StatementSwitchToExpressionSwitch: handle empty statement blocks

Stephan202:
Fixes #3638.

PR #3656, opened 2023-01-02
StatementSwitchToExpressionSwitch: skip rule cases

Stephan202:
The cases in a switch statement must all be of kind `STATEMENT` or
`RULE`. In the latter case `StatementSwitchToExpressionSwitch` does not
apply.

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/StatementSwitchToExpressionSwitch.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/StatementSwitchToExpressionSwitch.java
@@ -86,7 +86,7 @@ public final class StatementSwitchToExpressionSwitch extends BugChecker
 
   @Override
   public Description matchSwitch(SwitchTree switchTree, VisitorState state) {
-    if (!SourceVersion.supportsSwitchExpressions(state.context)) {
+    if (!SourceVersion.supportsSwitchExpressions(state.context) || isRuleSwitch(switchTree)) {
       return NO_MATCH;
     }
 
@@ -97,6 +97,18 @@ public final class StatementSwitchToExpressionSwitch extends BugChecker
     }
 
     return NO_MATCH;
+  }
+
+  private static boolean isRuleSwitch(SwitchTree switchTree) {
+    return switchTree.getCases().stream().anyMatch(StatementSwitchToExpressionSwitch::isRuleCase);
+  }
+
+  private static boolean isRuleCase(CaseTree caseTree) {
+    try {
+      return "RULE".equals(CaseTree.class.getMethod("getCaseKind").invoke(caseTree).toString());
+    } catch (ReflectiveOperationException e) {
+      throw new LinkageError(e.getMessage(), e);
+    }
   }
 
   /**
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/google/errorprone/bugpatterns/StatementSwitchToExpressionSwitch.java
```diff
@@ -122,7 +134,7 @@ private static AnalysisResult analyzeSwitchTree(SwitchTree switchTree) {
 
       List<? extends StatementTree> statements = caseTree.getStatements();
       CaseFallThru caseFallThru = CaseFallThru.MAYBE_FALLS_THRU;
-      if (statements == null || statements.isEmpty()) {
+      if (statements.isEmpty()) {
```
Stephan202, line 137: This reverts the change in #3645, because there are actually many more `CaseTree#getStatements()` calls in this class, and they _all_ assume a  non-`null` result. The `isRuleSwitch` check now guarantees that.

core/src/main/java/com/google/errorprone/bugpatterns/StatementSwitchToExpressionSwitch.java
```diff
@@ -99,6 +99,18 @@ public Description matchSwitch(SwitchTree switchTree, VisitorState state) {
...
+    return switchTree.getCases().stream().anyMatch(StatementSwitchToExpressionSwitch::isRuleCase);
+  }
+
+  private static boolean isRuleCase(CaseTree caseTree) {
+    try {
+      return "RULE".equals(CaseTree.class.getMethod("getCaseKind").invoke(caseTree).toString());
+    } catch (ReflectiveOperationException e) {
+      throw new LinkageError(e.getMessage(), e);
+    }
+  }
```
Stephan202, line 112: Given the [documented](https://github.com/openjdk/jdk/blob/67086ebf80109a623f3c2ad24d4e1a65de43d986/src/jdk.compiler/share/classes/com/sun/source/tree/CaseTree.java#L80-L87) invariant on which we now rely, perhaps it's more accurate to test for `"STATEMENT"`. But it seems unlikely that a third `Kind` will be introduced any time soon, and this way we avoid a negation at the call site. But I'm happy to switch this around.

core/src/test/java/com/google/errorprone/bugpatterns/StatementSwitchToExpressionSwitchTest.java
```diff
@@ -1030,6 +1030,25 @@ public void emptySwitchCases_noMatch() {
             "    }",
             "  }",
             "}")
+        .setArgs("-XepOpt:StatementSwitchToExpressionSwitch:EnableDirectConversion")
```
Stephan202, line 1033: Added this for symmetry with other tests, and because in theory it covers more code paths (except that in practice `isRuleSwitch(switchTree)` is tested before the flag is consulted).

NB: the other tests use `.setArgs(ImmutableList.of(...))`; I wouldn't mind opening a separate cleanup PR for that.

Stephan202 (commented):
Added some notes. This PR avoids the NPE mentioned [here](https://github.com/google/error-prone/pull/3645#issuecomment-1368693207).

Stephan202:
Rebased and resolved conflict.

One way to look at this PR is that without it, it is unlikely that JDK 14+ users can enable `StatementSwitchToExpressionSwitch` by default in order to prevent fresh statement-style `switch` expressions from being introduced in their code. (Because without this change, any already-migrated rule-style `switch` expression is likely to trigger an NPE.)

Stephan202:
The solution in #3676 doesn't require reflection; will close this PR.

Issue #1650, opened 2020-05-31
@SuppressWarnings("InlineFormatString") doesn't work


anthonyvdotbe:
### Description of the problem / feature request:

I'm unable to suppress `InlineFormatString`, neither by annotating the constant, nor by annotating the method that uses the constant.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

````java
class Foo {

    @SuppressWarnings("InlineFormatString")
    private static final String FMT = "%s";

    @SuppressWarnings("InlineFormatString")
    String foo() {
        return String.format(FMT, "bar");
    }

}
````

### What version of Error Prone are you using?

2.4.0

###  Have you found anything relevant by searching the web?

No.

cushon:
This was an oversight, I agree `@SuppressWarnings("InlineFormatString")` should work if the constant or an enclosing scope is annotated.

Out of curiosity, was there a particular real-world example where this suggestion wasn't useful? If you disagree with it in general, you can also disable the check with `-Xep:InlineFormatString:OFF`.

anthonyvdotbe:
In this case, I had relatively large text blocks (think: templates) which contained just 1/2 arguments, and the format strings were used like:
````java
builder
.with(Foo.class, o -> String.format(FMT_FOO, v.apply(o.content().text())))
.with(Bar.class, (v, o) -> String.format(FMT_BAR, v.apply(o.content())))
...
.with(Baz.class, (v, o) -> String.format(FMT_BAZ, v.apply(o.content())))
````

So inlining the format Strings would make the code look messy, and I felt that refactoring the code to use separate methods wasn't worth the effort. I don't disagree with it in general, though.

PR #3690, opened 2023-01-05
Make MemoizeConstantVisitorStateLookups check suppressible

msridhar:
It's nice to be able to suppress this check, e.g., while you're initially prototyping a checker.  I based my fix on the fix for #1650; hope it's the right approach.

### DIFF
--- a/core/src/test/java/com/google/errorprone/bugpatterns/MemoizeConstantVisitorStateLookupsTest.java
+++ b/core/src/test/java/com/google/errorprone/bugpatterns/MemoizeConstantVisitorStateLookupsTest.java
@@ -164,7 +164,7 @@ public class MemoizeConstantVisitorStateLookupsTest {
   }
 
   @Test
-  public void testSuppressWarnings() {
+  public void respectSuppressWarnings() {
     compilationTestHelper
         .addSourceLines(
             "Test.java",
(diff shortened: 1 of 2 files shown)


core/src/test/java/com/google/errorprone/bugpatterns/MemoizeConstantVisitorStateLookupsTest.java
```diff
@@ -162,4 +162,31 @@ public void negative_doesntMemoizeTwice() {
         .expectUnchanged()
         .doTest();
   }
+
+  @Test
+  public void testSuppressWarnings() {
```
rickie, line 167:
```suggestion
  public void suppressWarnings() {
```
It looks like we should drop the `test` prefix here, based on https://github.com/google/error-prone/commit/dbd04752b51372998f7fbd75d1e88ef731cefffe.

(As pointed out in https://github.com/google/error-prone/pull/3677#issuecomment-1370677956, we have a check for this in Error Prone Support which we are open to contributing upstream 😄, see related [documentation](https://error-prone.picnic.tech/bugpatterns/JUnitMethodDeclaration/))

rickie (commented):
Small drive-by review 😄.

core/src/test/java/com/google/errorprone/bugpatterns/MemoizeConstantVisitorStateLookupsTest.java
msridhar: @rickie thanks!  I tried to resolve this, but the Google system to pull in the PR may have already kicked in...we'll see what actually lands :-)

PR #3737, opened 2023-01-27
Handle `yield` statement case in `ASTHelpers#targetType`

hisener:
https://github.com/google/error-prone/commit/7504736304148615ab53fe1479cc18027d822ff7 fixes the switch expression case, but `ImmutableChecker` can still throw `NullPointerException` for the `yield` statement + method reference case. This PR handles that case, too.

### DIFF
--- a/check_api/src/main/java/com/google/errorprone/util/ASTHelpers.java
+++ b/check_api/src/main/java/com/google/errorprone/util/ASTHelpers.java
@@ -1724,13 +1724,15 @@ public class ASTHelpers {
 
     Type type = new TargetTypeVisitor(current, state, parent).visit(parent.getLeaf(), null);
     if (type == null) {
-      if (CONSTANT_CASE_LABEL_TREE != null
+      Tree actualTree = null;
+      if (YIELD_TREE != null && YIELD_TREE.isAssignableFrom(parent.getLeaf().getClass())) {
+        actualTree = parent.getParentPath().getParentPath().getParentPath().getLeaf();
+      } else if (CONSTANT_CASE_LABEL_TREE != null
           && CONSTANT_CASE_LABEL_TREE.isAssignableFrom(parent.getLeaf().getClass())) {
-        type =
-            getType(
-                TargetTypeVisitor.getSwitchExpression(
-                    parent.getParentPath().getParentPath().getLeaf()));
+        actualTree = parent.getParentPath().getParentPath().getLeaf();
       }
+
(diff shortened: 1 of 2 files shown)


core/src/test/java/com/google/errorprone/bugpatterns/threadsafety/ImmutableCheckerTest.java
```diff
@@ -2993,4 +2993,62 @@ public void switchExpressionsResultingInGenericTypes_doesNotThrow() {
...
+            "class Test {",
+            "  Supplier<Double> test(String mode) {",
+            "    return switch (mode) {",
+            "      case \"random\" -> Math::random;",
+            "      default -> throw new IllegalArgumentException();",
+            "    };",
+            "  }",
+            "}")
+        .doTest();
+  }
```
hisener, line 3033: These two cases are irrelevant for this change, but I added to make sure NPE is only related to `yield AClass::methodReference`.

check_api/src/main/java/com/google/errorprone/util/ASTHelpers.java
```diff
@@ -1724,13 +1724,15 @@ public static TargetType targetType(VisitorState state) {
 
     Type type = new TargetTypeVisitor(current, state, parent).visit(parent.getLeaf(), null);
     if (type == null) {
-      if (CONSTANT_CASE_LABEL_TREE != null
+      Tree actualTree = null;
+      if (YIELD_TREE != null && YIELD_TREE.isAssignableFrom(parent.getLeaf().getClass())) {
+        actualTree = parent.getParentPath().getParentPath().getParentPath().getLeaf();
+      } else if (CONSTANT_CASE_LABEL_TREE != null
           && CONSTANT_CASE_LABEL_TREE.isAssignableFrom(parent.getLeaf().getClass())) {
```
hisener, line 1731: There should be a better way of handling these. 😬

cushon: Agree, but I think reflection is the best option we have right now.

I've been wondering if we should try to use mrjars to avoid some of the reflection, I filed an issue to track that idea: https://github.com/google/error-prone/issues/3756

cushon (approved):
Thanks for the fix!

check_api/src/main/java/com/google/errorprone/util/ASTHelpers.java
hisener: That's a good idea! Thanks for filing the issue. 🙇

Issue #3774, opened 2023-02-09
InvalidThrows doesn't work


delanym:
This rule simply doesn't work. EP v2.16
I tried it by adding an IOException and a custom checked exception.

Stephan202:
@delanym the latest release is 2.18.0. For me the rule does work, and there are [tests](https://github.com/google/error-prone/blob/b80fcaf00a7bd0702f78a536b84ac0fa1407d657/core/src/test/java/com/google/errorprone/bugpatterns/javadoc/InvalidThrowsTest.java). Are you looking for [CheckedExceptionNotThrown](https://errorprone.info/bugpattern/CheckedExceptionNotThrown) instead?

If not, please provide a [minimal reproducible example](https://en.wikipedia.org/wiki/Minimal_reproducible_example).

delanym:
@Stephan202 ok I tried to update to 2.17.0 or 2.18.0 but now for some reason the **ImmutableEnumChecker** rule is being activated (I didn't ask for it) and its failing

delanym:
Has some other rule changed to depend on this rule perhaps?

Stephan202:
@delanym asking this kind of open-ended question without providing details is not helpful. You'll get better support if you spend time providing a minimal reproducible example.

delanym:
@tbroyer Having a bad day?

I see what's going on. I had the "Immutable" rule set using error-prone 2.16, which only became a valid alias for "ImmutableEnumChecker" in 2.17.
Since invalid pattern names cause the build to fail, the Immutable pattern must be present but inactive in 2.16. That's my guess anyway.

I've had similar issues in the past https://github.com/google/error-prone/issues/976
The https://errorprone.info/bugpatterns page makes no attempt to differentiate versions so no surprises there. There should at least be a version number on the page. Thanks for your efforts!

@Stephan202 yes it is CheckedExceptionNotThrown rather than InvalidThrows. I have that set as well, but its not working on my large closed source project. My simplistic attempt at a reproducer did not prove anything, as I expected.

Its at times like these when I'm really just giving you a hint that there's a potential problem with the rule, and if you, the maintainer, the next person searching through issues — you get the sense that something might be wrong and you can be extra vigilant around that code. I simply don't have time to create complex reproducers for every issue I find, but ill reply to this ticket if the situation changes.

delanym:
I see there is a separate checker called https://errorprone.info/bugpattern/Immutable

Please sort out the docs

[...]

PR #3782, opened 2023-02-12
Improve clarity of CheckedExceptionNotThrown description

delanym:
Improve clarity of CheckedExceptionNotThrown description re: #3774

### DIFF
--- a/check_api/src/main/java/com/google/errorprone/suppliers/Suppliers.java
+++ b/check_api/src/main/java/com/google/errorprone/suppliers/Suppliers.java
@@ -181,6 +181,14 @@ public final class Suppliers {
         }
       };
 
+  public static final Supplier<Type> SHORT_TYPE =
+      new Supplier<Type>() {
+        @Override
+        public Type get(VisitorState state) {
+          return state.getSymtab().shortType;
+        }
+      };
+
   public static final Supplier<Type> INT_TYPE =
       new Supplier<Type>() {
         @Override
--- a/core/src/main/java/com/google/errorprone/bugpatterns/ComparisonOutOfRange.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/ComparisonOutOfRange.java
@@ -19,9 +19,13 @@ package com.google.errorprone.bugpatterns;
 import static com.google.errorprone.BugPattern.SeverityLevel.ERROR;
 import static com.google.errorprone.fixes.SuggestedFix.replace;
 import static com.google.errorprone.matchers.Description.NO_MATCH;
+import static com.google.errorprone.matchers.Matchers.anyOf;
 import static com.google.errorprone.matchers.Matchers.isSameType;
 import static com.google.errorprone.suppliers.Suppliers.BYTE_TYPE;
 import static com.google.errorprone.suppliers.Suppliers.CHAR_TYPE;
+import static com.google.errorprone.suppliers.Suppliers.INT_TYPE;
+import static com.google.errorprone.suppliers.Suppliers.LONG_TYPE;
+import static com.google.errorprone.suppliers.Suppliers.SHORT_TYPE;
 import static com.google.errorprone.util.ASTHelpers.constValue;
 import static com.google.errorprone.util.ASTHelpers.isSameType;
 import static com.google.errorprone.util.ASTHelpers.matchBinaryTree;
@@ -32,6 +36,7 @@ import static java.lang.String.format;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Range;
 import com.google.errorprone.BugPattern;
+import com.google.errorprone.ErrorProneFlags;
 import com.google.errorprone.VisitorState;
 import com.google.errorprone.bugpatterns.BugChecker.BinaryTreeMatcher;
 import com.google.errorprone.matchers.Description;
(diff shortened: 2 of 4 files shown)


core/src/main/java/com/google/errorprone/bugpatterns/CheckedExceptionNotThrown.java
```diff
@@ -64,8 +64,8 @@
 /** Flags checked exceptions which are claimed to be thrown, but are not. */
 @BugPattern(
     summary =
-        "This method cannot throw a checked exception that it claims to. This may cause consumers"
-            + " of the API to incorrectly attempt to handle, or propagate, this exception.",
+        "This method is overridable and cannot throw any of the checked exceptions it claims to. This may cause consumers"
```
tbroyer, line 67: Shouldn't that say "final" rather than "overridable"?

Also, about the "any of the checked exceptions", is that really the current behavior? Shouldn't that be "one of the checked exceptions"? (in which case, the original phrasing was clearer; what I find unclear, as a non-native English speaker, is the "cannot", "will not" or "does not", or "never throws" would be clearer)

docs/bugpattern/CheckedExceptionNotThrown.md
```diff
@@ -18,6 +16,8 @@ Response handle(Request request) {
...
 ```
 
-Including unthrown exceptions can be reasonable where the method is overridable,
-as overriding methods will not be able to declare that they throw any exceptions
-not included.
+Unfortunately this can lead to call sites being forced to explicitly handle or
+propagate exceptions which provably can never occur. It may also lead readers of
+the code to search for possibly throwing paths where none exist.
+
+The rule checks only whether the method is overridable, not whether it is overridden.
```
tbroyer, line 23: It would probably be enough, IMO, to add to the previous wording: "; this check will thus only flag non-overridable methods." (or "not flag overridable methods")

tbroyer (changes requested):

core/src/main/java/com/google/errorprone/bugpatterns/CheckedExceptionNotThrown.java
delanym: Actually it should say "non-overridable". Final methods are non-overridable AND they also can't be hidden. Static methods are non-overridable but can be hidden. Private makes the method un-overridable through being inaccessible. I confirmed the check considers all these cases.

Yes you're right the original is clearer, but even clearer is to avoid the indefinite article, so "one or more".

Even as native English speaker we often miss points like this. To say "cannot" might suggest the method is trying to "throw new IOException" or the like. But this is about how the method throws/propagates unhandled exceptions. The method "will never" be required to throw the exception. "does not" is something to consider at runtime.

docs/bugpattern/CheckedExceptionNotThrown.md
delanym: I wanted to answer the question why Java allows redundant throws in the opening statement and not leave it for later. Currently the two statements stand in isolation, yet this is the crux of how the issue comes up in the first place.

As I look at the example now I see it showing the undesirable propagation but not the main issue: checkArgument doesn't throw IOException. I'll try come up with a better example

delanym: @tbroyer do you understand this "It may also lead readers of the code to search for possibly throwing paths where none exist." ?

docs/bugpattern/CheckedExceptionNotThrown.md
```diff
@@ -1,11 +1,11 @@
...
-don't. This can lead to call sites being forced to explicitly handle or
+Java allows methods to declare that they throw checked exceptions even when they don’t.
+This can lead to call sites being forced to explicitly handle or
 propagate exceptions which provably can never occur. It may also lead readers of
 the code to search for possibly throwing paths where none exist.
 
 ```java
 private static void validateRequest(Request request) throws IOException {
-  checkArgument(request.hasFoo(), "foo must be specified");
+  System.out.println("Request " + request.name + " validated");
```
tbroyer, line 8: That doesn't really look like a "validateRequest" implementation. Maybe replace checkArgument with the equivalent non-Guava code instead if you think it's unclear?
```suggestion
  if (!request.hasFoo()) {
    throw new IllegalArgumentException("foo must be specified");
  }
```

docs/bugpattern/CheckedExceptionNotThrown.md
```diff
@@ -1,11 +1,11 @@
-Java allows methods to declare that they throw checked exceptions even when they
-don't. This can lead to call sites being forced to explicitly handle or
+Java allows methods to declare that they throw checked exceptions even when they don’t.
+This can lead to call sites being forced to explicitly handle or
```
tbroyer, line 2: nit: unnecessary change

docs/bugpattern/CheckedExceptionNotThrown.md
```diff
@@ -18,6 +18,4 @@ Response handle(Request request) {
 }
 ```
 
-Including unthrown exceptions can be reasonable where the method is overridable,
-as overriding methods will not be able to declare that they throw any exceptions
-not included.
+The check will not flag overridable methods.
```
tbroyer, line 21: I think the previous text was clearer in that it described **why** that would be the case ("it can be reasonable"). I'd rather **add** to the previous text than replace it.
```suggestion
Including unthrown exceptions can be reasonable where the method is overridable,
as overriding methods will not be able to declare that they throw any exceptions
not included. The check will therefore not flag overridable methods.
```

delanym: Moved up and parenthesized. The last sentence describes the scope of the check, as a kind of disclaimer. Once the check is adapted to inspect overridable methods it might say "The check will flag non-overridable methods and methods that declare throwing a checked exception where none of the overridden methods throw that exception."

Issue #3759, opened 2023-02-06
Internal usage of guava and caffeine `Cache` seems inconsistent


XN137:
In https://github.com/google/error-prone/commit/e1acde927336e52bfcde8343979feff7ed84630a caffeine was added to the build:
```
This adds a new dependency on Caffeine; I considered using a Guava
Cache since we already depend on Guava, but Caffeine's version is 3
times as fast for this use case.
```

however on `master` a guava cache is also used in a few places:
https://github.com/google/error-prone/blob/1c3c09f5c286989f76eb545613c9ce91dac4acad/check_api/src/main/java/com/google/errorprone/dataflow/nullnesspropagation/inference/NullnessQualifierInference.java#L22-L24
https://github.com/google/error-prone/blob/1c3c09f5c286989f76eb545613c9ce91dac4acad/check_api/src/main/java/com/google/errorprone/dataflow/DataFlow.java#L21-L23

so i was wondering:
is this intentional or would there be value in unifying this?

- using caffeine could (allegedly) give a performance boost
- dropping caffeine would remove the extra dependency

cushon:
We are slowly moving towards caffeine, PRs to migrate those usages of `common.cache` are welcome

Sineaggi:
@cushon I've god a pr that should fix existing usages of `common.cache` here https://github.com/google/error-prone/pull/3796

PR #3796, opened 2023-02-25
Replace guava cache with caffeine

Sineaggi:
Resolves https://github.com/google/error-prone/issues/3759

### DIFF
--- a/check_api/src/main/java/com/google/errorprone/dataflow/nullnesspropagation/inference/NullnessQualifierInference.java
+++ b/check_api/src/main/java/com/google/errorprone/dataflow/nullnesspropagation/inference/NullnessQualifierInference.java
@@ -18,16 +18,15 @@ package com.google.errorprone.dataflow.nullnesspropagation.inference;
 
 import static com.google.common.base.Preconditions.checkArgument;
 
+import com.github.benmanes.caffeine.cache.CacheLoader;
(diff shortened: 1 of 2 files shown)


cushon (approved):

check_api/src/main/java/com/google/errorprone/dataflow/nullnesspropagation/inference/NullnessQualifierInference.java
```diff
@@ -100,8 +100,8 @@ public static InferredNullability getInferredNullability(Tree methodOrInitialize
         "Tree `%s` is not a lambda, initializer, or method.",
         methodOrInitializerOrLambda);
     try {
-      return inferenceCache.getUnchecked(methodOrInitializerOrLambda);
-    } catch (UncheckedExecutionException e) {
+      return inferenceCache.get(methodOrInitializerOrLambda);
+    } catch (CompletionException e) {
```
ben-manes, line 104:
```java
@throws CompletionException if a checked exception was thrown while loading the value
```
[CompletionFailure](https://javadoc.io/static/io.earcam.wrapped/jdk.compiler/1.8.132/com/sun/tools/javac/code/Symbol.CompletionFailure.html) is a runtime exception, so I think that you can catch it directly?

PR #3797, opened 2023-02-26
Fix bug where nested `MissingBraces` violations' suggested fixes result in broken code

keithl-stripe:
@holtherndon-stripe came across this issue while applying `MissingBraces` to Stripe's code base of 4 million lines of Java code. 

See unit test for sample code.

I'm not sure about a few things:
- Why does the coalescing code suppress duplicate inserts? It seems a bit odd that this is the default. I wonder which checkers' fixes need this, and why.
- Is `Fix` the right level of abstraction? Or should the coalescing policy be per `Replacement`?

### DIFF
--- a/check_api/src/main/java/com/google/errorprone/apply/DescriptionBasedDiff.java
+++ b/check_api/src/main/java/com/google/errorprone/apply/DescriptionBasedDiff.java
@@ -99,7 +99,7 @@ public final class DescriptionBasedDiff implements DescriptionListener, Diff {
     importsToRemove.addAll(fix.getImportsToRemove());
     for (Replacement replacement : fix.getReplacements(endPositions)) {
       try {
-        replacements.add(replacement, Replacements.CoalescePolicy.EXISTING_FIRST);
+        replacements.add(replacement, fix.getCoalescePolicy());
       } catch (IllegalArgumentException iae) {
         if (!ignoreOverlappingFixes) {
           throw iae;
--- a/check_api/src/main/java/com/google/errorprone/fixes/Fix.java
+++ b/check_api/src/main/java/com/google/errorprone/fixes/Fix.java
@@ -40,6 +40,8 @@ public interface Fix {
     return "";
   }
 
+  Replacements.CoalescePolicy getCoalescePolicy();
+
   Set<Replacement> getReplacements(EndPosTable endPositions);
 
   Collection<String> getImportsToAdd();
--- a/check_api/src/main/java/com/google/errorprone/fixes/SuggestedFix.java
+++ b/check_api/src/main/java/com/google/errorprone/fixes/SuggestedFix.java
@@ -23,6 +23,7 @@ import com.google.auto.value.extension.memoized.Memoized;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
+import com.google.errorprone.fixes.Replacements.CoalescePolicy;
 import com.google.errorprone.util.ASTHelpers;
 import com.sun.source.tree.MethodTree;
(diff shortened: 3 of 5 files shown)


cpovirk:
I don't know enough to offer a definite way forward here, but I happen to know a couple relevant pieces of information:

- I imagine that a large number of checkers depend on deduplication. That's not to say that it's a great idea for them to do so, but it can be convenient :) For example, I wrote a checker that looks for `return null` and recommends inserting `@Nullable` on the enclosing method. It sometimes finds [methods that contain `return null` more than once, in which case it outputs duplicate suggestions](https://github.com/google/error-prone/blob/91cb47379594bb98b25f0a1cf9417213e86b0f47/core/src/main/java/com/google/errorprone/bugpatterns/nullness/ReturnMissingNullable.java#L383-L385).
- I recall that we had a problem very similar to the one you're reporting but with `MustBeClosedChecker`. We [added some code to manually insert _n_ closing braces as a single `postFix` with call](https://github.com/google/error-prone/commit/099a46bcf20350e1d4186d98a54a8500c2139b68#diff-e6911fcc1099aa7c464cfba814f703ef1a96bd2a556f837ce985bc0b8f08f3f3R497), and then we were able to [rip it out](https://github.com/google/error-prone/commit/b19ba52e167778ba509db9f9953cf13600db8e5f#diff-e6911fcc1099aa7c464cfba814f703ef1a96bd2a556f837ce985bc0b8f08f3f3L509) for reasons that I'm not actually clear on. I don't know if that's evidence that we should support a proper fix (like you're proposing), evidence that we'd rather hack around it, or evidence that there's actually some superior way of doing it that I'm unaware of.

Actual core Error Prone team members may have more useful advice.

cushon:
> * I imagine that a large number of checkers depend on deduplication. That's not to say that it's a great idea for them to do so, but it can be convenient :) For example, I wrote a checker that looks for `return null` and recommends inserting `@Nullable` on the enclosing method. It sometimes finds [methods that contain `return null` more than once, in which case it outputs duplicate suggestions](https://github.com/google/error-prone/blob/91cb47379594bb98b25f0a1cf9417213e86b0f47/core/src/main/java/com/google/errorprone/bugpatterns/nullness/ReturnMissingNullable.java#L383-L385).

I had a quick look at whether we could 'just' preserve duplicate inserts by default. ReturnMissingNullable was one example I ran into, there were a couple of others that similarly emitted the same method-level fix once per occurrence of a problem int he method body, leading to things like duplicate `@SuppressWarnings` annotations, or duplicate exceptions in `throws`, etc.

> * I recall that we had a problem very similar to the one you're reporting but with `MustBeClosedChecker`. We [added some code to manually insert _n_ closing braces as a single `postFix` with call](https://github.com/google/error-prone/commit/099a46bcf20350e1d4186d98a54a8500c2139b68#diff-e6911fcc1099aa7c464cfba814f703ef1a96bd2a556f837ce985bc0b8f08f3f3R497), and then we were able to [rip it out](https://github.com/google/error-prone/commit/b19ba52e167778ba509db9f9953cf13600db8e5f#diff-e6911fcc1099aa7c464cfba814f703ef1a96bd2a556f837ce985bc0b8f08f3f3L509) for reasons that I'm not actually clear on. I don't know if that's evidence that we should support a proper fix (like you're proposing), evidence that we'd rather hack around it, or evidence that there's actually some superior way of doing it that I'm unaware of.

I think part of what happened with MustBeClosedChecker was that the `DuplicateInsertPolicy` stuff got added, and that was configured when using MustBeClosedChecker as a refactoring tool.

I suspect we're not done play whack-a-mole with edge cases involving overlapping and adjacent fixes, but I also think allowing MissingBraces to opt-in to this is an improvement over the status quo.

cushon (approved):

PR #3818, opened 2023-03-11
Improve clarity of CheckedExceptionNotThrown description

delanym:
Improve clarity of CheckedExceptionNotThrown description


cushon:
Thanks for the PR, can you say more about why the current message is unclear? This text is usually going to be read by someone looking at a specific diagnostic, which will only be reported on methods that aren't overridable.

I think it's useful to mention the caveat about this guidance not applying to overridable methods in the docs, but I'm not sure that it's necessary to lead with it.

delanym:
Fine

Issue #3835, opened 2023-03-25
Dependency org.yaml:snakeyaml, leading to CVE problem


CVEDetect:
Hi, In **/docgen**，there is a dependency **org.yaml:snakeyaml:1.30** that calls the risk method.

[CVE-2022-25857](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-25857)

The scope of this CVE affected version is **[0,1.31)**

After further analysis, in this project, the main Api called is **org.yaml.snakeyaml.composer.Composer: composeNode(org.yaml.snakeyaml.nodes.Node)Lorg.yaml.snakeyaml.nodes.Node;**

Risk method repair link : [GitHub](https://github.com/snakeyaml/snakeyaml/commit/fc300780da21f4bb92c148bc90257201220cf174)

**CVE Bug Invocation Path--**

**Path Length : 5**

```
CVE Bug Invocation Path : 
com.google.errorprone.BugPatternIndexWriter: dump(java.util.Collection,java.io.Writer,com.google.errorprone.DocGenTool$Target,java.util.Set)V /.m2/repository/com/google/errorprone/error_prone_type_annotations/HEAD-SNAPSHOT/error_prone_type_annotations-HEAD-SNAPSHOT.jar
org.yaml.snakeyaml.Yaml$1: next()Ljava.lang.Object; /.m2/repository/com/beust/jcommander/1.82/jcommander-1.82.jar
org.yaml.snakeyaml.constructor.BaseConstructor: getData()Ljava.lang.Object; /.m2/repository/com/beust/jcommander/1.82/jcommander-1.82.jar
org.yaml.snakeyaml.composer.Composer: getNode()Lorg.yaml.snakeyaml.nodes.Node; /.m2/repository/com/beust/jcommander/1.82/jcommander-1.82.jar
org.yaml.snakeyaml.composer.Composer: composeNode(org.yaml.snakeyaml.nodes.Node)Lorg.yaml.snakeyaml.nodes.Node;

```
**Dependency tree--**

```
[INFO] com.google.errorprone:error_prone_docgen:jar:HEAD-SNAPSHOT
[INFO] +- com.google.errorprone:error_prone_annotation:jar:HEAD-SNAPSHOT:compile
[INFO] +- com.google.errorprone:error_prone_core:jar:HEAD-SNAPSHOT:compile
[INFO] |  +- com.google.errorprone:error_prone_type_annotations:jar:HEAD-SNAPSHOT:compile
[...]

PR #3836, opened 2023-03-25
Fix CVE dependency issue

CVEDetect:
Fix issue #3835   by update dependency org.yaml:snakeyaml:1.33 @cushon


cushon:
This will be resolved by https://github.com/google/error-prone/pull/3882

PR #3837, opened 2023-03-27
Fix UnusedVariable false positives for private record parameters

### LINKED ISSUE #2713: UnusedVariable sometimes gives false positives for record parameters
Using error-prone 2.10.0.

Here's a trivial example of a record where I get a warning that the `x` value is never being used, even though there are public methods that access it both directly and via its getter.

```java
public class Example
{
    private static record MyRecord(int x) { }

    public static int testMethod1(int x) {
        return new MyRecord(x).x();
    }

    public static int testMethod2(int x) {
        return new MyRecord(x).x;
    }
}
```

```
Example.java:[3,12] [UnusedVariable] The parameter 'x' is never read.
```

I don't know the exact conditions that trigger this, but for example if I make the record itself `public` the warning goes away.

Marcono1234:
Fixes #2713

It looks like the tree emitted by the compiler is incomplete for the canonical constructor of a record, which is why the parameters were erroneously flagged as unused. Apparently this only affected `private` record classes because for them the implicit canonical constructor is `private` as well, and the `UnusedVariable` check treats parameters of `private` methods and constructors differently.

https://github.com/google/error-prone/issues/2713#issuecomment-1079038446 showed an example where this also affected record fields, however I was unable to reproduce that. Maybe they were using an older Error Prone version which does not include https://github.com/google/error-prone/commit/47da3af7490e6dbb214af58ea6460ae5823f50e3, their sample code is incomplete, or my test setup was incorrect.
Edit: It looks like they were using an older Error Prone version, see https://github.com/google/error-prone/issues/2713#issuecomment-1489916690.


Marcono1234:
@cushon, are the changes ok like this or do you want anything changed?

core/src/main/java/com/google/errorprone/bugpatterns/UnusedVariable.java
```diff
@@ -623,6 +623,13 @@ && exemptedFieldBySuperType(getType(variableTree), state)) {
           if (variableTree.getName().contentEquals("this")) {
             return null;
           }
+          // Ignore if parameter is part of canonical record constructor; tree does not seem
+          // to contain usage in that case, but parameter is always used implicitly
+          // For compact canonical constructor parameters don't have record flag so need to
+          // check constructor flags (`symbol.owner`) instead
+          if (hasRecordFlag(symbol) || hasRecordFlag(symbol.owner)) {
```
cushon, line 630: Are there any cases where `hasRecordFlag(symbol)` is true but `hasRecordFlag(symbol.owner)` isn't?

core/src/test/java/com/google/errorprone/bugpatterns/UnusedVariableTest.java
```diff
@@ -1385,7 +1385,7 @@ public void simpleRecord() {
     assumeTrue(RuntimeVersion.isAtLeast16());
     helper
         .addSourceLines(
-            "SimpleRecord.java", //
+            "SimpleRecord.java",
```
cushon, line 1388: Please revert, the line comment forces google-java-format to add a line break

cushon (changes requested):
Thanks for the fix, and sorry for the slow review!

core/src/main/java/com/google/errorprone/bugpatterns/UnusedVariable.java
Marcono1234: Are you asking if this could be simplified to only check `hasRecordFlag(symbol.owner)`?
It looks like the tests still pass if I locally remove the additional `hasRecordFlag(symbol)` check, should I change this then?

core/src/test/java/com/google/errorprone/bugpatterns/UnusedVariableTest.java
Marcono1234: Ah sorry. Do you think it would make sense to change the comment to make this clearer (for example by changing it to `// (force line break for google-java-format)`, in case that still works then)? Or is this not worth it?

core/src/main/java/com/google/errorprone/bugpatterns/UnusedVariable.java
cushon: > Are you asking if this could be simplified to only check `hasRecordFlag(symbol.owner)`?

Yes

> It looks like the tests still pass if I locally remove the additional `hasRecordFlag(symbol)` check, should I change this then?

I think so, unless you suspect there might be cases where checking both is necessary?

core/src/test/java/com/google/errorprone/bugpatterns/UnusedVariableTest.java
cushon: I'd prefer just `//`

(And hopefully in the not too distant future we can start using raw string literals, at least in tests...)

cushon (approved):

cushon (approved):

Issue #2967, opened 2022-02-08
Errorprone not an OSGi bundle


Lana11s:
Currently, com.google.errorprone.annotations (version 2.10.0) is not an OSGi bundle. Some entries are missing in MANIFEST.MF:

```
Manifest-Version: 1.0
Automatic-Module-Name: com.google.errorprone.annotations
Archiver-Version: Plexus Archiver
Built-By: cushon
Created-By: Apache Maven 3.6.3
Build-Jdk: 1.8.0_202
```

Would be great if you can change this. Would allow us to upgrate to google/guice 5.1.0

Lana11s:
Hello @cushon,
I looked into previous releases of error-prone (incl. 2.0) and manifest file had the same structure.
Guice worked in the OSGi environment because the dependency to error-prone was introduced with 5.1.0.
I am not familiar with Maven Jar Plugin, but it seems that it needs some configuration so that entries (like import-package, export-package) are added: https://maven.apache.org/plugins/maven-jar-plugin/examples/manifest-customization.html

cdietrich:
i also just stumbled on this issue when testing the new guice

HannesWell:
@cushon if you are interested I can provide a PR to implement this (probably next week).

> I am not familiar with Maven Jar Plugin, but it seems that it needs some configuration so that entries (like import-package, export-package) are added: https://maven.apache.org/plugins/maven-jar-plugin/examples/manifest-customization.html

It is simpler to use the [`maven-bundle-plugin`](https://felix.apache.org/documentation/subprojects/apache-felix-maven-bundle-plugin-bnd.html) or [`bnd-maven-plugin`](https://github.com/bndtools/bnd/blob/master/maven-plugins/bnd-maven-plugin/README.md), because they can generate high quality OSGi metadata out of the box and can be fine tuned as necessary.
The latter plugin is used by the former one under the hood, which is why I recommend to use the latter one.

HannesWell:
Just created https://github.com/google/error-prone/pull/3903. Everybody interested, please join the discussion there.

PR #3903, opened 2023-05-10
Add OSGi runtime metadata to error-prone's MANIFEST.MF files

HannesWell:
As suggested in https://github.com/google/error-prone/issues/2967#issuecomment-1535520417 this PR aims to add OSGi runtime metadata to the error-prone jars.
It achieves this by employing the [bnd-maven-plugin:bnd-process](https://github.com/bndtools/bnd/tree/master/maven-plugins/bnd-maven-plugin#bnd-process-goal) goal to generate the OSGi metadata, like the values of the Manifest headers `Bundle-SymbolicName`, `Import-Package` and `Export-Package`.

The schema used for `Bundle-SymbolicName` headers is based on the schema used for the `Automatic-Module-Name` in the error-prone-annotations artifact.
With this change additionally all other error-prone artifacts get a `Automatic-Module-Name` too.

@cushon can you please review this PR? If you have any question, I would be happy to answer them.

Fixes https://github.com/google/error-prone/issues/2967

### DIFF
--- a/annotations/pom.xml
+++ b/annotations/pom.xml
@@ -54,16 +54,6 @@
           <compilerArgs combine.self="override" />
         </configuration>
       </plugin>
-      <plugin>
-        <artifactId>maven-jar-plugin</artifactId>
-        <configuration>
-          <archive>
-            <manifestEntries>
-              <Automatic-Module-Name>com.google.errorprone.annotations</Automatic-Module-Name>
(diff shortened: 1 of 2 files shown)


cushon (approved):

cushon (approved):

HannesWell:
Thank you for your review.

PR #3988, opened 2023-06-22
Replace ZoneId.of("UTC") with ZoneOffset.UTC

pkoenig10:
This PR updates the `ZoneIdOfZ` check to also replace `ZoneId.of("UTC")`.

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/time/ZoneIdOfZ.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/time/ZoneIdOfZ.java
@@ -40,9 +40,9 @@ import com.sun.source.tree.MethodInvocationTree;
  * @author kak@google.com (Kurt Alfred Kluever)
  */
 @BugPattern(
-    summary = "Use ZoneOffset.UTC instead of ZoneId.of(\"Z\").",
+    summary = "Use ZoneOffset.UTC instead of ZoneId.of(\"Z\") or ZoneId.of(\"UTC\").",
     explanation =
-        "Avoid the magic constant (ZoneId.of(\"Z\")) in favor of a more descriptive API: "
+        "Avoid the magic constants (ZoneId.of(\"Z\")) and (ZoneId.of(\"UTC\")) in favor of a more descriptive API: "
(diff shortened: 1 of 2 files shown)


core/src/test/java/com/google/errorprone/bugpatterns/time/ZoneIdOfZTest.java
```diff
@@ -63,6 +76,18 @@ public void zoneIdOfLowerCaseZ() {
...
   }
 
+  @Test
+  public void zoneIdOfLowerCaseUtc() {
+    helper
+        .addSourceLines(
+            "TestClass.java",
+            "import java.time.ZoneId;",
+            "public class TestClass {",
+            "  private static final ZoneId UTC = ZoneId.of(\"utc\");",
```
kluever, line 86: Not a huge deal, but http://errorprone.info/bugpattern/InvalidZoneId should prevent this from compiling in the first place :-)

core/src/main/java/com/google/errorprone/bugpatterns/time/ZoneIdOfZ.java
```diff
@@ -57,7 +57,7 @@ public final class ZoneIdOfZ extends BugChecker implements MethodInvocationTreeM
   public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
     if (ZONE_ID_OF.matches(tree, state)) {
       String zone = constValue(tree.getArguments().get(0), String.class);
-      if (zone != null && zone.equals("Z")) {
+      if (zone != null && (zone.equals("Z") || zone.equals("UTC"))) {
```
kluever, line 60: So technically, `ZoneId.of("UTC")` is not equal to `ZoneOffset.UTC`. They have different values for `getId()`, (the former is `"Z"` and the latter is `"UTC"` I believe)...whether this matters in practice, I'm not sure.

Any ideas?

kluever: Yea let's revert this to be on the safe side, especially since this is enabled as an error, not just a warning.

PR #4003, opened 2023-07-07
Fixed a bug in frangmentinjection.java

bryanspacex:
Found an error in this line:

```
Type preferenceActivityType = ANDROID_PREFERENCE_PREFERENCEACTIVITY.get(state);
```

replaced it with this:

```
private static final Supplier<Type> ANDROID_PREFERENCE_PREFERENCEACTIVITY =
    new Supplier<Type>() {
      @Override
      public Type get() {
        return state.getTypeFromString("android.preference.PreferenceActivity");
      }
    };
```

The variable `ANDROID_PREFERENCE_PREFERENCEACTIVITY` is used as a `Supplier<Type>`, but it is not defined correctly. It should be defined as a method that returns the `Type` object representing `android.preference.PreferenceActivity`.


cushon:
Hi, it doesn't look like the commit matches the description, I just see a change to `test_helpers/src/main/java/com/google/errorprone/BugCheckerRefactoringTestHelper.java` that adds an import?

PR #4028, opened 2023-07-31
Ignore disabled checks passed to `-XepPatchChecks`

### LINKED ISSUE #3908: 2.19.x: Exception thrown when a disabled check is passed to -XepPatchChecks
Standalone repro: https://github.com/AlexLandau/errorprone-disabled-and-patch-issue

We have some odd situations involving automated tooling where sometimes, we will run patch creation on checks that have also been disabled; that is, we end up with both `-Xep:CheckName:OFF` and `-XepPatchChecks:CheckName`. In 2.18.0, this resulted in the patch being created as if it weren't disabled. Starting in 2.19.0 (and through the current version, 2.19.1), an exception is thrown instead:

```
/home/alandau/code/gradle-disabled-and-patch-issue/lib/src/main/java/com/github/alexlandau/repro/Library.java:5: error: An unhandled exception was thrown by the Error Prone static analysis plugin.
    public String toString() {
                  ^

     error-prone version: 2.19.1
     BugPattern: MissingOverride
     Stack Trace:
        at com.google.errorprone.matchers.Description.severity(Description.java:78)
        at com.google.errorprone.ErrorProneAnalyzer.lambda$finished$1(ErrorProneAnalyzer.java:138)
        at com.google.errorprone.VisitorState.reportMatch(VisitorState.java:301)
        at com.google.errorprone.scanner.Scanner.reportMatch(Scanner.java:127)
        at com.google.errorprone.scanner.ErrorProneScanner.processMatchers(ErrorProneScanner.java:448)
        at com.google.errorprone.scanner.ErrorProneScanner.visitMethod(ErrorProneScanner.java:739)
        (etc., full stacktrace in repo linked above)
```

We can work around this for the time being (editing the `-Xep` args to make sure the check isn't disabled).

oxkitsune:
Rather than throwing an `NoSuchElementException`.

Fixes #3908.

### DIFF
--- a/core/src/test/java/com/google/errorprone/scanner/ScannerTest.java
+++ b/core/src/test/java/com/google/errorprone/scanner/ScannerTest.java
@@ -95,6 +95,22 @@ public class ScannerTest {
         .doTest();
   }
 
+  @Test
+  public void dontRunPatchForDisabledChecks() {
+    compilationHelper
+        .addSourceLines(
+            "Test.java",
+            "import com.google.errorprone.scanner.ScannerTest.Foo;",
+            "class Test {",
+            "  Foo foo;",
+            "}")
+        .setArgs(
+            "-XepPatchLocation:IN_PLACE",
+            "-XepPatchChecks:ShouldNotUseFoo",
+            "-Xep:ShouldNotUseFoo:OFF")
+        .doTest();
+  }
+
   @OkToUseFoo // Foo can use itself. But this shouldn't suppress errors on *usages* of Foo.
   public static final class Foo<T> {}
 
--- a/core/src/test/java/com/google/errorprone/scanner/ScannerTest.java
+++ b/core/src/test/java/com/google/errorprone/scanner/ScannerTest.java
@@ -106,9 +106,24 @@ public class ScannerTest {
             "}")
         .setArgs(
             "-XepPatchLocation:IN_PLACE",
-            "-XepPatchChecks:ShouldNotUseFoo",
+            "-XepPatchChecks:",
             "-Xep:ShouldNotUseFoo:OFF")
         .doTest();
+  }
+
+  @Test
+  public void dontRunPatchUnlessRequested() {
+    compilationHelper.addSourceLines(
+            "Test.java",
+            "import com.google.errorprone.scanner.ScannerTest.Foo;",
+            "class Test {",
+            "  Foo foo;",
+            "}")
+            .setArgs(
+                    "-XepPatchLocation:IN_PLACE",
+                    "-Xep:ShouldNotUseFoo:WARN",
+                    "-XepPatchChecks:DeadException"
+            ).doTest();
   }
 
   @OkToUseFoo // Foo can use itself. But this shouldn't suppress errors on *usages* of Foo.
--- a/check_api/src/main/java/com/google/errorprone/BaseErrorProneJavaCompiler.java
+++ b/check_api/src/main/java/com/google/errorprone/BaseErrorProneJavaCompiler.java
@@ -228,8 +228,10 @@ public class BaseErrorProneJavaCompiler implements JavaCompiler {
                   ImmutableSet<String> namedCheckers = epOptions.patchingOptions().namedCheckers();
                   if (!namedCheckers.isEmpty()) {
                     toUse = toUse.filter(bci -> namedCheckers.contains(bci.canonicalName()));
+                  } else {
+                    toUse = toUse.applyOverrides(epOptions);
                   }
-                  return ErrorProneScannerTransformer.create(toUse.applyOverrides(epOptions).get());
+                  return ErrorProneScannerTransformer.create(toUse.get());
                 })
             .get();
 
--- a/check_api/src/main/java/com/google/errorprone/BaseErrorProneJavaCompiler.java
+++ b/check_api/src/main/java/com/google/errorprone/BaseErrorProneJavaCompiler.java
@@ -224,14 +224,12 @@ public class BaseErrorProneJavaCompiler implements JavaCompiler {
             .customRefactorer()
             .or(
                 () -> {
-                  ScannerSupplier toUse =
-                      ErrorPronePlugins.loadPlugins(scannerSupplier, context)
-                          .applyOverrides(epOptions);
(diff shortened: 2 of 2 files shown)


rickie (approved):

Stephan202 (approved):
@oxkitsune these PRs are generally merged with the PR text as commit message, so I would also add this sentence to the PR body (just like in the commit message):
> Rather than throwing an `NoSuchElementException`.

rickie:
> @oxkitsune these PRs are generally merged with the PR text as commit message, so I would also add this sentence to the PR body (just like in the commit message):

That's already the case right? It should be good like it is right now.

Stephan202:
I would swear that my browser didn't show it. But I must have been confused. :thinking:

cushon (approved):

cushon (approved):

cushon (approved):

cushon:
Thanks for this fix!

I finally did some internal testing of this, and one thing I encountered is that this changes of behaviour of things like `-Xep:DeadException:WARN -XepPatchChecks:GetClassOnClass`--previously refactorings weren't emitted for `DeadException` even though it was explicitly enabled, because it wasn't listed in `-XepPatchChecks`. With this change refactorings are emitted for `DeadException`.

Do you have thoughts on what the least surprising behaviour here would be?

cushon (changes requested):
Un-approving until the question in https://github.com/google/error-prone/pull/4028#issuecomment-1710798061 is resolved

oxkitsune:
> Do you have thoughts on what the least surprising behaviour here would be?

Personally I think it would make the most sense, if checks passed to `-Xep:PatchChecks` are the only checks emitting patches, unless they're explicitly disabled.

cushon:
> > Do you have thoughts on what the least surprising behaviour here would be?
> 
> Personally I think it would make the most sense, if checks passed to `-Xep:PatchChecks` are the only checks emitting patches, unless they're explicitly disabled.
> 
> E.g.
> 
> ```
> -Xep:DeadException:WARN -XepPatchChecks:GetClassOnClass
> ```
> 
> would emit patches for `GetClassOnClass`, and warnings for `DeadException`.
> 
> ```
> -Xep:DeadException:OFF -XepPatchChecks:GetClassOnClass,DeadException
> ```
> 
> would emit _just_ patches for `GetClassOnClass`

That sounds good to me.

I think that there isn't an order that `applyOverrides` and `filter` can be called in that does quite what we want here: calling `applyOverrides` first causes #3908, and calling `filter` first causes refactorings to be emitted for checks that are explicitly enabled with `-Xep:` flags.

I wonder if it makes sense of `-XepPatchChecks` to just override any `-Xep:...:[OFF|WARN|ERROR]` flags. That would mean that we went back to the pre-`2.19` behaviour where refactorings were emitted for `-Xep:CheckName:OFF -XepPatchChecks:CheckName`. Would that work for you?

core/src/test/java/com/google/errorprone/scanner/ScannerTest.java
```diff
@@ -95,6 +95,22 @@ public void suppressionAnnotationIgnoredWithOptions() {
...
+            "class Test {",
+            "  Foo foo;",
+            "}")
+        .setArgs(
+            "-XepPatchLocation:IN_PLACE",
+            "-XepPatchChecks:ShouldNotUseFoo",
+            "-Xep:ShouldNotUseFoo:OFF")
+        .doTest();
+  }
+
```
cushon, line 113: Can you add another test case for

```java
  @Test
  public void dontRunPatchUnlessRequested() {
    compilationHelper
        .addSourceLines(
            "Test.java",
            "import com.google.errorprone.scanner.ScannerTest.Foo;",
            "class Test {",
            "  Foo foo;",
            "}")
        .setArgs(
            "-XepPatchLocation:IN_PLACE",
            "-Xep:ShouldNotUseFoo:WARN",
            "-XepPatchChecks:DeadException")
        .doTest();
  }
  ```

**Comment by Stephan202** [comment]
> I wonder if it makes sense of `-XepPatchChecks` to just override any `-Xep:...:[OFF|WARN|ERROR]` flags. That would mean that we went back to the pre-`2.19` behaviour where refactorings were emitted for `-Xep:CheckName:OFF -XepPatchChecks:CheckName`. Would that work for you?

I'll need to find some time later to think this through more thoroughly, but I wonder whether we'll regret that if #947 is ever merged (still hoping :wink:): then `-XepPatchChecks:` (without further args) would enable patching for all checks, and realistically users will want to apply some exclusions.

That said, we have the #947 change [in our fork](https://github.com/PicnicSupermarket/error-prone/compare/v2.21.1...v2.21.1-picnic-1), and right now we do have this "patch all except those that are `OFF`" behavior, with a baseline more recent than 2.19.

Easiest is perhaps to first add a few more tests, and then assess impact on #947.

**Comment by cushon** [comment]
> I'll need to find some time later to think this through more thoroughly

My too, I think :)

The existing logic only calls filter if `!namedCheckers.isEmpty()`, if we keep that part and do the rest of what I described, I think that would interact well with #947

i.e. something like:

```java
                  ScannerSupplier toUse = ErrorPronePlugins.loadPlugins(scannerSupplier, context);
                  ImmutableSet<String> namedCheckers = epOptions.patchingOptions().namedCheckers();
                  if (!namedCheckers.isEmpty()) {
                    toUse = toUse.filter(bci -> namedCheckers.contains(bci.canonicalName()));
                  } else {
                    toUse = toUse.applyOverrides(epOptions);
                  }
                  return ErrorProneScannerTransformer.create(toUse.get());
```

Stephan202:
Rebased this branch to include the changes of #947 (diff applied cleanly; thanks for merging @cushon!), which should make it marginally easier to reason about the solution space. (Actual reasoning still TBD :sweat_smile:)

oxkitsune:
Hey @cushon apologies for the inactivity :sweat_smile: 
I've just applied the suggestions and I'm happy with this solution!

cushon:
> Hey @cushon apologies for the inactivity 😅 I've just applied the suggestions and I'm happy with this solution!

No worries, thanks for coming back to this!

cushon (approved):

PR #4043, opened 2023-08-09
Prevent `SuggestedFixes#renameMethod` from modifying return type declaration

oxkitsune:
By ensuring the return type `IDENTIFIER` token is not matched.

### DIFF
--- a/check_api/src/main/java/com/google/errorprone/fixes/SuggestedFixes.java
+++ b/check_api/src/main/java/com/google/errorprone/fixes/SuggestedFixes.java
@@ -745,17 +745,14 @@ public final class SuggestedFixes {
 
   /** Be warned, only changes method name at the declaration. */
   public static SuggestedFix renameMethod(MethodTree tree, String replacement, VisitorState state) {
-    // Search tokens from beginning of method tree to beginning of method body.
-    int basePos = getStartPosition(tree);
+    // Search tokens from end of return type tree to beginning of method body.
+    int basePos = state.getEndPosition(tree.getReturnType());
     int endPos =
         tree.getBody() != null ? getStartPosition(tree.getBody()) : state.getEndPosition(tree);
     List<ErrorProneToken> methodTokens = state.getOffsetTokens(basePos, endPos);
 
-    int returnTypeEndPos = state.getEndPosition(tree.getReturnType());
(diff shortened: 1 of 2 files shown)


rickie (approved):

core/src/test/java/com/google/errorprone/bugpatterns/MemberNameTest.java
```diff
@@ -463,4 +463,32 @@ public void methodReference() {
...
+            "  }",
+            "",
+            "  void call() {",
+            "     Object();",
+            "  }",
+            "}")
+        .addOutputLines(
+            "Test.java",
+            "class Test {",
+            "  private Object object() {",
```
Stephan202, line 484: For other reviewers: before this fix Error Prone would suggest:
```suggestion
            "  private object Object() {",
```

Stephan202 (approved):

check_api/src/main/java/com/google/errorprone/fixes/SuggestedFixes.java
```diff
@@ -750,8 +750,12 @@ public static SuggestedFix renameMethod(MethodTree tree, String replacement, Vis
     int endPos =
         tree.getBody() != null ? getStartPosition(tree.getBody()) : state.getEndPosition(tree);
     List<ErrorProneToken> methodTokens = state.getOffsetTokens(basePos, endPos);
+
+    int returnTypeEndPos = state.getEndPosition(tree.getReturnType());
     for (ErrorProneToken token : methodTokens) {
-      if (token.kind() == TokenKind.IDENTIFIER && token.name().equals(tree.getName())) {
+      if (token.kind() == TokenKind.IDENTIFIER
+          && token.pos() > returnTypeEndPos
```
cushon, line 757: Instead of checking the token positions here, I think we might be able to adjust `basePos` earlier and only tokenize from after the return type. Do you see any issues with that approach?

cushon (changes requested):

check_api/src/main/java/com/google/errorprone/fixes/SuggestedFixes.java
oxkitsune: That's a good catch thanks! Will push a commit.

cushon (approved):

PR #4060, opened 2023-09-02
Flag unused Refaster template parameters

Stephan202:


core/src/main/java/com/google/errorprone/bugpatterns/UnusedVariable.java
```diff
@@ -149,6 +149,8 @@ public final class UnusedVariable extends BugChecker implements CompilationUnitT
   // TODO(ghm): Find a sensible place to dedupe this with UnnecessarilyVisible.
```
Stephan202, line 149: If acceptable I don't mind also updating [this list](https://github.com/google/error-prone/blob/b9e01a8d4f2a7a822eaf6b8f95f72f2f0ca72912/core/src/main/java/com/google/errorprone/bugpatterns/UnnecessarilyVisible.java#L59-L64). Alternatively we can upstream our [`RefasterRuleModifiers`](https://github.com/PicnicSupermarket/error-prone-support/blob/87f79dae60bd1be34f71ebeeba7d783670f96e70/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/RefasterRuleModifiers.java) check, which normalizes modifier usage by Refaster rule definitions more generally.

core/src/main/java/com/google/errorprone/bugpatterns/UnusedVariable.java
```diff
@@ -149,6 +149,8 @@ public final class UnusedVariable extends BugChecker implements CompilationUnitT
...
   private static final ImmutableSet<String> ANNOTATIONS_INDICATING_PARAMETERS_SHOULD_BE_CHECKED =
       ImmutableSet.of(
+          "com.google.errorprone.refaster.annotation.AfterTemplate",
+          "com.google.errorprone.refaster.annotation.BeforeTemplate",
```
Stephan202, line 153: Of this list, currently there's only a test for `javax.inject.Inject`, so I didn't add any test cases for these new annotations, either. Let me know if I should. (I did test these changes against a collection of real Refaster templates.)

core/src/main/java/com/google/errorprone/bugpatterns/UnusedVariable.java
cushon: > If acceptable I don't mind also updating [this list](https://github.com/google/error-prone/blob/b9e01a8d4f2a7a822eaf6b8f95f72f2f0ca72912/core/src/main/java/com/google/errorprone/bugpatterns/UnnecessarilyVisible.java#L59-L64).

Thanks, updating the list until we get around to deduping them SGTM

> Alternatively we can upstream our [`RefasterRuleModifiers`](https://github.com/PicnicSupermarket/error-prone-support/blob/87f79dae60bd1be34f71ebeeba7d783670f96e70/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/RefasterRuleModifiers.java) check, which normalizes modifier usage by Refaster rule definitions more generally.

That also sounds good if you have time/interest, but it also doesn't need to block this PR

cushon (changes requested):
LG pending the update to UnnecessarilyVisible

core/src/main/java/com/google/errorprone/bugpatterns/UnusedVariable.java
Stephan202: > Thanks, updating the list until we get around to deduping them SGTM

Top; will do.

> That also sounds good if you have time/interest, but it also doesn't need to block this PR

I'll make a note to consider this later then :heavy_check_mark:. (When we contributed #3813 and #2232 that caused some issues for downstream users due to a name collision (PicnicSupermarket/error-prone-support#622 and PicnicSupermarket/error-prone-support#686), so maybe this time I'll try to find an alternative name; hehe.)

Stephan202 (commented):
Rebased and added a commit to also update `UnnecessarilyVisible`.

cushon (approved):

PR #4086, opened 2023-09-13
Drop unused `DiffNotApplicableException`

Stephan202:
This exception type was introduced along with Refaster in
01f713649844a4cc331d8055d7a1e2f693e2e8b0, but no instance is ever
instantiated.


cushon:
> I noticed this while working on #4085. It could of course be that some Google-internal code does reference this type, in which case this PR may not make sense or require significant modification.

That is unfortunately the case, but I will take a closer look at whether the single place that's instantiating this exception type still makes sense

Stephan202:
Interesting! I don't immediately have a strong opinion either. Some initial thoughts:
- Going for option (1) would likely be easier to do if the other diff logic is open-source/included, so that the requisite changes can be tested more easily.
- On the other hand, one could argue that in case of invalid replacements, the proper thing to do is to fix the check that emits these invalid replacements. :sweat_smile:
    - This stance assumes that the suggested fixes aren't made invalid by _other_ changes, but IIUC, as long as all suggested fixes are applied in one go, that shouldn't be an issue.<sup>&dagger;</sup>
    - (It is true that occasionally emitting the right fix can be tricky, because the AST-source correspondence isn't perfect. But so far I've found that a solution always exists.)

So all-in-all I'm tending to "let's keep it simple" and go for option (2), but that does assume of course that the Google-internal diff is no longer relied upon.

<sup>&dagger;</sup>: Distantly related here is #4088, but as far as I can tell the general case of chaining dependent bug checkers will always require an intermediate compilation phase, in which case there's no reason to assume that the replacements emitted by each round can't be accurate, relative to the source code produced by the previous round.

cushon:
Thanks for thinking this through! I am leaning towards (2).

> one could argue that in case of invalid replacements, the proper thing to do is to fix the check that emits these invalid replacements.

I think that is an extremely valid argument :) I think the motivation might have been for iterating on new checks, where it's useful to apply the fixes we can even if some of them are bogus. We'd also want to fix the bad findings, but it takes some time to run the check across the depot, and being able to apply a subset of the fixes in that case cuts down on the number of iterations. But (2) is still fine for that.

cushon:
DiffNotApplicableException has been removed in https://github.com/google/error-prone/commit/737dec07570ae14f71c808281187adbb5cd947cb. Thanks for pointing this out!

jeandersonbc:
> but as far as I can tell the general case of chaining dependent bug checkers will always require an intermediate compilation phase

Thanks for the input @Stephan202!

PR #4114, opened 2023-09-24
Update DefaultCharset.java

delanym:
Prefer (?, Charset) to (?, String) constructor

Such as https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/Scanner.html#%3Cinit%3E(java.io.File,java.nio.charset.Charset)

### DIFF
--- a/core/src/test/java/com/google/errorprone/bugpatterns/DefaultCharsetTest.java
+++ b/core/src/test/java/com/google/errorprone/bugpatterns/DefaultCharsetTest.java
@@ -389,8 +389,8 @@ public class DefaultCharsetTest {
             "        new BufferedWriter(new OutputStreamWriter(System.err, UTF_8)), true);",
             "    PrintWriter pw2 = new PrintWriter(",
             "        new BufferedWriter(new OutputStreamWriter(System.err, UTF_8)));",
-            "    PrintWriter pw3 = new PrintWriter(\"test\", UTF_8.name());",
-            "    PrintWriter pw4 = new PrintWriter(new File(\"test\"), UTF_8.name());",
+            "    PrintWriter pw3 = new PrintWriter(\"test\", UTF_8);",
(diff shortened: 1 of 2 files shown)


cushon (changes requested):
Looks good overall, thanks for the PR.

The tests are currently failing because they are assertion on the old version of the refactoring that included `.name()`, can you update them?

* https://github.com/google/error-prone/blob/11471840328880711e58d009b0d4a29517620d16/core/src/test/java/com/google/errorprone/bugpatterns/DefaultCharsetTest.java#L392-L393
* https://github.com/google/error-prone/blob/11471840328880711e58d009b0d4a29517620d16/core/src/test/java/com/google/errorprone/bugpatterns/DefaultCharsetTest.java#L497-L500

core/src/test/java/com/google/errorprone/bugpatterns/DefaultCharsetTest.java
```diff
@@ -389,8 +389,8 @@ public void printWriter() {
             "        new BufferedWriter(new OutputStreamWriter(System.err, UTF_8)), true);",
             "    PrintWriter pw2 = new PrintWriter(",
             "        new BufferedWriter(new OutputStreamWriter(System.err, UTF_8)));",
-            "    PrintWriter pw3 = new PrintWriter(\"test\", UTF_8.name());",
-            "    PrintWriter pw4 = new PrintWriter(new File(\"test\"), UTF_8.name());",
+            "    PrintWriter pw3 = new PrintWriter(\"test\", UTF_8);",
+            "    PrintWriter pw4 = new PrintWriter(new File(\"test\"), UTF_8);",
```
cushon, line 393: I missed that the change was only updating the refactoring for `Scanner`, not `PrintWriter`

Can you also remove the use of `.name()` on L547 in `handlePrintWriter`?

cushon (changes requested):

cushon (approved):

PR #4123, opened 2023-09-28
VoidMissingNullable: handle implicit parameterized types

Stephan202:
While there, simplify a few other places where implicit AST nodes are
identified.

### DIFF
--- a/check_api/src/main/java/com/google/errorprone/fixes/SuggestedFixes.java
+++ b/check_api/src/main/java/com/google/errorprone/fixes/SuggestedFixes.java
@@ -27,7 +27,7 @@ import static com.google.errorprone.util.ASTHelpers.getAnnotationWithSimpleName;
 import static com.google.errorprone.util.ASTHelpers.getModifiers;
 import static com.google.errorprone.util.ASTHelpers.getStartPosition;
 import static com.google.errorprone.util.ASTHelpers.getSymbol;
-import static com.google.errorprone.util.ASTHelpers.hasNoExplicitType;
+import static com.google.errorprone.util.ASTHelpers.hasImplicitType;
 import static com.sun.source.tree.Tree.Kind.ASSIGNMENT;
 import static com.sun.source.tree.Tree.Kind.CONDITIONAL_EXPRESSION;
 import static com.sun.source.tree.Tree.Kind.NEW_ARRAY;
@@ -1118,7 +1118,7 @@ public final class SuggestedFixes {
                         && ((ClassTree) tree).getSimpleName().length() != 0)
                     // Lambda parameters can't be suppressed unless they have Type decls
                     || (tree instanceof VariableTree
-                        && !hasNoExplicitType((VariableTree) tree, state)))
+                        && !hasImplicitType((VariableTree) tree, state)))
         .findFirst()
         .orElse(null);
   }
--- a/check_api/src/main/java/com/google/errorprone/fixes/SuggestedFixes.java
+++ b/check_api/src/main/java/com/google/errorprone/fixes/SuggestedFixes.java
@@ -27,6 +27,7 @@ import static com.google.errorprone.util.ASTHelpers.getAnnotationWithSimpleName;
 import static com.google.errorprone.util.ASTHelpers.getModifiers;
 import static com.google.errorprone.util.ASTHelpers.getStartPosition;
 import static com.google.errorprone.util.ASTHelpers.getSymbol;
+import static com.google.errorprone.util.ASTHelpers.hasNoExplicitType;
 import static com.sun.source.tree.Tree.Kind.ASSIGNMENT;
 import static com.sun.source.tree.Tree.Kind.CONDITIONAL_EXPRESSION;
 import static com.sun.source.tree.Tree.Kind.NEW_ARRAY;
@@ -1018,7 +1019,7 @@ public final class SuggestedFixes {
       @Nullable String lineComment,
       boolean commentOnNewLine) {
     // Find the nearest tree to add @SuppressWarnings to.
-    Tree suppressibleNode = suppressibleNode(state.getPath());
+    Tree suppressibleNode = suppressibleNode(state.getPath(), state);
     if (suppressibleNode == null) {
       return;
     }
@@ -1069,7 +1070,7 @@ public final class SuggestedFixes {
   public static void removeSuppressWarnings(
       SuggestedFix.Builder fixBuilder, VisitorState state, String warningToRemove) {
     // Find the nearest tree to remove @SuppressWarnings from.
-    Tree suppressibleNode = suppressibleNode(state.getPath());
+    Tree suppressibleNode = suppressibleNode(state.getPath(), state);
     if (suppressibleNode == null) {
       return;
     }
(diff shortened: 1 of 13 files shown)


check_api/src/main/java/com/google/errorprone/fixes/SuggestedFixes.java
```diff
@@ -1117,7 +1118,7 @@ private static Tree suppressibleNode(TreePath path) {
                         && ((ClassTree) tree).getSimpleName().length() != 0)
                     // Lambda parameters can't be suppressed unless they have Type decls
                     || (tree instanceof VariableTree
-                        && getStartPosition(((VariableTree) tree).getType()) != -1))
+                        && !hasNoExplicitType((VariableTree) tree, state)))
```
Stephan202, line 1121: This shows that `hasNoExplicitType` should perhaps be renamed to `hasExplicitType`, with negated semantics. WDYT?

core/src/main/java/com/google/errorprone/bugpatterns/nullness/VoidMissingNullable.java
```diff
@@ -86,6 +87,11 @@ public class VoidMissingNullable extends BugChecker
   @Override
   public Description matchParameterizedType(
       ParameterizedTypeTree parameterizedTypeTree, VisitorState state) {
+    if (!hasExplicitSource(parameterizedTypeTree, state)) {
+      /* Implicit types cannot be annotated. */
+      return NO_MATCH;
+    }
```
Stephan202, line 93: We could expand the comment [like here](https://github.com/google/error-prone/blob/43a6c45b20acb4395913eeb152b1d06e28c10022/core/src/main/java/com/google/errorprone/bugpatterns/nullness/VoidMissingNullable.java#L144-L152), but I'm not sure that we want to make a similarly strong statement about inference of local parameterized types.

core/src/test/java/com/google/errorprone/bugpatterns/nullness/VoidMissingNullableTest.java
```diff
@@ -294,6 +294,20 @@ public void negativeLambdaParameterNoType() {
...
+        .addSourceLines(
+            "Test.java",
+                "import org.jspecify.annotations.Nullable;",
+            "interface Test {",
+            "  void consume(Iterable<@Nullable Void> it);",
+            "",
+            "  Test TEST = it -> {};",
+            "}")
+        .doTest();
+  }
```
Stephan202, line 309: Without these changes the test fails with:
```
java.lang.AssertionError: An unhandled exception was thrown by the Error Prone static analysis plugin.
  
     error-prone version: unknown version
     BugPattern: VoidMissingNullable
     Stack Trace:
  	at com.google.errorprone.bugpatterns.nullness.NullnessUtils.fixByAddingKnownTypeUseNullableAnnotation(NullnessUtils.java:235)
  	at com.google.errorprone.bugpatterns.nullness.NullnessUtils.fixByAnnotatingTypeUseOnlyLocationWithNullableAnnotation(NullnessUtils.java:200)
  	at com.google.errorprone.bugpatterns.nullness.VoidMissingNullable.checkTree(VoidMissingNullable.java:191)
  	at com.google.errorprone.bugpatterns.nullness.VoidMissingNullable.matchParameterizedType(VoidMissingNullable.java:107)
 ```

**Review by Stephan202 — COMMENTED** [review]
I'm open to alternative implementations. Also, let me know if I should extract the cleanup logic to a separate PR.

CC @cpovirk as the author of `VoidMissingNullable`.

File: check_api/src/main/java/com/google/errorprone/util/ASTHelpers.java [hunk 4]
```diff
@@ -2647,13 +2648,17 @@ private void scan(Type from, Type to) {
...
   public static boolean hasNoExplicitType(VariableTree tree, VisitorState state) {
     /*
-     * We detect the absence of an explicit type by looking for an absent start position for the
-     * type tree.
-     *
      * Note that the .isImplicitlyTyped() method on JCVariableDecl returns the wrong answer after
      * type attribution has occurred.
      */
-    return getStartPosition(tree.getType()) == -1;
+    return !hasExplicitSource(tree.getType(), state);
```
cushon, line 2654: We used to check `getEndPosition` in `hasNoExplicitType` to work around a javac8 bug, and the workaround was removed in https://github.com/google/error-prone/commit/1b8bf175843e3acaa6f22aa9c3070607af103273. I can't remember if there was any motivation for that beyond removing references to javac8 bugs after we dropped JDK 8 support.

I also remembered that recent versions of `JCVariableDecl` (including in JDK 11u) have a `declaredUsingVar` getter that I think we could be using here. We might also need to check `isImplicitlyTyped`.

Thoughts on using that approach instead of delegating? I think it reads as well and avoids having to check end positions.

check_api/src/main/java/com/google/errorprone/fixes/SuggestedFixes.java
cushon: I agree the double-negation is unfortunate, either `hasExplicitType` with negated semantics, or `hasImplicitType` with the same semantics seem like an improvement. If you want to make one of those changes here I'm happy to approve it :)

cushon (changes requested):
Looks good overall, thanks!

check_api/src/main/java/com/google/errorprone/util/ASTHelpers.java
Stephan202: Though the unconditional downcast casts always make me a little uncomfortable, there's precedent even in this class to unconditionally cast to `JCVariableDecl`. So I tried this:
```java
public static boolean hasImplicitType(VariableTree tree, VisitorState state) {
  JCVariableDecl varTree = (JCVariableDecl) tree;
  return varTree.declaredUsingVar() || varTree.isImplicitlyTyped();
}
```

Unfortunately this causes the existing `negativeLambdaParameterNoType` test to fail:
```
java.lang.AssertionError: Saw unexpected error on line 5. All errors:
/Test.java:5: Note: [VoidMissingNullable] The type Void is not annotated @Nullable
  Test TEST = v -> {};
              ^
    (see https://errorprone.info/bugpattern/VoidMissingNullable)
  Did you mean 'Test TEST = @Nullable v -> {};'?

	at org.junit.Assert.fail(Assert.java:89)
	at com.google.errorprone.DiagnosticTestHelper.assertHasDiagnosticOnAllMatchingLines(DiagnosticTestHelper.java:348)
	at com.google.errorprone.CompilationTestHelper.doTest(CompilationTestHelper.java:342)
	at com.google.errorprone.bugpatterns.nullness.VoidMissingNullableTest.negativeLambdaParameterNoType(VoidMissingNullableTest.java:294)
	...
```

When I attach a debugger then I see that indeed `declaredUsingVar` is false, and `vartype` is a `JCIdent` representing `Void`. I'm guessing (but did not investigate deeper) that this has something to do with the lambda expression being assigned the type of the targeted functional interface.

I'll propose a new comment to document this observation.

(Tested with Temurin-11.0.20.1+1 and Temurin-17.0.8.1+1.)

Stephan202 (commented):
Added a commit; also fixed some formatting.

check_api/src/main/java/com/google/errorprone/util/ASTHelpers.java
cushon: Thanks for trying that out, and for the comment.

cushon (approved):

PR #4157, opened 2023-10-21
UnnecessaryBoxedAssignment: avoid CCE on method reference

Stephan202:


core/src/test/java/com/google/errorprone/bugpatterns/testdata/UnnecessaryBoxedAssignmentCases.java
```diff
@@ -76,4 +78,8 @@ Integer positive_wrappedAgain(int aInteger) {
     a = Integer.valueOf(aInteger);
     return Integer.valueOf(a);
   }
+
+  void negative_methodReference() {
+    Function<String, Boolean> toBoolean = Boolean::valueOf;
```
Stephan202, line 83: Without the fix this triggers:
```
java.lang.ClassCastException: class com.sun.tools.javac.tree.JCTree$JCMemberReference cannot be cast to class com.sun.source.tree.MethodInvocationTree (com.sun.tools.javac.tree.JCTree$JCMemberReference and com.sun.source.tree.MethodInvocationTree are in module jdk.compiler of loader 'app')
	at com.google.errorprone.bugpatterns.UnnecessaryBoxedAssignment.matchCommon(UnnecessaryBoxedAssignment.java:72)
	at com.google.errorprone.bugpatterns.UnnecessaryBoxedAssignment.matchVariable(UnnecessaryBoxedAssignment.java:64)
```

Stephan202 (commented):
I did not really investigate what it would take to _support_ method references, but I suspect that would be a check all by itself, if such a thing would make sense at all.

cushon (approved):

cushon (approved):

PR #4162, opened 2023-10-25
Suggest maven-compiler-plugin 3.11.0 in docs

msridhar:
I found that with 3.10.1, Error Prone (and NullAway) warnings aren't even printed (!).  Some related discussion at https://github.com/uber/NullAway/issues/830.  I created a relevant sample repo:

https://github.com/msridhar/ep-maven-test-warnings

You can compare running `mvn clean compile` as is with changing [this line](https://github.com/msridhar/ep-maven-test-warnings/blob/main/pom.xml#L23) to be `3.10.1`; with the latter version many fewer warnings are printed, and no Error Prone warnings are printed as far as I can see.

I'm open to adding text that warns users off of versions earlier than 3.11.0 if we think that is warranted.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -46,7 +46,7 @@ Edit your `pom.xml` file to add settings to the maven-compiler-plugin:
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
-        <version>3.10.1</version>
+        <version>3.11.0</version>
         <configuration>
           <source>8</source>
           <target>8</target>


cushon:
I think this is explained by https://issues.apache.org/jira/browse/MCOMPILER-457, when enables `<showWarnings>` by default in `3.11.0`.

With your repro if I switch back to `3.10.1` and set `<showWarnings>true</showWarnings>` I see warnings:

```diff
diff --git a/pom.xml b/pom.xml
index fab2fc8..437a101 100644
--- a/pom.xml
+++ b/pom.xml
@@ -20,11 +20,12 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
-        <version>3.11.0</version>
+        <version>3.10.1</version>
         <configuration>
           <source>11</source>
           <target>11</target>
           <encoding>UTF-8</encoding>
+          <showWarnings>true</showWarnings>
           <compilerArgs>
             <arg>-XDcompilePolicy=simple</arg>
             <arg>-Xlint:all</arg>

```

cushon (approved):

msridhar:
@cushon does something else need to happen for this PR to land?  No hurry on my end, just curious if a step got missed.

cushon:
> @cushon does something else need to happen for this PR to land? No hurry on my end, just curious if a step got missed.

The automation to import it had gotten stuck, I have it a nudge. It should land in the next day or so.

PR #4228, opened 2024-01-01
Add support for specifying badEnclosingTypes for `BadImport` via flags

hisener:
Following the discussions in https://github.com/google/error-prone/pull/2236, this PR introduces the `BadImport:BadEnclosingTypes` flag to disallow nested types of specified enclosing type.

This PR also takes the flag into account in `UnnecessarilyFullyQualified` to suggest importing `EnclosingType.NestedType` when possible.

For instance, when `-XepOpt:BadImport:BadEnclosingTypes=org.immutables.value.Value` is set, it would suggest the following: 

```java
final class Test {
  @org.immutables.value.Value.Immutable
  abstract class AbstractType {}
}
```
->
```java
final class Test {
  @Value.Immutable
  abstract class AbstractType {}
}
```

Also, it won't require suppressing `UnnecessarilyFullyQualified` in the example in https://github.com/google/error-prone/pull/2236.

```java
import org.springframework.beans.factory.annotation.Value;

final class Test {
  Demo(@Value("some.property") String property) {}

  @org.immutables.value.Value.Immutable
  abstract class AbstractType {}
}
```

Closes https://github.com/google/error-prone/pull/2236.

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/UnnecessarilyFullyQualified.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/UnnecessarilyFullyQualified.java
@@ -33,6 +33,7 @@ import com.google.common.collect.SetMultimap;
 import com.google.common.collect.Table;
 import com.google.errorprone.BugPattern;
 import com.google.errorprone.BugPattern.SeverityLevel;
+import com.google.errorprone.ErrorProneFlags;
 import com.google.errorprone.VisitorState;
 import com.google.errorprone.bugpatterns.BugChecker.CompilationUnitTreeMatcher;
 import com.google.errorprone.fixes.SuggestedFix;
@@ -60,6 +61,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
+import javax.inject.Inject;
 import javax.lang.model.element.Name;
 
 /** Flags uses of fully qualified names which are not ambiguous if imported. */
@@ -70,6 +72,25 @@ public final class UnnecessarilyFullyQualified extends BugChecker
     implements CompilationUnitTreeMatcher {
 
   private static final ImmutableSet<String> EXEMPTED_NAMES = ImmutableSet.of("Annotation");
+
+  /**
+   * Exempted types that fully qualified name usages are acceptable for their nested types when
+   * importing the enclosing type is ambiguous.
+   *
+   * <p>Some types are meant to provide a namespace; therefore, imports for their nested types can
+   * be confusing.
+   *
+   * <p>For instance, unlike its name suggests, {@code org.immutables.value.Value.Immutable} is used
+   * to generate immutable value types, and its import can be misleading. So, importing {@code
+   * org.immutables.value.Value} and using {@code @Value.Immutable} is more favorable than importing
+   * {@code org.immutables.value.Value.Immutable} and using {@code @Immutable}.
+   */
+  private final ImmutableSet<String> exemptedTypes;
+
+  @Inject
+  UnnecessarilyFullyQualified(ErrorProneFlags errorProneFlags) {
+    this.exemptedTypes = errorProneFlags.getSetOrEmpty("BadImport:BadEnclosingTypes");
+  }
 
   @Override
   public Description matchCompilationUnit(CompilationUnitTree tree, VisitorState state) {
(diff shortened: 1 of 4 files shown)


core/src/test/java/com/google/errorprone/bugpatterns/BadImportTest.java
```diff
@@ -388,4 +388,63 @@ public void doesNotMatchProtos() {
...
+        .setArgs("-XepOpt:BadImport:BadEnclosingTypes=com.google.common.collect.ImmutableList")
+        .addSourceLines(
+            "Test.java",
+            "import static com.google.common.collect.ImmutableList.toImmutableList;",
+            "import com.google.common.collect.ImmutableList;",
+            "import java.util.stream.Collector;",
+            "",
+            "class Test {",
+            "  // BUG: Diagnostic contains: ImmutableList.toImmutableList()",
+            "  Collector<?, ?, ImmutableList<Object>> immutableList = toImmutableList();",
```
hisener, line 446: Is disallowing static import desirable? Should we change the behavior only for NestedTypes? :thinking:  https://github.com/google/error-prone/blob/63cf19274e382089693102a19bf31d7dbe791807/core/src/main/java/com/google/errorprone/bugpatterns/BadImport.java#L107-L109

core/src/main/java/com/google/errorprone/bugpatterns/UnnecessarilyFullyQualified.java
```diff
@@ -138,7 +159,8 @@ private void handle(TreePath path) {
         if (!isFullyQualified(tree)) {
           return;
         }
-        if (BadImport.BAD_NESTED_CLASSES.contains(tree.getIdentifier().toString())) {
+        if (exemptedTypes.contains(tree.getExpression().toString())
```
cushon, line 162: I think this is fine because of the guarantees made by the checks in `isFullyQualified`, but we generally try to avoid using `Tree#toString`. There's some performance cost for larger trees, and the string representation isn't stable.

What do you think about something like:

```java
if (!BadImport.BAD_NESTED_CLASSES.contains(tree.getIdentifier().toString())) {
  return;
}
if (!(tree.getExpression() instanceof MemberSelectTree)) {
  return;
}
Symbol sym = getSymbol(tree.getExpression());
if (!(sym instanceof ClassSymbol)) {
  return;
}
if (exemptedTypes.contains(((ClassSymbol) sym).getQualifiedName().toString())) {
  return;
}
handle(new TreePath(path, tree.getExpression()));
```

core/src/test/java/com/google/errorprone/bugpatterns/BadImportTest.java
cushon: The current behaviour of treating static imports consistently seems OK to me. Obviously `toImmutableList` is just an example, in general if `com.example.Foo` is supposed to provide a namespace and not have its members be imports, it seems OK to discourage both `import com.example.Foo.Bar` and `import static com.example.Foo.baz`.

cushon (changes requested):
LGTM aside from the comment about `Tree#toString`. Thanks for the PR!

core/src/main/java/com/google/errorprone/bugpatterns/UnnecessarilyFullyQualified.java
hisener: :eyes: We shouldn't directly return if it's not a bad nested class, but I think I got your point.

I tried to refactor in https://github.com/google/error-prone/pull/4228/commits/67aa36d36647ac776fdeed3dbf0711cabbb03f54, PTAL.

cushon (approved):

core/src/main/java/com/google/errorprone/bugpatterns/UnnecessarilyFullyQualified.java
```diff
@@ -160,6 +179,23 @@ && getSymbol(tree.getExpression()) instanceof ClassSymbol) {
         treePaths.add(path);
       }
 
+      private boolean isBadNestedClass(MemberSelectTree tree) {
+        return BadImport.BAD_NESTED_CLASSES.contains(tree.getIdentifier().toString());
+      }
+
+      private boolean isExemptedEnclosingType(MemberSelectTree tree) {
```
hisener, line 186: Does it make sense to early-exit if the flag is not set? Or, is the cost negligible? :thinking: 

```suggestion
      private boolean isExemptedEnclosingType(MemberSelectTree tree) {
        if (exemptedEnclosingTypes.isEmpty()) {
          return false;
        }
```

cushon: Either way is fine with me, I think the cost is probably negligible.

hisener: Leaving it as-is then :+1:

PR #4314, opened 2024-03-09
Add the 'compile' goal for 'compile-java9'

cushon:
and update to a newer version of bnd for multi-release jar support, see https://github.com/bndtools/bnd/issues/2227.

### DIFF
--- a/annotations/pom.xml
+++ b/annotations/pom.xml
@@ -64,6 +64,9 @@
           </execution>
           <execution>
             <id>compile-java9</id>
+            <goals>
+              <goal>compile</goal>
+            </goals>
             <configuration>
               <source>9</source>
               <target>9</target>
--- a/annotations/pom.xml
+++ b/annotations/pom.xml
@@ -85,6 +85,9 @@
               <Multi-Release>true</Multi-Release>
             </manifestEntries>
           </archive>
+          <excludes>
+            <exclude>/META-INF/versions/9/com/**/*.class</exclude>
+          </excludes>
         </configuration>
       </plugin>
     </plugins>
--- a/annotations/src/main/java/module-info.java
+++ b/annotations/src/main/java/module-info.java
@@ -15,7 +15,6 @@
  */
 
 open module com.google.errorprone.annotation {
-  requires java.base;
   requires java.compiler;
 
   exports com.google.errorprone.annotations;
--- a/pom.xml
+++ b/pom.xml
@@ -139,7 +139,7 @@
       <plugin>
         <groupId>biz.aQute.bnd</groupId>
         <artifactId>bnd-maven-plugin</artifactId>
-        <version>7.0.0</version>
+        <version>6.4.0</version>
         <executions>
           <execution>
             <id>generate-OSGi-manifest</id>
--- a/pom.xml
+++ b/pom.xml
@@ -139,7 +139,7 @@
       <plugin>
         <groupId>biz.aQute.bnd</groupId>
         <artifactId>bnd-maven-plugin</artifactId>
-        <version>6.4.0</version>
+        <version>7.0.0</version>
         <executions>
           <execution>
             <id>generate-OSGi-manifest</id>
--- a/pom.xml
+++ b/pom.xml
@@ -150,7 +150,7 @@
               <bnd><![CDATA[
                 Bundle-SymbolicName: com.google.$<replacestring;$<replacestring;${project.artifactId};^error_prone;errorprone>;_;.>
                 Automatic-Module-Name: $<Bundle-SymbolicName>
-                -exportcontents: com.google.errorprone*
+                -exportcontents: com.google.errorprone*,!META-INF.*
                 -noextraheaders: true
                 -removeheaders: Private-Package
                 -fixupmessages: ^Classes found in the wrong directory: .*
--- a/pom.xml
+++ b/pom.xml
@@ -153,6 +153,7 @@
                 -exportcontents: com.google.errorprone*
                 -noextraheaders: true
                 -removeheaders: Private-Package
+                -fixupmessages: ^Classes found in the wrong directory: .*
               ]]></bnd>
             </configuration>
           </execution>


cushon:
@sgammon does this look reasonable to you?

Without this change I'm not seeing the `compile-java9` step added in https://github.com/google/error-prone/commit/0e9536471b6b8e7e03d29584eceb8a0e0f3240d7 get executed, and the `META-INF/versions/9/` classes aren't ending up in the artifact.

cushon:
Hmm, the version of bnd that supports mr-jars requires JDK 17:

```
Error:  Failed to execute goal biz.aQute.bnd:bnd-maven-plugin:7.0.0:bnd-process (generate-OSGi-manifest) on project error_prone_parent: Execution generate-OSGi-manifest of goal biz.aQute.bnd:bnd-maven-plugin:7.0.0:bnd-process failed: Unable to load the mojo 'bnd-process' in the plugin 'biz.aQute.bnd:bnd-maven-plugin:7.0.0' due to an API incompatibility: org.codehaus.plexus.component.repository.exception.ComponentLookupException: aQute/bnd/maven/plugin/BndMavenPlugin has been compiled by a more recent version of the Java Runtime (class file version 61.0), this version of the Java Runtime only recognizes class file versions up to 55.0
```

sgammon:
@cuson Whoops! I can't believe that wasn't running. I thought I verified the JAR output. I'll check locally with this change applied and verify it with Guava downstream.

Re/BND, the JAR is a MRJAR is assembled by the JAR step, and BND tools isn't aware of it. I don't think it _needs_ to be aware of it. There are BND directives for generating `module-info.java`, but we have a definition so we don't need to generate one.

As far as BND loading MRJARs, I know that support is lacking there too, but the classes under `META-INF/versions` should be identical anyway and only consist of `module-info.class`.

sgammon:
@cushon Adding this directive to the BND tools fixes the build issue with classes in the MRJAR root:

```
-fixupmessages: ^Classes found in the wrong directory: .*
```

The final block in root `pom.xml` is:
```xml
              <bnd><![CDATA[
                Bundle-SymbolicName: com.google.$<replacestring;$<replacestring;${project.artifactId};^error_prone;errorprone>;_;.>
                Automatic-Module-Name: $<Bundle-SymbolicName>
                -exportcontents: com.google.errorprone*
                -noextraheaders: true
                -removeheaders: Private-Package
                -fixupmessages: ^Classes found in the wrong directory: .*
              ]]></bnd>
```

I'm just starting my day, so I'll be around. If you want me to push a PR to fix this let me know

anthonyvdotbe:
@cushon While you're at it, please remove `requires java.base;` from [module-info.java](https://github.com/google/error-prone/blob/master/annotations/src/main/java/module-info.java). Every module requires `java.base`. Specifying it is like specifying `import java.lang.*;`.

sgammon:
@cushon A quick patch, I've verified it includes the classes and only the `module-info.class` (trying to prevent duplication of the actual classes)

```patch
diff --git a/annotations/pom.xml b/annotations/pom.xml
index 39a0627585..e68d412cbc 100644
--- a/annotations/pom.xml
+++ b/annotations/pom.xml
@@ -64,6 +64,9 @@
           </execution>
           <execution>
             <id>compile-java9</id>
+            <goals>
+              <goal>compile</goal>
+            </goals>
             <configuration>
               <source>9</source>
               <target>9</target>
@@ -82,6 +85,9 @@
               <Multi-Release>true</Multi-Release>
             </manifestEntries>
           </archive>
+          <excludes>
+            <exclude>/META-INF/versions/9/com/**/*.class</exclude>
+          </excludes>
         </configuration>
       </plugin>
     </plugins>
index f7cad3c75b..7c0fc99120 100644
--- a/pom.xml
+++ b/pom.xml
@@ -153,6 +153,7 @@
                 -exportcontents: com.google.errorprone*
                 -noextraheaders: true
                 -removeheaders: Private-Package
+                -fixupmessages: ^Classes found in the wrong directory: .*
               ]]></bnd>
             </configuration>
           </execution>
```

sgammon:
@anthonyvdotbe I was under the impression that _all_ Java modules had to depend on `java.base`? That's why it's there. Is that not the case?

sgammon:
I'm seeing the JDK 11 failure @cushon. I think I have a fix for that too.

sgammon:
Okay. Thank you @anthonyvdotbe, I learned something today.

The `java.base` module is not required to be stated in `module-info.java`. The JDK 11 failure happens because BND tries to load those classes, but we can make them inert instead:

This patch
- Removes `java.base`
- Fixes compile goal
- Excludes Java 9 classes but includes `module-info.class`
- Excludes `META-INF` from BND (see `!META-INF...`)
- Applies the `fixupmessages` to BND

```patch
diff --git a/annotations/pom.xml b/annotations/pom.xml
index 39a0627585..e68d412cbc 100644
--- a/annotations/pom.xml
+++ b/annotations/pom.xml
@@ -64,6 +64,9 @@
           </execution>
           <execution>
             <id>compile-java9</id>
+            <goals>
+              <goal>compile</goal>
+            </goals>
             <configuration>
               <source>9</source>
               <target>9</target>
@@ -82,6 +85,9 @@
               <Multi-Release>true</Multi-Release>
             </manifestEntries>
           </archive>
+          <excludes>
+            <exclude>/META-INF/versions/9/com/**/*.class</exclude>
+          </excludes>
         </configuration>
       </plugin>
     </plugins>
diff --git a/annotations/src/main/java/module-info.java b/annotations/src/main/java/module-info.java
index 72215b7677..4c9077dd52 100644
--- a/annotations/src/main/java/module-info.java
+++ b/annotations/src/main/java/module-info.java
@@ -15,7 +15,6 @@
  */
 
 open module com.google.errorprone.annotation {
-  requires java.base;
   requires java.compiler;
   exports com.google.errorprone.annotations;
   exports com.google.errorprone.annotations.concurrent;
index f7cad3c75b..3dc704b4f8 100644
--- a/pom.xml
+++ b/pom.xml
@@ -150,9 +150,10 @@
               <bnd><![CDATA[
                 Bundle-SymbolicName: com.google.$<replacestring;$<replacestring;${project.artifactId};^error_prone;errorprone>;_;.>
                 Automatic-Module-Name: $<Bundle-SymbolicName>
-                -exportcontents: com.google.errorprone*
+                -exportcontents: com.google.errorprone*,!META-INF.*
                 -noextraheaders: true
                 -removeheaders: Private-Package
+                -fixupmessages: ^Classes found in the wrong directory: .*
               ]]></bnd>
             </configuration>
           </execution>
```

cpovirk (approved):

cushon:
Thanks @sgammon! I pushed another commit that includes the additional changes.

PR #4318, opened 2024-03-12
Make patching API public

rsalvador:
to be used by bazel’s error prone plugin: https://github.com/bazelbuild/bazel/pull/21640

### DIFF
--- a/check_api/src/main/java/com/google/errorprone/ErrorProneAnalyzer.java
+++ b/check_api/src/main/java/com/google/errorprone/ErrorProneAnalyzer.java
@@ -22,7 +22,9 @@ import static com.google.common.base.Verify.verify;
 
 import com.google.common.base.Supplier;
 import com.google.common.base.Suppliers;
+import com.google.common.collect.ImmutableSet;
 import com.google.errorprone.BugPattern.SeverityLevel;
+import com.google.errorprone.RefactoringCollection.RefactoringResult;
 import com.google.errorprone.scanner.ErrorProneScannerTransformer;
 import com.google.errorprone.scanner.ScannerSupplier;
 import com.google.errorprone.util.ASTHelpers;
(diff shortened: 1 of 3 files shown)


msridhar:
@cushon this change is being successfully used at Uber to enable Error Prone auto-patching to work with Bazel, along with https://github.com/bazelbuild/bazel/pull/21640.  I have a follow-on Error Prone fix that makes auto-patching also work for plug-in checks, which I'm happy to contribute as a PR as well (though it makes more sense to land after this one).

Independent of the Bazel change, this PR seems straightforward and could be helpful for any build system trying to integrate Error Prone.  So, it'd be great if someone from the EP team could review it sometime.  Thanks!

check_api/src/main/java/com/google/errorprone/ErrorProneAnalyzer.java
```diff
@@ -57,6 +61,73 @@ public class ErrorProneAnalyzer implements TaskListener {
...
+    refactoringCollection[0] = RefactoringCollection.refactor(epOptions.patchingOptions(), context);
+
+    // Refaster refactorer or using builtin checks
+    CodeTransformer codeTransformer =
+        epOptions
+            .patchingOptions()
+            .customRefactorer()
+            .or(
+                () -> {
+                  ScannerSupplier toUse =
```
cushon, line 81: This un-does some changes from https://github.com/google/error-prone/commit/9da2d5580e3939f97ef2e91278b330a56b5ed1fe compared to the original code, I wonder if that was just an oversight when picking up 9da2d5580e3939f97ef2e91278b330a56b5ed1fe?

rsalvador: Yes, was an oversight, re-pushed the changes using the latest code.

cushon (approved):

msridhar:
@cushon are any other steps required for this one to land?  Thanks for the help!

PR #4384, opened 2024-05-01
Add DoNotMockRecords to warn against mocking record data objects

ralstonba:
## Overview

>Enhance the Java programming language with [records](http://cr.openjdk.java.net/~briangoetz/amber/datum.html), which are classes that act as transparent carriers for immutable data. Records can be thought of as nominal tuples.

[Records](https://openjdk.org/jeps/395) are _typically_ used as simple wrappers of data without any methods or complex behavior requiring external dependencies or expensive IO operations. As such, they should (almost) never be mocked, and instead a real instance should be used instead. 

## Details
The class hierarchy has changed from 
```
|- AbstractMockChecker
  |-- DoNotMockAutoValue
  |-- DoNotMockChecker
```
to 
```
|- AbstractMockChecker
  |-- DoNotMockRecords
  |-- AbstractAnnotationMockChecker
    |-- DoNotMockAutoValue
    |-- DoNotMockChecker
```

This was because `AbstractMockChecker` assumed that only annotations would be used in determining that a class should not be mocked. The majority of `AbstractAnnotationMockChecker` is a straight lift and shift of the original `AbstractMockChecker`.

## Open Questions
- `ElementKind.RECORD` results in an IDE error in Intellij as it is annotated with `@since 16`, suggesting that the language level of the module be changed. This does not cause issues with complilation, but as I am entirely new to projects that deal with the AST I am wondering if there is a better approach here.

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/AbstractAnnotationMockChecker.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/AbstractAnnotationMockChecker.java
@@ -1,0 +1,151 @@
+/*
+ * Copyright 2024 The Error Prone Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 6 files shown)


ralstonba:
Signed CLA

kluever:
@eamonnmcmanus said

> both `@Mock MyRecord foo;` and `MyRecord foo = mock(MyRecord.class)` trigger the Error Prone `CannotMockFinalClass` check:

```
error: [CannotMockFinalClass] Mockito cannot mock final classes
  private @Mock MyRecord myRecord;
                         ^
    (see https://errorprone.info/bugpattern/CannotMockFinalClass)
```

```
error: [CannotMockFinalClass] Mockito cannot mock final classes
  private MyRecord myRecord = mock(MyRecord.class);
                                  ^
    (see https://errorprone.info/bugpattern/CannotMockFinalClass)
```

Therefore I don't think we need to do anything further here.

Issue #4392, opened 2024-05-10
Suggest to use `Double.compare` or `Float.compare` when ObjectEqualsForPrimitives is triggered on floating point primitives


findepi:
for a class like
```java
public class DoubleLiteral
{
    private final double value;

    public DoubleLiteral(double value) { this.value = value; }

    @Override
    public boolean equals(Object o)
    {
        if (this == o) {
            return true;
        }
        if (o == null || getClass() != o.getClass()) {
            return false;
        }
        DoubleLiteral that = (DoubleLiteral) o;
        return Objects.equals(value, that.value);
    }

    @Override
    public int hashCode()
    {
        return Objects.hash(value);
    }
}
```

`ObjectEqualsForPrimitives`' current suggestion is:

```
DoubleLiteral.java:[37,30] [ObjectEqualsForPrimitives] Avoid unnecessary boxing by using plain == for primitive types.
[ERROR]     (see https://errorprone.info/bugpattern/ObjectEqualsForPrimitives)
[ERROR]   Did you mean 'return (value == that.value);'?
```

That changes semantics -- different result when both compared values are NaN. Object Equals should be value based and generally should compare NaNs as equal.
It's probably worth noting that equals generated for Java record classes also compares NaN values as equal (as if Objects.equals was used).

Therefore `ObjectEqualsForPrimitives` should not suggest `==`, but it should suggest `Double.compare(...) == 0` instead.

PR #4394, opened 2024-05-14
Suggest Double.compare in ObjectEqualsForPrimitives

findepi:
`Objects.equal(a_double, another_double)` and `a_double == another_double` have different semantics (different result when both compared values are NaN). Suggest replacement using `Double.compare` to avoid changing code semantics. This is especially relevant when implementing `equals` methods.

Fixes https://github.com/google/error-prone/issues/4392

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/ObjectEqualsForPrimitives.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/ObjectEqualsForPrimitives.java
@@ -29,7 +29,12 @@ import com.google.errorprone.fixes.Fix;
 import com.google.errorprone.fixes.SuggestedFix;
 import com.google.errorprone.matchers.Description;
 import com.google.errorprone.matchers.Matcher;
+import com.google.errorprone.util.ASTHelpers;
+import com.sun.source.tree.ExpressionTree;
 import com.sun.source.tree.MethodInvocationTree;
+import com.sun.tools.javac.code.Type;
+
+import javax.lang.model.type.TypeKind;
 
 /**
  * Check for usage of {@code Objects.equal} on primitive types.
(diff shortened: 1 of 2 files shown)


findepi:
@cushon can you please approve the build to run?

cc @msridhar you might be interesting in this in context of automated code fixes.

cushon:
We have some internal guidance to use `Double.toLongBits` when implementing exact equality on doubles if the values could be `NaN`, instead of `Double.compare`. What do you think of that alternative?

findepi:
@cushon thank you for your review!

> We have some internal guidance to use `Double.toLongBits` when implementing exact equality on doubles if the values could be `NaN`, instead of `Double.compare`. What do you think of that alternative?

`Double.doubleToRawLongBits` is obviously not a suitable replacement, since it compares two different NaN representations as numerically different, while `Double.equals(Object)` compares them as equal (and so does `Double.compare(double, double)`.

`Double.doubleToLongBits` looks as a good alternative & may perform better, changed the PR, thanks!

Would you mind triggering the build again?


> I'm also wondering if we should exclude `double` and `float` from this change, since it's just a micro-optimization, and there's a readability tradeoff with using `Double.compare` or `Double.toLongBits`

i am not sure i understand. Can you please rephrase?

cushon:
> > I'm also wondering if we should exclude `double` and `float` from this change, since it's just a micro-optimization, and there's a readability tradeoff with using `Double.compare` or `Double.toLongBits`
> 
> i am not sure i understand. Can you please rephrase?

I was wondering about having ObjectEqualsForPrimitives not emit findings for `double` and `float`. There's a tradeoff between `Objects.equal(a_double, another_double)` and `Double. toLongBits(a_double) == Double. toLongBits(another_double)`, the latter avoids some boxing and is theoretically faster, but the first is potentially more readable.

findepi:
Sure, I can do that.

findepi:
@cushon Would you mind triggering the build again?

cushon (approved):

findepi:
@cushon thanks for approval.
I see the [JDK EA on ubuntu-latest](https://github.com/google/error-prone/actions/runs/9106991750/job/25043381422?pr=4394#logs) build is failing. I don't know how to fix it.

cushon:
JDK EA is also failing at head, there are some breaking changes we'll need to prepare for before JDK 23 is released, but it doesn't need to be a blocker for this

findepi:
oh, and thanks for fixing the commit message.

cushon:
> thanks for the merge!

sure thing!

> oh, and thanks for fixing the commit message.

@cpovirk gets the credit for that part :)

PR #4429, opened 2024-06-10
Check for chance of class init deadlock with non-nested subclass

tkindy:
This PR fixes a bug with the `ClassInitializationDeadlock` check to find chances for a class initialization deadlock when the subclass isn't nested within the superclass. This situation [can still produce a deadlock](https://gist.github.com/tkindy/a83ec86c95bb7d7119f6b208d239ea94).

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/ClassInitializationDeadlock.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/ClassInitializationDeadlock.java
@@ -144,7 +144,7 @@ public class ClassInitializationDeadlock extends BugChecker implements BugChecke
         if (!use.isSubClass(classSymbol, state.getTypes())) {
           return;
         }
-        if (!isStatic(use)) {
+        if (use.isEnclosedBy(classSymbol) && !isStatic(use)) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/google/errorprone/bugpatterns/ClassInitializationDeadlock.java
```diff
@@ -144,7 +144,7 @@ private void handle(ExpressionTree tree) {
         if (!use.isSubClass(classSymbol, state.getTypes())) {
           return;
         }
-        if (!isStatic(use)) {
+        if (use.isEnclosedBy(classSymbol) && !isStatic(use)) {
           // Nested inner classes implicitly take the enclosing instance as a constructor parameter,
           // and can't be initialized without first initializing their containing class.
           return;
```
tkindy, line 150: From the comment, it's clear the intention was to only match situations where the subclass is a non-static nested class of the superclass. But it's overly broad; any class that isn't a static inner class (since only inner classes can be marked `static`) will be matched by the current conditional. First checking if the class is enclosed by the defining class makes the conditional true to the comment.

cushon:
Thanks for catching this!

cushon (approved):

PR #4487, opened 2024-07-20
Add DefaultLocale check to discourage use of the default locale

### LINKED ISSUE #632: BugPattern for using the default system Locale
Similar to http://errorprone.info/bugpattern/DefaultCharset, it seems reasonable to have a warning of using the default system Locale, e. g. in `String.format()`.

tbroyer:
Fixes #632


cushon (commented):
Thanks very much for this, it looks great.

I left a few minor comments, I'm going to run it over Google's codebase and see if that turns up any interesting false positives or other issues, and then will plan on getting this merged soon.

core/src/main/java/com/google/errorprone/bugpatterns/DefaultLocale.java
```diff
@@ -0,0 +1,698 @@
...
+
+  /**
+   * Should only refactor String.format() and similar methods' invocations where specifiers aren't
+   * locale-dependent. For %s this includes checking for non-Formattable arguments. Format strings
+   * (first argument) as variables or constants are excluded from refactoring.
+   */
+  private boolean shouldRefactorStringFormat(
+      ExpressionTree pattern, List<? extends ExpressionTree> arguments, VisitorState state) {
+    if (!(pattern instanceof LiteralTree)) return true;
+    if (!onlyContainsSpecifiersInAllowList((String) ((LiteralTree) pattern).getValue()))
```
cushon, line 464: This is only matching on string literals, not on other compile-time constant expressions like static final constant fields.

The following will handle compile-time constant expressions:

```
String patternValue = ASTHelpers.constValue(pattern, String.class)
if (patternValue != null) { ... }
```

core/src/main/java/com/google/errorprone/bugpatterns/DefaultLocale.java
```diff
@@ -0,0 +1,698 @@
...
+  private void dateFormatGetInstanceFixes(
+      Description.Builder description, MethodInvocationTree tree, VisitorState state) {
+    description.addFix(dateFormatGetInstanceFix(tree, state, LocaleFix.ROOT_FIX));
+    description.addFix(dateFormatGetInstanceFix(tree, state, LocaleFix.DEFAULT_LOCALE_FIX));
+    description.addFix(dateFormatGetInstanceFix(tree, state, LocaleFix.DEFAULT_FORMAT_LOCALE_FIX));
+  }
+
+  private Fix dateFormatGetInstanceFix(
+      MethodInvocationTree tree, VisitorState state, LocaleFix localeFix) {
+    // XXX: handle case of statically imported DateFormat.getInstance?
```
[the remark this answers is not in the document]
tbroyer: I finally used `renameMethodInvocation` directly, `merge`d into the suggested fix that handles the arguments.

core/src/main/java/com/google/errorprone/bugpatterns/DefaultLocale.java
```diff
@@ -0,0 +1,698 @@
...
+  @VisibleForTesting
+  static boolean onlyContainsSpecifiersInAllowList(String pattern) {
+    var noSpecifierFormatBase = SPECIFIER_ALLOW_LIST_REGEX.matcher(pattern).replaceAll("");
+    // If it still has a specifier after the replacement, it means that it was not on the allowlist.
+    return !noSpecifierFormatBase.contains("%");
+  }
+
+  private boolean containsSomeFormattableArgument(
+      List<? extends ExpressionTree> arguments, VisitorState state) {
+    return arguments.stream().anyMatch(tree -> mightBeFormattable(tree, state));
```
cushon, line 479: I noticed this is reporting errors for any `String.format` arguments that implement an interface (the example I noticed was `List<String>`), since any interface is castable to `Formattable`. Was that deliberate?

I think at Google we probably won't apply that approach as-is, because it would result in a large number of findings and false positives. I think there are a few ways to reconcile that. We could make the check less strict, or add configuration to control how strict it is.

core/src/main/java/com/google/errorprone/bugpatterns/DefaultLocale.java
cushon: A related question is the behaviour if the format string isn't available at compile time. The current behaviour rejects that, I mentioned the possibility of allowing it in the bug. It comes down to a tradeoff between soundness and false negatives.

core/src/main/java/com/google/errorprone/bugpatterns/DefaultLocale.java
tbroyer: Ah yes, that's going to match quite wide, but it's the "most correct" kind of match, right?
I mean, if you have `List<String>` it might actually be an instance of `class L<T> implements List<T>, Formattable`, that _could_ then be locale-dependent.

I suppose there are ways to detect a bunch of cases where this cannot happen (e.g. a local variable `List<String> list = new ArrayList<>()` is guaranteed to not implement `Formattable`) but this is beyond my skills I'm afraid.

cushon: Yes, agree it's the most correct option. I see a tradeoff between correctness and false positives here. I think it's reasonable to want the check to be as correct as possible, and e.g. report a match for `String.format("%s", someInterface)`. Partly I wanted to double-check: for your usage, do you want the most correct/sound version of the check, even if that results in false positives?

For Google's codebase the current version reports a lot of findings, and the ratio of `String.format` to everything else is about 9:1. I'm not sure we'll want to use the most correct version of the check. But that's also something I can look at as a follow-up. If you want the stricter version I'm happy to get this merged and then consider maybe adding flags to make it more lenient or something as a follow-up.

What do you think?

tbroyer: I have no idea how much Formattable is being used in the wild, and whether I might be affected by this (whether things I format implement Formattable or not, I definitely never implemented Formattable myself ever). Off the cuff I'd say the risk is rather low so I think I'd be OK with a more lenient check; but on the other hand it's not that hard to provide an explicit locale (even if `Locale.getDefault()`) "just in case".

So maybe make it strict by default (even if "too strict") and have a flag to make it more lenient?
(or if you prefer, start by making it lenient by default, with a flag to make it stricter, then later –possibly only after changes to have fewer false positives– switch the default behavior, with a flag to get back to the more lenient check, and eventually remove the flag)

Fwiw, I'm currently using Forbidden APIs which is even stricter as it flags all `String.format(String,Object...)` invocations irrespective of the pattern being used, so I'm already providing an explicit locale "just in case" everywhere, and even the strict behavior here would be better than Forbidden APIs.

core/src/main/java/com/google/errorprone/bugpatterns/DefaultLocale.java
tbroyer: When changing to `ASTHelpers.constValue`, I actually made it to allow non-constant patterns. I added a test, and a TODO to possibly add a flag to make it stricter.

core/src/main/java/com/google/errorprone/bugpatterns/DefaultLocale.java
cushon: I took a quick look at `Formattable` usage and it seems to be very rare. For Google's internal usage we might be OK not worrying about that case to reduce the number of noise from adding default locales just in case.

Anyway, I'm happy to merge the current version of this. Partly I wanted to confirm how strict a check you wanted for your usage. I'll think more about what I want to enable at Google, and if that ends up being slightly different we can express the difference with flags.

core/src/main/java/com/google/errorprone/bugpatterns/DefaultLocale.java
cushon: Thanks!

core/src/main/java/com/google/errorprone/bugpatterns/DefaultLocale.java
```diff
@@ -0,0 +1,698 @@
...
+import java.io.PrintWriter;
+import java.util.Formattable;
+import java.util.List;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/** A {@link BugChecker}; see the associated {@link BugPattern} annotation for details. */
+@BugPattern(
+    summary =
```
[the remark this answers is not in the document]
cushon: @tbroyer thoughts on this? This can also happen as a follow up if you don't have time right now.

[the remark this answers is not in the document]
tbroyer: Might I let you (or someone else) do it? :innocent: 

I started drafting something but couldn't clearly articulate my thoughts :disappointed:

core/src/main/java/com/google/errorprone/bugpatterns/DefaultLocale.java
```diff
@@ -0,0 +1,703 @@
...
+    }
+    return buildDescription(tree).build();
+  }
+
+  private Description prependLocales(
+      MethodInvocationTree tree, VisitorState state, LocaleFix... localeFixes) {
+    return prependLocales(tree, tree.getMethodSelect(), tree.getArguments(), state, localeFixes);
+  }
+
+  private Description prependLocales(
```
cushon, line 627: FYI I think this method is unused

PR #4521, opened 2024-08-05
Multi-Release jar experiments

cushon:


check_api/src/main/java/com/google/errorprone/util/ErrorProneSignatureGenerator.java
```diff
@@ -0,0 +1,57 @@
...
+  protected void append(Name name) {
+    buffer.appendName(name);
+  }
+
+  @Override
+  public String toString() {
+    // We could use buffer.toName(Names), but we want a string anyways and this
+    // avoids plumbing a Context or instances of Names through.
+    // Names always uses UTF-8 internally.
+    return new String(Arrays.copyOf(buffer.elems, buffer.length), UTF_8);
```
Stephan202, line 55: I recall once running a benchmark, on the basis of which I concluded that the JVM may be smart enough to conclude that in a scenario such as this one, the `String` constructor doesn't need to make a defensive copy of the provided array, as it'll hold the only remaining reference. Still, perhaps it's nicer to instead do:
```suggestion
    return new String(buffer.elems, 0, buffer.length, UTF_8);
```
(And likewise for the Java 24 variant below.)

Stephan202 (commented):
Just a drive-by comment :)

check_api/src/main/java/com/google/errorprone/util/ErrorProneSignatureGenerator.java
cushon: Thanks, good point. I had separately been considering refactoring this to pass through a `VisitorState` to these methods instead of `Type`, which would allow using `buffer.toName(Names)` as the comment here notes. I'll take a look at finishing that.

Issue #3756, opened 2023-02-04
Consider using Multi-Release jars for Error Prone


cushon:
This bug tracks trying to use [Multi-Release jar files](https://openjdk.org/jeps/238) for Error Prone.

The main use-case is that some checks need to handle AST nodes that not available in all supported JDK versions. Currently we're mostly using reflection to work around that, with mrjars we could move some of that logic behind an internal API and provide implementations that used the new AST nodes for JDK versions that supported them, avoiding the reflection. (e.g. #3737, https://github.com/google/error-prone/pull/3610#discussion_r1051486417)

One of the barriers to doing this is figuring out the best way to produce mrjars with maven. (If anyone reading this has relevant experience, suggestions are welcome!)

cushon:
Some WIP on this in https://github.com/google/error-prone/pull/4521

* maven's `<multiReleaseOutput>` assumes `--release`, which javac doesn't allow combing with `--add-exports=`. Maven doesn't appear to support `--system`. So the only option for compiling version-specific code seems to be running on the corresponding JDK's javac, which can be done with `<jdkToolchain><version>...</version></jdkToolchain>` and `<fork>`
* this requires having a maven toolchain registered for JDK EAs. `actions/setup-java` registers toolchains but doesn't download EAs from https://jdk.java.net/24/, `oracle-actions/setup-java` does the latter but not the former. There are some breadcrumbs here that might work: https://github.com/actions/setup-java/blob/main/docs/advanced-usage.md#modifying-the-toolchain-vendor-for-jdks

cpovirk:
- I just recently [set up Guava to automatically download toolchains _during the build_ with `toolchains-maven-plugin`](https://github.com/google/guava/pull/7333) (not to be confused with `maven-toolchains-plugin`, which we also now use). I didn't test downloading an EA JDK, but I see versions like "24-ea+7" in [this list](https://api.foojay.io/disco/v3.0/distributions) linked [from the plugin's GitHub page](https://github.com/linux-china/toolchains-maven-plugin?tab=readme-ov-file#how-to-get-a-list-of-all-supported-distributions-and-jdk-versions), so I would expect it to work. I do suspect that that is slower (_maybe_ ~30 seconds in the case of Guava, for which we're downloading 1-2 toolchains during the build), but I saw varied enough numbers (discussed in internal issue 334979149) that I didn't draw firm conclusions.
[...]

PR #4586, opened 2024-09-21
Optimize VisitorState#getConstantExpression

Stephan202:
By avoiding a second pass over the string.

### DIFF
--- a/check_api/src/main/java/com/google/errorprone/VisitorState.java
+++ b/check_api/src/main/java/com/google/errorprone/VisitorState.java
@@ -51,6 +51,7 @@ import com.sun.tools.javac.tree.JCTree;
 import com.sun.tools.javac.tree.JCTree.JCCompilationUnit;
 import com.sun.tools.javac.tree.TreeMaker;
 import com.sun.tools.javac.util.Context;
+import com.sun.tools.javac.util.Convert;
 import com.sun.tools.javac.util.Name;
 import com.sun.tools.javac.util.Names;
 import com.sun.tools.javac.util.Options;
@@ -770,24 +771,20 @@ public class VisitorState {
    * Like {@link Elements#getConstantExpression}, but doesn't over-escape single quotes in strings.
    */
   public String getConstantExpression(Object value) {
-    String escaped = getElements().getConstantExpression(value);
-    if (value instanceof String) {
-      // Don't escape single-quotes in string literals
-      StringBuilder sb = new StringBuilder();
-      for (int i = 0; i < escaped.length(); i++) {
-        char c = escaped.charAt(i);
-        if (c == '\\' && i + 1 < escaped.length()) {
-          char next = escaped.charAt(++i);
-          if (next != '\'') {
-            sb.append(c);
-          }
-          sb.append(next);
-        } else {
-          sb.append(c);
-        }
-      }
-      return sb.toString();
+    if (!(value instanceof String str)) {
+      return getElements().getConstantExpression(value);
     }
-    return escaped;
+
+    // Don't escape single-quotes in string literals.
+    StringBuilder sb = new StringBuilder("\"");
+    for (int i = 0; i < str.length(); i++) {
+      char c = str.charAt(i);
+      if (c == '\'') {
+        sb.append('\'');
+      } else {
+        sb.append(Convert.quote(c));
+      }
+    }
+    return sb.append('"').toString();
   }
 }
--- a/check_api/src/main/java/com/google/errorprone/VisitorState.java
+++ b/check_api/src/main/java/com/google/errorprone/VisitorState.java
@@ -771,7 +771,7 @@ public class VisitorState {
    * Like {@link Elements#getConstantExpression}, but doesn't over-escape single quotes in strings.
    */
   public String getConstantExpression(Object value) {
-    if (!(value instanceof String str)) {
+    if (!(value instanceof CharSequence str)) {
       return getElements().getConstantExpression(value);
     }
 
--- a/core/src/main/java/com/google/errorprone/bugpatterns/RobolectricShadowDirectlyOn.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/RobolectricShadowDirectlyOn.java
@@ -73,7 +73,7 @@ public class RobolectricShadowDirectlyOn extends BugChecker implements MethodInv
     MethodSymbol symbol = getSymbol(parent);
     String argReplacement =
         Streams.concat(
-                Stream.of(state.getConstantExpression(symbol.getSimpleName().toString())),
+                Stream.of(state.getConstantExpression(symbol.getSimpleName())),
                 Streams.zip(
                     symbol.getParameters().stream(),
                     parent.getArguments().stream(),
--- a/core/src/main/java/com/google/errorprone/bugpatterns/flogger/FloggerArgumentToString.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/flogger/FloggerArgumentToString.java
@@ -351,7 +351,7 @@ public class FloggerArgumentToString extends BugChecker implements MethodInvocat
     if (!fixed) {
       return NO_MATCH;
     }
-    fix.replace(tree.getArguments().get(0), state.getConstantExpression(sb.toString()));
+    fix.replace(tree.getArguments().get(0), state.getConstantExpression(sb));
     return describeMatch(tree, fix.build());
   }
 
--- a/core/src/main/java/com/google/errorprone/bugpatterns/flogger/FloggerStringConcatenation.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/flogger/FloggerStringConcatenation.java
@@ -112,7 +112,7 @@ public class FloggerStringConcatenation extends BugChecker implements MethodInvo
         tree,
         SuggestedFix.replace(
             argument,
-            state.getConstantExpression(formatString.toString())
+            state.getConstantExpression(formatString)
                 + ", "
                 + formatArguments.stream().map(state::getSourceForNode).collect(joining(", "))));
   }
--- a/core/src/test/java/com/google/errorprone/VisitorStateTest.java
+++ b/core/src/test/java/com/google/errorprone/VisitorStateTest.java
@@ -106,6 +106,8 @@ public class VisitorStateTest {
     assertThat(visitorState.getConstantExpression("hello \n world"))
         .isEqualTo("\"hello \\n world\"");
     assertThat(visitorState.getConstantExpression('\'')).isEqualTo("'\\''");
+    assertThat(visitorState.getConstantExpression(new StringBuilder("hello ' world")))
+        .isEqualTo("\"hello ' world\"");
   }
 
   // The following is taken from ErrorProneJavacPluginTest. There may be an easier way.


check_api/src/main/java/com/google/errorprone/VisitorState.java
```diff
@@ -770,24 +771,20 @@ private static final class SharedState {
...
-        if (c == '\\' && i + 1 < escaped.length()) {
-          char next = escaped.charAt(++i);
-          if (next != '\'') {
-            sb.append(c);
-          }
-          sb.append(next);
-        } else {
-          sb.append(c);
-        }
+    if (!(value instanceof CharSequence str)) {
```
Stephan202, line 774: I added a second commit that enables a further optimization by accepting any `CharSequence`; this can be backed out if desired.

Stephan202 (commented):
I filed this PR as a result of work in PicnicSupermarket/error-prone-support#1138, where I found out that I overlooked `VisitorState#getConstantExpression` and instead reinvented the wheel.

check_api/src/main/java/com/google/errorprone/VisitorState.java
```diff
@@ -770,24 +771,20 @@ private static final class SharedState {
...
+    }
+
+    // Don't escape single-quotes in string literals.
+    StringBuilder sb = new StringBuilder("\"");
+    for (int i = 0; i < str.length(); i++) {
+      char c = str.charAt(i);
+      if (c == '\'') {
+        sb.append('\'');
+      } else {
+        sb.append(Convert.quote(c));
```
cushon, line 785: Incidentally I recently learned that `Convert.quote` does the thing we'd want now, but only for JDK 23+: https://github.com/openjdk/jdk/commit/144a08ee5092980859379c7664c8036eec1e264d

check_api/src/main/java/com/google/errorprone/VisitorState.java
cushon, line 774: Unfortunately we're not _quite_ ready to start using Java > 11 features in non-test code internally. I can remove the `instanceof` pattern when importing this, I'm just leaving a note here in case you wonder why it disappears.

cushon (approved):

check_api/src/main/java/com/google/errorprone/VisitorState.java
cushon: Er, and as a result of that change `Convert.quote(char)` was replaced with `Convert.quote(char, boolean)`, so this change doesn't work with the latest JDK EA builds.

I'm not sure what the best option is. We could use reflection or MR-JARs to call the method.

cushon (changes requested):
The `Convert.quote` part will need more work to be compatible with the latest JDK EA builds

check_api/src/main/java/com/google/errorprone/VisitorState.java
Stephan202: Ah, nice catch! I haven't looked at the MR-JAR setup yet, but indeed reflection should work. TBD whether I'll find some time today for a new pass.

check_api/pom.xml
```diff
@@ -181,6 +181,22 @@
...
+              <jdkToolchain>
+                <version>23</version>
+              </jdkToolchain>
+              <compileSourceRoots>
+                <compileSourceRoot>${basedir}/src/main/java23</compileSourceRoot>
+              </compileSourceRoots>
+              <!-- multiReleaseOutput requires setting release -->
+              <outputDirectory>${project.build.outputDirectory}/META-INF/versions/23</outputDirectory>
+            </configuration>
+          </execution>
```
Stephan202, line 199: A few things are of note here:
- Without `<goal>compile</goal>` this execution is skipped. This also means that the `java24` execution below is currently a no-op.
- This configuration doesn't override the inherited  `<source>17</source>` and `<target>17</target>` settings, though it could. (If so, we should do the same below.)
- I'm not sure each of these executions should use an exactly-matching toolchain JDK version. Perhaps it's better to use the latest JDK whenever possible, so as to minimize the number of JDKs required.

check_api/pom.xml
```diff
@@ -197,6 +213,17 @@
...
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-jar-plugin</artifactId>
+        <configuration>
+          <archive>
+            <manifestEntries>
+              <Multi-Release>true</Multi-Release>
+            </manifestEntries>
+          </archive>
+        </configuration>
+      </plugin>
```
Stephan202, line 226: Also had to add this to make sure that the class file bundled under `META-INF/versions/23` is picked up.

check_api/src/main/java/com/google/errorprone/ConstantStringExpressions.java
```diff
@@ -0,0 +1,45 @@
...
+
+package com.google.errorprone;
+
+import com.sun.tools.javac.util.Convert;
+
+/**
+ * @see VisitorState#getConstantExpression(Object)
+ */
+final class ConstantStringExpressions {
+  private ConstantStringExpressions() {}
```
Stephan202, line 25: Certainly open to approaches other than a special-purpose static utility class named `com.google.errorprone.ConstantStringExpressions`.

check_api/src/main/java/com/google/errorprone/ConstantStringExpressions.java
```diff
@@ -0,0 +1,45 @@
...
+
+  static String toConstantStringExpression(Object value, VisitorState state) {
+    if (!(value instanceof CharSequence)) {
+      return state.getElements().getConstantExpression(value);
+    }
+
+    // Don't escape single-quotes in string literals.
+    CharSequence str = (CharSequence) value;
```
Stephan202, line 33: Dropped the `instanceof` pattern matching as suggested [here](https://github.com/google/error-prone/pull/4586#discussion_r1769669576).

check_api/src/main/java23/com/google/errorprone/ConstantStringExpressions.java
```diff
@@ -0,0 +1,34 @@
...
+ * @see VisitorState#getConstantExpression(Object)
+ */
+final class ConstantStringExpressions {
+  private ConstantStringExpressions() {}
+
+  static String toConstantStringExpression(Object value, VisitorState state) {
+    return state
+        .getElements()
+        .getConstantExpression(
+            value instanceof CharSequence charSequence ? charSequence.toString() : value);
```
Stephan202, line 32: Here I did keep the `instanceof` pattern matching, because this code anyway only applies to JDK 23+, but if that anyway impacts the Google-internal setup, that can be changed, of course.

Stephan202 (commented):
Rebased and added a commit with a multi-release JAR approach. Locally this makes the build pass with JDK 25.ea.3-open.

I suspect that changes to `.github/workflows/ci.yml` are required, but as the GitHub Actions workflow isn't auto-triggered, I'm not sure :eyes:.

I'll cross-reference this PR in #3756.

check_api/src/main/java/com/google/errorprone/ConstantStringExpressions.java
cushon: We're finally able to use Java 21 features here internally, the earlier comment is obsolete.

check_api/src/main/java/com/google/errorprone/ConstantStringExpressions.java
cushon: This seems good for now, thanks.

check_api/pom.xml
cushon: > This also means that the java24 execution below is currently a no-op.

Huh, thanks for pointing that out. This is also used with an Bazel-based build that does the multi-release jar stuff differently, so maybe there are no uses (or test coverage) for the maven build and ErrorProneSignatureGenerator that require JDK 24 support.

> I'm not sure each of these executions should use an exactly-matching toolchain JDK version. Perhaps it's better to use the latest JDK whenever possible, so as to minimize the number of JDKs required.

The answer might be 'it depends'. For javac API changes, it's probably best to try to use an exact match, but that does require installing extra JDKs for the github actions setup. If the API changes once and is stable on newer versions we'd get away with a newer version, though.

cushon (approved):

check_api/src/main/java/com/google/errorprone/ConstantStringExpressions.java
```diff
@@ -0,0 +1,45 @@
...
+
+import com.sun.tools.javac.util.Convert;
+
+/**
+ * @see VisitorState#getConstantExpression(Object)
+ */
+final class ConstantStringExpressions {
+  private ConstantStringExpressions() {}
+
+  static String toConstantStringExpression(Object value, VisitorState state) {
```
cushon, line 27: I'd prefer to pass in the `javax.lang.model.util.Elements` here instead of `VisitorState`, it's helpful if the multi-release part is more self contained and does depend on other parts of check_api

cushon:
I tried merging this an am seeing failures like the ones in https://github.com/google/error-prone/actions/runs/12602521683/job/35125772504?pr=4744

```
Error: [ERROR] com/google/errorprone/InvalidCommandLineOptionException
Error:  	at org.apache.maven.plugin.surefire.booterclient.ForkStarter.fork(ForkStarter.java:628)
```

I'm able to reproduce those locally with:

```
mvn clean install -DskipTests=true
mvn test -e
```

Any ideas what's going wrong?

Stephan202:
@cushon this one was tricky! First to debug, then to find a (hacky...) workaround. It's a shame that we can't use `multiReleaseOutput`, as that would likely have avoided these shenanigans. Maybe one of the other watchers of this project knows of a more elegant solution :eyes: :crossed_fingers:. I rebased the branch and added a commit; PTAL :)

NB: instead of introducing a new `reset-output-directory` step I also tried to move down the `default-compile` step, but that did not cause the order of the compilation steps to change.

Stephan202:
Added another commit to apply the `Elements` change.

One issue I just identified: with my change `mvn clean install -DskipTests=true && mvn test -e` still doesn't fully work when building with JDK 23+, with test failures due to the wrong `Convert.quote` being invoked. I think that this is because without reference to the JAR, the `META-INF/MANIFEST.MF` file is ignored. As-is `./check_api/target/classes/META-INF/MANIFEST.MF` also doesn't contain `Multi-Release: true`, thought the following patch would "fix" that:
```diff
diff --git a/check_api/pom.xml b/check_api/pom.xml
index 9e8bc1e8df..a5de90556e 100644
--- a/check_api/pom.xml
+++ b/check_api/pom.xml
@@ -229,17 +229,6 @@
         </executions>
 
       </plugin>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-jar-plugin</artifactId>
-        <configuration>
-          <archive>
-            <manifestEntries>
-              <Multi-Release>true</Multi-Release>
-            </manifestEntries>
-          </archive>
-        </configuration>
-      </plugin>
     </plugins>
   </build>
 
diff --git a/pom.xml b/pom.xml
index ac716fb30f..11238fe292 100644
--- a/pom.xml
+++ b/pom.xml
@@ -147,6 +147,7 @@
             </goals>
             <configuration>
               <bnd><![CDATA[
+                Multi-Release: true
                 Bundle-SymbolicName: com.google.$<replacestring;$<replacestring;${project.artifactId};^error_prone;errorprone>;_;.>
                 Automatic-Module-Name: $<Bundle-SymbolicName>
                 -exportcontents: com.google.errorprone*
```

... but as said, that doesn't fix `mvn clean install -DskipTests=true && mvn test -e` on JDK 23+.

A simple fix would be to update the CI configuration to just combine installation and testing:
```diff
diff --git a/.github/workflows/ci.yml b/.github/workflows/ci.yml
index 25684fd42f..afcc2c6970 100644
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,12 +84,9 @@ jobs:
           java-version: ${{ matrix.java }}
           distribution: 'zulu'
           cache: 'maven'
-      - name: 'Install'
+      - name: 'Install and test'
         shell: bash
-        run: mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
-      - name: 'Test'
-        shell: bash
-        run: mvn test -B
+        run: mvn install -Dmaven.javadoc.skip=true -B -V
       - name: 'Javadoc'
         shell: bash
         run: mvn -P '!examples' javadoc:javadoc
```

Right now I don't have other ideas. :thinking:

cushon:
I'm fine with changing the invoke in CI invocation, but if separate `mvn install` and `mvn test` don't work that seems like it might also affect non-CI development builds? Or would you not expect this to affect development builds?

(I guess another option here would be to give up on multi-release jars for now and just use reflection for this method. But there isn't a good reflection-based alternative alternative for the problem https://github.com/google/error-prone/commit/08e0d108da02ee6b413822db1628f7948b335b10 solves, so it would be nice to make mr-jars work.)

PR #4620, opened 2024-10-15
A couple of fixes in MoreAnnotations

msridhar:
I discovered these issues while using this code in NullAway; see https://github.com/uber/NullAway/pull/1055.  (We support older Error Prone versions so can't directly call this API and had to adapt the code instead.)  The previous code didn't handle explicit annotations on lambda parameters (e.g., `(@Nullable Object x) -> { ... }`) and annotations on enum constants.  Not sure the best way to add tests for this but happy to add them if given a suggestion.  Actually the enum constant case was weird; I was unable to repro with bytecodes output by `javac` and I only observed the case with an `ijar` from Bazel.

FYI @cpovirk @cushon

### DIFF
--- a/check_api/src/main/java/com/google/errorprone/util/MoreAnnotations.java
+++ b/check_api/src/main/java/com/google/errorprone/util/MoreAnnotations.java
@@ -30,6 +30,8 @@ import com.sun.tools.javac.code.TargetType;
 import com.sun.tools.javac.code.Type;
 import com.sun.tools.javac.code.TypeAnnotationPosition;
 import com.sun.tools.javac.code.TypeTag;
+import com.sun.tools.javac.tree.JCTree;
+
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
@@ -130,6 +132,8 @@ public final class MoreAnnotations {
       case LOCAL_VARIABLE:
         return position.type == TargetType.LOCAL_VARIABLE;
       case FIELD:
+      // treated like a field
+      case ENUM_CONSTANT:
         return position.type == TargetType.FIELD;
       case CONSTRUCTOR:
       case METHOD:
@@ -137,8 +141,15 @@ public final class MoreAnnotations {
       case PARAMETER:
         switch (position.type) {
           case METHOD_FORMAL_PARAMETER:
-            return ((MethodSymbol) sym.owner).getParameters().indexOf(sym)
-                == position.parameter_index;
+            int parameterIndex = position.parameter_index;
+            if (position.onLambda != null) {
+              com.sun.tools.javac.util.List<JCTree.JCVariableDecl> lambdaParams =
+                      position.onLambda.params;
+              return parameterIndex < lambdaParams.size()
+                      && lambdaParams.get(parameterIndex).sym.equals(sym);
+            } else {
+              return ((Symbol.MethodSymbol) sym.owner).getParameters().indexOf(sym) == parameterIndex;
+            }
           default:
             return false;
         }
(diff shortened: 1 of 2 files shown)


msridhar:
I'll try to look at tests later.  For the enum constant case, this was the jar:

https://repo1.maven.org/maven2/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.52.0/beam-runners-google-cloud-dataflow-java-2.52.0.jar

I don't know if I can share the ijar.  The reference in the client code was to `DataflowWorkerLoggingOptions.Level.INFO` and that's where the `ENUM_CONSTANT` case was popping up.  (Also strange as I couldn't find the annotations in the source code.)  Maybe that's enough to repro?

cushon:
The bytecode for `DataflowWorkerLoggingOptions.Level.INFO` is the following. (I also compared the ijar output, ijar doesn't appear to do anything interesting to that field.) The [source code doesn't have annotations](https://github.com/apache/beam/blob/f7b987dbf9b4f34f60ea61cfa9cb4b6b5488be0b/runners/google-cloud-dataflow-java/src/main/java/org/apache/beam/runners/dataflow/options/DataflowWorkerLoggingOptions.java#L53), so I think their release builds are using the Checker Framework which is doing its magic to add the annotations:

```
  public static final org.apache.beam.runners.dataflow.options.DataflowWorkerLoggingOptions$Level INFO;
    descriptor: Lorg/apache/beam/runners/dataflow/options/DataflowWorkerLoggingOptions$Level;
    flags: (0x4019) ACC_PUBLIC, ACC_STATIC, ACC_FINAL, ACC_ENUM
    RuntimeVisibleTypeAnnotations:
      0: #8(): FIELD
        org.checkerframework.checker.nullness.qual.UnknownKeyFor
      1: #9(): FIELD
        org.checkerframework.checker.nullness.qual.NonNull
      2: #10(): FIELD
        org.checkerframework.checker.initialization.qual.Initialized
```

This isn't bytecode that you can get from javac, there's no way to write type annotations on the type of a desugared enum constant field. Annotating enum constants just results in declaration annotations, not type annotations:

```java
import org.checkerframework.checker.initialization.qual.Initialized;
import org.checkerframework.checker.nullness.qual.NonNull;
import org.checkerframework.checker.nullness.qual.UnknownKeyFor;

enum E {
  @UnknownKeyFor
  @NonNull
  @Initialized
  ONE
}
```

```
  public static final E ONE;
    descriptor: LE;
    flags: (0x4019) ACC_PUBLIC, ACC_STATIC, ACC_FINAL, ACC_ENUM
    RuntimeVisibleAnnotations:
      0: #33()
        org.checkerframework.checker.nullness.qual.UnknownKeyFor
      1: #34()
        org.checkerframework.checker.nullness.qual.NonNull
      2: #35()
        org.checkerframework.checker.initialization.qual.Initialized
```

So this is kind of weird, but I guess it's a field in bytecode, and if there are type annotations there it's probably fine to just interpret them like we would for normal fields.

msridhar:
Thanks for digging in, @cushon!  And sorry for the `ijar` red herring.  Somehow I couldn't repro this in a NullAway unit test even using that exact jar, but probably I messed something else up, as the annotations are clearly there in the bytecodes.

msridhar:
@cushon I've attempted to add a test for the explicitly-annotated lambda case but I'm not sure if it will pass.  It's possible the `MoreAnnotationsTester` checker needs to be changed to also match lambda trees.

cushon (approved):

msridhar:
As a follow up i re-tested NullAway with the linked Apache Beam jar and confirmed I could reproduce the previous crash, and that the line `case ENUM_CONSTANT:` from this PR fixes it.  I must have messed up with my testing before.

cushon:
Sounds good, thanks for confirming. The test passes as-is, thanks for adding it.

Issue #4616, opened 2024-10-13
`MemberName` checks class names too


Stephan202:
As of 43ed6f3b9de2b6cb4d93488792287a945eeb5a1d, the `MemberName` check also flags type names. Resolving such violations is often much more involved than renaming (private) field names, possibly even prohibitive due to compatibility requirements.

Additionally, some of the suggestions are possibly controversial:
- To rename [`MoreASTHelpers`](https://github.com/PicnicSupermarket/error-prone-support/blob/beb96f0f4b1474283e13ff8cb32d519644436383/error-prone-utils/src/main/java/tech/picnic/errorprone/utils/MoreASTHelpers.java) to `MoreAstHelpers`, while it relates to [`ASTHelpers`](https://github.com/google/error-prone/blob/2bebaf2dfbf6e18d09e6bcdb9fcf59626908688a/check_api/src/main/java/com/google/errorprone/util/ASTHelpers.java).
- To rename [`AssertThatThrownByIOException`](https://github.com/PicnicSupermarket/error-prone-support/blob/beb96f0f4b1474283e13ff8cb32d519644436383/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJThrowingCallableRules.java#L360-L371) to `AssertThatThrownByIoException`, while this Refaster rule relates to Java's built-in `IOException` type.
- To rename [`TestNGToAssertJRules`](https://github.com/PicnicSupermarket/error-prone-support/blob/beb96f0f4b1474283e13ff8cb32d519644436383/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/TestNGToAssertJRules.java) to `TestNgToAssertJRules`, while the associated testing framework is named TestNG, not TestNg.
- Similarly, to rename [`MongoDbTextFilterUsage`](https://github.com/PicnicSupermarket/error-prone-support/blob/beb96f0f4b1474283e13ff8cb32d519644436383/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MongoDBTextFilterUsage.java) to `MongoDbTextFilterUsage`, while the database is named MongoDB, not MongoDb.

Suggestion: perhaps this type name logic can be moved to a separate check, with a more suitable name. Otherwise, what about allowing the type name logic to be disabled using a flag?

CC @graememorgan.

graememorgan:
[...]

PR #4646, opened 2024-10-30
Introduce `IdentifierName:AllowInitialismsInTypeName` flag

Stephan202:
While default `IdentifierName` behavior is unchanged, this flag allows
users to slightly relax the type name validation performed by this
check.


Stephan202:
See [this message](https://github.com/google/error-prone/issues/4616#issuecomment-2446102499) and the preceding comments for context. Naturally happy to rework the PR to have the flag trigger slightly different behavior, if desired.

cushon (approved):

PR #4699, opened 2024-11-24
Propagate check flags in patch mode

Stephan202:


check_api/src/main/java/com/google/errorprone/ErrorProneAnalyzer.java
```diff
@@ -80,13 +80,12 @@ public static ErrorProneAnalyzer createAnalyzer(
...
-                          ErrorPronePlugins.loadPlugins(scannerSupplier, context);
+                          ErrorPronePlugins.loadPlugins(scannerSupplier, context)
+                              .applyOverrides(epOptions);
                       ImmutableSet<String> namedCheckers =
                           epOptions.patchingOptions().namedCheckers();
                       if (!namedCheckers.isEmpty()) {
                         toUse = toUse.filter(bci -> namedCheckers.contains(bci.canonicalName()));
-                      } else {
-                        toUse = toUse.applyOverrides(epOptions);
                       }
```
Stephan202, line 89: Besides that all existing tests still pass: this change seems reasonable, at least on a conceptual level. I didn't thoroughly assess whether the non-flag changes applied by `applyOverrides` are no-ops, though. Could have a closer look at that later.

core/src/test/java/com/google/errorprone/ErrorProneJavaCompilerTest.java
```diff
@@ -410,6 +417,59 @@ public void withExcludedPaths() {
...
+                "-XepOpt:AssignmentUpdater:NewValue=new-value"),
+            Collections.singletonList(AssignmentUpdater.class));
+    assertThat(result.succeeded).isTrue();
+    assertThat(Files.readString(location))
+        .isEqualTo(
+            """
+            class StringConstantWrapper {
+              String s = "new-value";
+            }
+            """);
```
Stephan202, line 470: Without the fix above, this test fails with:
```
[ERROR] Failures: 
[ERROR]   ErrorProneJavaCompilerTest.patchWithBugPatternCustomization:465 value of:
    readString(...)
expected:
    class StringConstantWrapper {
      String s = "new-value";
    }
    
but was:
    class StringConstantWrapper {
      String s = "flag-not-set";
    }
```

Stephan202 (commented):
I found this issue while working on PicnicSupermarket/error-prone-support#1426, where I noticed that a custom flag `-XepOpt:Slf4jLogDeclaration:CanonicalStaticLoggerName=LOGGER` was respected during regular compilation, but not when in patch mode.

Stephan202:
An issue that's likely fixed by this change was just reported against NullAway: uber/NullAway#1080. CC @cushon; might be worthy of a new release.

msridhar:
Would it be better for Error Prone to detect the case where a disabled check is passed to `-XepPatchChecks` and throw an _understandable_ exception for that case?  It seems that with #4028 such checks are just ignored, but that could lead to some unexpected behaviors.  Not sure this idea makes sense or helps with finding a simpler overall fix.

check_api/src/main/java/com/google/errorprone/ErrorProneAnalyzer.java
```diff
@@ -79,15 +80,18 @@ public static ErrorProneAnalyzer createAnalyzer(
...
+                              .applyOverrides(epOptions)
+                              .filter(
+                                  bci -> {
+                                    String name = bci.canonicalName();
+                                    return epOptions.getSeverityMap().get(name) != Severity.OFF
+                                        && (namedCheckers.isEmpty()
+                                            || namedCheckers.contains(name));
+                                  });
                       return ErrorProneScannerTransformer.create(toUse.get());
                     }));
```
Stephan202, line 96: So patching now ignores checks:
- Whose severity is `OFF`, or
- That are not explicitly enabled for patching, while other checks are.

Stephan202 (commented):
I rebased the PR and added a second commit. I think this PR is now ready for review. @cushon if deemed acceptable, it'd be nice to have a new release with this change. :pray:

Happy holidays!

cushon (approved):

krisso-rtb:
Hi @Stephan202 
Could you please take a look at https://github.com/google/error-prone/issues/4943 and let us know if the issue is related with this PR.
Thanks in advance!

Stephan202:
@krisso-rtb sorry for the tardy response. It looks like this change uncovered non-intuitive behavior of the `ScannerSupplier#filter` method. I filed #5102 with a proposed fix.

PR #4742, opened 2025-01-03
Remove redundant JDK toolchain version

tkindy:
This PR removes the redundant specification of JDK version in the configuration of the `maven-toolchains-plugin`.

Each property can only have one value, and the way the Maven parameter implementation works, the last value wins. I confirmed this by running a build of the current `master` using `mvnDebug` and adding a breakpoint in the `maven-toolchains-plugin`'s `ToolchainsMojo` to see what the configuration values it pulled in were:

So the toolchains plugin is always picking `25` given the current configuration; the `17` is silently ignored.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -302,7 +302,6 @@
         <configuration>
           <toolchains>
             <jdk>
-              <version>17</version>
               <version>25</version>
             </jdk>
           </toolchains>


cushon (approved):

tkindy:
Thanks @cushon!

PR #4782, opened 2025-01-20
Add support for jakarta annotations to some checks

gavlyukovskiy:
This MR adds jakarta annotations to the exempt lists. 
We're currently migrating to Jakarta EE and we have to temporarily disable the rule or add exceptions in the code.

I didn't add new tests since it's just additional exempt annotations, but let me know if you want explicit tests.


core/src/main/java/com/google/errorprone/bugpatterns/UnusedVariable.java
```diff
@@ -168,6 +173,7 @@ public final class UnusedVariable extends BugChecker implements CompilationUnitT
           "com.google.inject.multibindings.ProvidesIntoMap",
           "com.google.inject.multibindings.ProvidesIntoSet",
           "dagger.Provides",
+          "jakarta.inject.Inject",
```
tbroyer, line 176: Maybe also add this to UnnecessarilyVisible? (as hinted in the TODO comment above the constant)

gavlyukovskiy: Added it there as well as to few other places I found. Thanks!
I've seen it also in `InjectMatchers` and other places that seem to relate to dependency injection / Guice. I didn't change that, because the logic is a bit more complicated, and since we're not using Guice, I'm not that confident in changing those and I'm not sure whether it even supports Jakarta EE already.

tbroyer: > since we're not using Guice, I'm not that confident in changing those. I'm not sure whether it even supports Jakarta EE already.

FYI, it does!
https://github.com/google/guice/wiki/Guice600
https://github.com/google/guice/wiki/Guice700

gavlyukovskiy: @tbroyer good to know :) But do you think we can have these checks merged separately from Guice or are they somehow intertwined? 
I also noticed that sometimes there are suggested fixes to add `javax.inject.Inject` annotation, but it doesn't make sense if the project is already using jakarta 🤷  I think there could be some heuristic, but I'm not the best person to define that as we're using spring instead 😅

gavlyukovskiy:
Hi @tbroyer, did you have a chance to take a look at this PR? I've updated all the places that do not touch guice, let me know if you need anything else from me.

gavlyukovskiy:
@tbroyer thanks for taking a look :) `UnusedMethod` already has `jakarta` equivalents for everything else, but I found two that are missing - `jakarta.validation.constraints.AssertFalse` and `jakarta.validation.constraints.AssertTrue` so I've added those as well.
You definitely have more experience contributing to this project, so would you suggest reaching out somewhere or just wait until a maintainer takes a look? In our project we are already using patched `UnusedVaraible`, so it's not urgent to be released, but it would be nice to have this fix upstream in general.

cushon (approved):

gavlyukovskiy:
@cushon thank you for reviewing this PR, I've just rebased it on the latest master to resolve some conflicts.

cushon (approved):

PR #4801, opened 2025-02-01
Fix edge case in `InconsistentCapitalization`

PBoddington:
`InconsistentCapitalization` does not correctly handle the case that the class has multiple fields with names that are equivalent except for capitalization.

For example, the following code
```
class Test {
    
    private int abc;
    private int ABC;
    
    void foo(int ABC) {
    
    }
}
```
results in an error
```
[InconsistentCapitalization] Found the field 'abc' with the same name as the parameter 'ABC' but with different capitalization.
```

This proposed change fixes the behaviour for this edge case.

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/InconsistentCapitalization.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/InconsistentCapitalization.java
@@ -16,11 +16,12 @@
 
 package com.google.errorprone.bugpatterns;
 
-import static com.google.common.collect.ImmutableMap.toImmutableMap;
+import static com.google.common.collect.ImmutableListMultimap.toImmutableListMultimap;
 import static com.google.errorprone.BugPattern.SeverityLevel.WARNING;
 import static com.google.errorprone.util.ASTHelpers.isStatic;
 
 import com.google.common.base.Ascii;
+import com.google.common.collect.ImmutableListMultimap;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/google/errorprone/bugpatterns/InconsistentCapitalization.java
```diff
@@ -212,12 +215,10 @@ public Void visitVariable(VariableTree tree, Void unused) {
...
-      Symbol matchedField = fields.get(Ascii.toLowerCase(variableName));
-      if (matchedField != null) {
-        String fieldName = matchedField.toString();
-        if (!variableName.equals(fieldName)) {
-          matchedParameters.put(getCurrentPath(), matchedField);
-        }
+      List<Symbol> matchedFields = fields.get(Ascii.toLowerCase(variableName));
+      if (!matchedFields.isEmpty()
+          && matchedFields.stream().map(Symbol::toString).noneMatch(variableName::equals)) {
+        matchedParameters.put(getCurrentPath(), matchedFields.getFirst());
```
PBoddington, line 221: I chose to use a `List` because using `matchedFields.getFirst()` is the simplest way to choose a matching field in a manner that is obviously deterministic. It would have to be a very weird class for the performance of iterating over the list to be a concern. However, I'd happily change the `ListMultimap` to something like a `Map<String, Map<String, Symbol>>` if requested.

Issue #886, opened 2017-12-31
Option to disallow unused checker suppressions


ben-manes:
As code evolves, some suppressions may be added that become unnecessary later one. However, these are easy to miss during refactoring and remains as benign. Unfortunately, this leaves open the possibility of a new error being introduced and uncaught. It can be unfeasible to audit a large code base for unnecessary suppressions after the fact, and generally shouldn't be an error as each tool has its own subset.

It would be nice if ErrorProne allowed its checks to be verified when suppressed. If the suppression is enabled but no error was reported, then the failure it reported. This would be hooked in [ErrorProneScanner](https://github.com/google/error-prone/blob/master/check_api/src/main/java/com/google/errorprone/scanner/ErrorProneScanner.java#L489) and disabled by default.

A scenario is null data flow analysis (https://github.com/uber/NullAway/issues/100) where false positives are suppressed. Then as the code or checker evolves these become potential hazards.

jbduncan:
I don't really have anything to add to this other than I think this is, at least in theory, a great idea! I've no idea how practical it would be, but considering just the initial idea here, I'd be very, very happy to see something like this implemented.

cushon:
This is something we've discussed, and I agree it would be nice to have. The approach I'd take would be similar to the one described here: https://github.com/typetools/checker-framework/issues/1701#issuecomment-349403873

msridhar:
@cushon the naive approach I was thinking about was to just tweak the `ErrorProneScanner` code to look something like (for `AssignmentTree`s):
```java
    for (AssignmentTreeMatcher matcher : assignmentMatchers) {
      if (!isSuppressed(matcher, state)) {
        try {
          reportMatch(matcher.matchAssignment(tree, state), tree, state);
        } catch (Throwable t) {
          handleError(matcher, t);
        }
      } else if (checkSuppressionValidity) {
        if (!Description.NO_MATCH.equals(matcher.matchAssignment(tree, state))) {
          // warn about unnecessary suppression
        }
      }
    }
```
I don't think this would hurt performance much when not checking for valid suppressions.  It does clutter the ErrorProneScanner code a bit but I think it's still readable (the `else if` logic could be extracted to a function).  I might be missing something here, though.

PR #4828, opened 2025-02-18
[draft] initial implementation of emitting warnings for unneeded suppressions

msridhar:
See #886.  This change requires various cleanups, better naming, documentation, etc., and it doesn't handle all cases, but I wanted to post it to get feedback on whether this approach seems reasonable or needs to be reworked in some way.

For the most part this PR piggybacks on the extant tracking of which suppressions are active and when checkers are being suppressed.  If a warning would be emitted by a checker and there is an active suppression of that warning, the suppression is marked as "used."  Then, in `Scanner`, warnings are emitted for unused suppressions when restoring the old suppression information.  Custom suppression annotations are not yet handled.  Also we don't yet report the warning on the proper `@SuppressWarnings` annotation directly and support auto-fixing.

The main complication is that checkers may discover and use suppression information in ways that are hard to automatically detect.  E.g., see [this logic in `UngroupedOverloads`](https://github.com/google/error-prone/blob/dd0016168e79b5c0754c3dd0abcea0d7fe91610d/core/src/main/java/com/google/errorprone/bugpatterns/UngroupedOverloads.java#L141-L143); NullAway also has such custom logic.  If these uses of suppression annotations are not detected, it will lead to suppressions falsely being reported as unused.  To avoid that, for now checkers need to _opt in_ to supporting unused warning suppression checking.  I opted in `JdkObsolete` as an example, as it does not seem to have any non-standard checking of suppressions.  If this approach makes sense, maybe we can find a way to automatically opt in more checkers in the future (or just do so by hand).  I also added an API to enable checkers to expose their custom uses of suppressions, and made some initial use of it in [this WIP NullAway change](https://github.com/msridhar/NullAway/commit/a4545a2ef25ce9b9df16eb7707b79b53526b6d0c).

If this approach looks reasonable / acceptable I'll clean up the changes and make them ready for a proper review.  FYI @cushon @cpovirk


msridhar:
@cushon was curious if you had any thoughts on this approach and whether it's acceptable.  Happy to do more cleanup if you feel it's needed before you can take a quick-ish look.  And I do intend to implement proper support for auto-fix before requesting a proper review.

ben-manes:
PMD recently [added](https://docs.pmd-code.org/pmd-doc-7.14.0/pmd_release_notes.html#new-rule-unnecessarywarningsuppression-experimental) this feature and allowed me to clear out 38 stale suppressions. We ran this draft and cleaned up a bunch of those too.

check_api/src/main/java/com/google/errorprone/SuppressionInfo.java
```diff
@@ -82,18 +164,29 @@ private static boolean isGenerated(Symbol sym) {
...
+      String name;
+      if (suppressWarningsStrings.contains("all")) {
+        name = "all";
+      } else {
+        // find the first name that suppresses this check
+        name =
+            suppressible.allNames().stream()
+                .filter(suppressWarningsStrings::contains)
+                .findAny()
+                .get();
```
cushon, line 181: I wonder about refactoring to avoid repeating this logic between the `disjoint` call above and here.

check_api/src/main/java/com/google/errorprone/SuppressionInfo.java
```diff
@@ -61,11 +69,85 @@ public class SuppressionInfo {
 
   private final boolean inGeneratedCode;
 
+  @BugPattern(summary = "Warns when a @SuppressWarnings is not needed", severity = WARNING)
+  private static final class UnusedSuppressionChecker extends BugChecker {}
```
cushon, line 73: Is it possible to suppress UnusedSuppressionChecker findings? This isn't a feature request to do that, I'm just thinking through it.

I suppose one reason to want that would be if an Error Prone check name shadowed a suppression string used by another tool, and Error Prone didn't report the findings but the other tool would.

cushon (commented):
Thanks for working on this!

I took a quick pass. I think this makes sense overall, I will try to do some internal testing.

check_api/src/main/java/com/google/errorprone/SuppressionInfo.java
msridhar: Good idea, fixed in 7c5c0d5e6d83e17d8b418999a05863e67ec6d7d0

check_api/src/main/java/com/google/errorprone/SuppressionInfo.java
msridhar: I guess I'm not sure about this.  On the one hand, I think it's probably generally good for all warnings to be suppressible somehow.  [Javac is working on adding a similar check](https://github.com/openjdk/jdk/pull/25167) and there you can suppress the unneeded suppression warning.

On the other hand, I wasn't sure if users would always run with this unused suppression check enabled.  That could potentially negatively impact build times, since without this new check, Error Prone completely skips running checkers on sub-trees where the warnings would be suppressed.  But maybe users don't rely on that?

In any case, I can add this support if we decide to do it.

msridhar:
> I took a quick pass. I think this makes sense overall, I will try to do some internal testing.

Sounds good!  Let me know how things look after that.

cushon:
> > I took a quick pass. I think this makes sense overall, I will try to do some internal testing.
> 
> Sounds good! Let me know how things look after that.

Still waiting on results for the testing, but I left a couple of small comments on things I noticed while importing it to set that up.

check_api/src/main/java/com/google/errorprone/SuppressionInfo.java
```diff
@@ -61,11 +70,85 @@ public class SuppressionInfo {
...
+        usedSuppressWarningStrings.add(suppressionName);
+      } else if (parent != null) {
+        parent.markSuppressionStringAsUsed(suppressionName);
+      } else {
+        throw new IllegalArgumentException("Suppression string not found: " + suppressionName);
+      }
+    }
+  }
+
+  private final @Nullable SuppressionInfoForSymbol infoForClosestSymbol;
```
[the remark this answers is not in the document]
msridhar: Do we need to maintain the immutability of the class?  Right now we mutate the `usedSuppressWarningStrings` set when we detect that a suppression has actually caused a warning to be suppressed; mutable state seemed like the simplest option for that.  If we want the enclosing class to remain immutable, any suggestions on where else to track that a  suppression was used?  We could thread an extra parameter through a lot of code to track this, I guess, but it seems like would be a much more invasive change (didn't think through it deeply).

msridhar:
FYI, I am pushing changes on my branch, but this is _not_ ready for any kind of further review; just experimenting.

msridhar:
As a follow up, the complexity of this implementation got out of hand and I lost momentum on it.  In the meantime, I created a script for removing unnecessary suppressions based on repeated builds and parsing error messages, and put it in the NullAway repo:

https://github.com/uber/NullAway/tree/e97f8ff91c1596b95a43b447e1982209aa79cd75/scripts/nullaway-suppression-remover

It's not as good as an Error Prone mode to warn on unnecessary suppressions; in particular it's not easy to run on every build.  But, for periodically removing unnecessary NullAway suppressions, it's better than nothing.  I'll bet it would work fine for other Error Prone checks, but that hasn't been tested.

I'm going to give up on this PR, as I don't think I'll be able to get it working and reviewable anytime soon.

Issue #4832, opened 2025-02-21
IdentifierName rejects m-failsafe-p default inclusion patterns


commonquail:
By default, the Maven Failsafe plugin relies on a set of file name patterns to determine which test classes to include. This is the [`<includes>`](https://maven.apache.org/surefire/maven-failsafe-plugin/integration-test-mojo.html#includes) option, and its default values are

- `**/IT*.java`
- `**/*IT.java`
- `**/*ITCase.java`

The `IdentifierName` check (which I'm unreasonably excited about!) accurately detects those names as being [in violation](https://google.github.io/styleguide/javaguide.html#s5.3-camel-case).

Unfortunately, for a user wishing to take advantage of Failsafe, working around this a bit cumbersome; it seems they can...

- ... **not run Error Prone on test code at all**, but I would like to.
- ... **change `<includes>` to any compliant non-default value**, but that introduces non-standard behaviour for dubious reasons.
- ... **use `-XepOpt:IdentifierName:AllowInitialismsInTypeName=true`**, but actually, that allows a name like `XMLHTTPRequest` that is specifically called out as in violation, and incidentally, is exactly the sort of situations I see people struggle with the most.
- ... **suppress each violation**, which I personally consider the correct response but a noisy and tiresome one.

Every option is possible but all of them carry friction.

The behaviour I would most like to see is that `IdentifierName` automagically did not flag these cases so the defaults Just Work™. But Error Prone cannot know which arbitrary patterns I might have set in my own `<includes>`, so any support for this is bound to be either inflexible or (unreasonably?) complex. Correspondingly, I would suggest that **`IdentifierName` should _at most_ allow Failsafe's default `<includes>` values**

A more flexible approach is a new option, `-XepOpt:IdentifierName:MyBikeshed=*IT,*MXBean`, which passes a set of patterns to `IdentifierName` to allow.

See also #4776.

```
$ ./mvnw clean test-compile -DepFlags='-XepDisableAllChecks -Xep:IdentifierName:ERROR'
[INFO] Scanning for projects...
[INFO]
...
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR :
[INFO] -------------------------------------------------------------
[ERROR] /.../src/test/java/example/ITFoo.java:[1,8] [IdentifierName] Classes should be named in UpperCamelCase, with acronyms treated as words (https://google.github.io/styleguide/javaguide.html#s5.3-camel-case); did you mean 'ItFoo'?
[...]

Issue #4776, opened 2025-01-17
Allow MXBean in IdentifierName


PhilippWendler:
Naming an interface with the suffix `MXBean` marks it as being an [MXBean](https://docs.oracle.com/en/java/javase/17/docs/api/java.management/javax/management/MXBean.html). This should be allowed by the `IdentifierName` check such that one does not have to add a suppression to every MXBean interface.

There is also an `@MXBean` annotation that can be used as an alternative for the name suffix, but MXBeans are basically always a public API (they can be accessed via the JVM management interface), so probably many users will not want to rename existing types.

PR #4842, opened 2025-02-25
IdentifierName: accept regex to exempt type name violations

commonquail:
Here is an attempt at teaching `IdentifierName` to allow custom type name patterns that would otherwise be considered (true positive) violations, for use where external factors necessitate a violating form but the existing `AllowInitialismsInTypeName` is either too accepting or fails to match.

This implementation satisfies the JMX M(X)Bean type name pattern (#4776) and the Maven Failsafe plugin filter pattern (#4832), however, I consider some of its details contradictory; see notes.

For the new option I'm using the working name of `AllowRegexInTypeName`, derived from `AllowInitialismsInTypeName`, but I have reservations about that.

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/IdentifierName.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/IdentifierName.java
@@ -60,6 +60,7 @@ import com.sun.tools.javac.code.Symbol.ClassSymbol;
 import com.sun.tools.javac.code.Symbol.MethodSymbol;
 import com.sun.tools.javac.code.Symbol.VarSymbol;
 import com.sun.tools.javac.util.Name;
+import java.util.Optional;
 import java.util.regex.Pattern;
 import java.util.stream.Stream;
 import javax.inject.Inject;
@@ -103,10 +104,14 @@ public final class IdentifierName extends BugChecker
 
   private final boolean allowInitialismsInTypeName;
 
+  private final Pattern allowRegexInTypeName;
+
   @Inject
   IdentifierName(ErrorProneFlags flags) {
     this.allowInitialismsInTypeName =
         flags.getBoolean("IdentifierName:AllowInitialismsInTypeName").orElse(false);
+    this.allowRegexInTypeName =
+        Pattern.compile(flags.get("IdentifierName:AllowRegexInTypeName").orElse(".+"));
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/google/errorprone/bugpatterns/IdentifierName.java
```diff
@@ -294,7 +299,9 @@ private static boolean isConformantLowerCamelName(String name) {
   private boolean isConformantTypeName(String name) {
     return !name.contains("_")
         && isUpperCase(name.charAt(0))
-        && (allowInitialismsInTypeName || !PROBABLE_INITIALISM.matcher(name).find());
+        && (allowInitialismsInTypeName
+            || allowRegexInTypeName.matcher(name).matches()
```
commonquail, line 303: `AllowInitialismsInTypeName` reads and works approximately like "allow initialisms [when they occur within a] type name". Similarly, `AllowRegexInTypeName` reads approximately like "allow [this] regex [when it matches within a] type name." However, I've used the anchored `matches()` instead of the unanchored `find()` and that makes me think that either

1) the option name and this placement are incorrect, or
2) `matches()` is incorrect.

When I imagine how I would reason about this functionality, I arrive at "a simple-name that satisfies this pattern should be allowed, even when it would have been rejected". That mental model seems to interact with the fewest other rules and therefore be the simplest and easiest to reason about, but it can also be a blanket pass to ignore every other defined rule and that may be too much flexibility. In any case, to accurately implement that mental model

1. the new regex should be the first thing we `matches()`, and only on failure should we proceed to the original checks.
2. the regex's current fallback match-anything pattern would have to change, perhaps just into a presence/absence form.
3. the name should probably not use the word "in", which can imply `find()`-semantics, but be something more like `AllowTypeNameMatchingRegex`.

A completely different approach is to parameterize `PROBABLE_INITIALISM` with ~its current value as the default. That grants maximal flexibility, but my objective with this proposal is not to bypass or challenge the styleguide's definition of camel case, only to provide what amounts to a high-level suppression mechanic. I think parameterizing `PROBABLE_INITIALISM` would be a fair bit more difficult to build and to use, and would grant an escape hatch that is too wide.

core/src/test/java/com/google/errorprone/bugpatterns/IdentifierNameTest.java
```diff
@@ -658,6 +658,21 @@ public void className_underscore() {
         .doTest();
   }
 
+  @Test
+  public void className_badPattern_allowed() {
```
commonquail, line 662: - Name?
- Explicit positive case?
- Interaction with other rules?

core/src/main/java/com/google/errorprone/bugpatterns/IdentifierName.java
commonquail: For demonstration, a version that more faithfully matches my mental model: <https://github.com/google/error-prone/compare/master...commonquail:error-prone:identifiername-allow-class-regex-harder>. If I had to explain to somebody how this worked, that's the version I would rather explain.

PR #4897, opened 2025-03-14
style: prefer https:// links in favour of http://

vlsi:
The following regular expression identified the deviations:

`http://(?!(www\.w3|www\.apache\.org/licenses|maven\.apache\.org/(POM|xsd)))`


vlsi:
Note: the front page uses lists `http://` link as well, however, I was not able to find the source for the doc

https://errorprone.info/ -> 

```
$ bazel build :hello
ERROR: example/myproject/BUILD:29:1: Java compilation in rule '//example/myproject:hello'
ShortSet.java:6: error: [CollectionIncompatibleType] Argument 'i - 1' should not be passed to this method;
its type int is not compatible with its collection's type argument Short
      s.remove(i - 1);
              ^
    (see http://errorprone.info/bugpattern/CollectionIncompatibleType)
1 error
```

tbroyer:
It's in the `gh-pages` branch: https://github.com/google/error-prone/blob/gh-pages/index.md

core/src/test/java/com/google/errorprone/bugpatterns/flogger/FloggerWithCauseTest.java
```diff
@@ -105,7 +105,7 @@ public void method(Exception e) {
         .doTest();
   }
 
-  // regression test for http://b/29131466
+  // regression test for https://b/29131466
```
cushon, line 108: All of the `b/` links should stay as `http:`

cushon (changes requested):
Thanks for the cleanup!

core/src/test/java/com/google/errorprone/bugpatterns/flogger/FloggerWithCauseTest.java
vlsi: I've reverted `/b/` links

PR #4906, opened 2025-03-17
Adds type_annotations back but as a relocation to annotations

tbroyer:
Adds a dependency to type_annotations into annotations to make sure it'll "update" an older type_annotations dependency present in the same "bag" of dependencies (works best in Gradle; Maven having a "nearest definition" rule so won't always work but hopefully people align versions one way or another).

### DIFF
--- a/annotations/pom.xml
+++ b/annotations/pom.xml
@@ -29,6 +29,11 @@
 
   <dependencies>
     <dependency>
+      <groupId>com.google.errorprone</groupId>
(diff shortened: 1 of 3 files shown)


tbroyer:
With Gradle, the relocation (being always present in the resolution) will print a note when run with `--info` logs (or more verbose).

With Maven, it'll only generate a warning if the type_annotations is directly present in the pom.xml (from my tests)

cushon (approved):

PR #4924, opened 2025-03-20
Remove dependency on error_prone_type_annotations

protocol7:
error_prone_type_annotations has been merged in error_prone_annotations, but the latter still depends on the former. Package managers that doesn't handle relocated missing dependencies (such as Bazel's rules_jvm_external) struggles with this.

This cleans up the no longer needed dependency.

### DIFF
--- a/annotations/pom.xml
+++ b/annotations/pom.xml
@@ -29,11 +29,6 @@
 
   <dependencies>
     <dependency>
-      <groupId>com.google.errorprone</groupId>
-      <artifactId>error_prone_type_annotations</artifactId>
-      <version>${project.version}</version>
-    </dependency>
-    <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
       <version>${junit.version}</version>


cushon (approved):

chadlwilson:
As this seems to have been merged - could we get a release cut with this?

The extra runtime transitive is now being pulled into gson (at least for Gradle users, as there appears to be no equivalent Gradle module metadata published to model the relocation and don't think Gradle understands such Maven magic) which seems unnecessary, especially for folks who never relied on the separate type annotations.

```
runtimeClasspath - Runtime classpath of source set 'main'.
+--- com.google.code.gson:gson:2.13.0
|    \--- com.google.errorprone:error_prone_annotations:2.37.0
|         \--- com.google.errorprone:error_prone_type_annotations:2.37.0
|              \--- com.google.errorprone:error_prone_annotations:2.37.0 (*)
```

cushon:
I pushed a 2.38.0 release that includes this change: https://github.com/google/error-prone/releases/tag/v2.38.0

chadlwilson:
Thx @cushon !

PR #5102, opened 2025-06-14
Have `-XepPatchChecks:` skip disabled checks

### LINKED ISSUE #4943: [v2.37.0] Disabled checks are executed when XepPatchChecks: is empty
While upgrading ErrorProne from `2.36` to `2.37` we observed checks defined in 
```
com.google.errorprone.scanner.BuiltInCheckerSuppliers.DISABLED_CHECKS
```
are surprisingly fired.
Our `maven-compiler-plugin` has config:
```
<arg>
    -Xplugin:ErrorProne \
    -XepPatchLocation:IN_PLACE \
    -XepPatchChecks:${errorprone.patchChecks} \
    -XepDisableAllChecks \
    -Xep:NullAway:ERROR \
    <!-- and our custom checkers: -->
```

where `errorprone.patchChecks` is dynamically-provided Maven param.
When it's empty we have effectively:
```
    -XepPatchLocation:IN_PLACE \
    -XepPatchChecks: \
   ...
```
and tons of unwanted checks are executed, e.g. `BooleanParameter` or `UnnecessarilyFullyQualified`.

When we remove these two lines it's fine.

**The issue does NOT exist in `2.36`.**

----

It might be added by https://github.com/google/error-prone/pull/4699.
In `ErrorProneAnalyzer` a few suspicious changes were made, especially around emptiness of `namedCheckers` variable.

Stephan202:
Resolves #4943.

### DIFF
--- a/check_api/src/main/java/com/google/errorprone/scanner/ScannerSupplier.java
+++ b/check_api/src/main/java/com/google/errorprone/scanner/ScannerSupplier.java
@@ -17,6 +17,7 @@
 package com.google.errorprone.scanner;
 
 import static com.google.common.collect.ImmutableListMultimap.flatteningToImmutableListMultimap;
+import static java.util.function.Predicate.not;
 
 import com.google.common.base.Predicate;
 import com.google.common.base.Supplier;
@@ -298,7 +299,7 @@ public abstract class ScannerSupplier implements Supplier<Scanner> {
   public ScannerSupplier filter(Predicate<? super BugCheckerInfo> predicate) {
     ImmutableSet<String> disabled =
         getAllChecks().values().stream()
-            .filter(predicate.negate())
+            .filter(bci -> disabled().contains(bci.canonicalName()) || !predicate.test(bci))
             .map(BugCheckerInfo::canonicalName)
             .collect(ImmutableSet.toImmutableSet());
     return new ScannerSupplierImpl(getAllChecks(), severities(), disabled, getFlags());
--- a/core/src/test/java/com/google/errorprone/ErrorProneJavaCompilerTest.java
+++ b/core/src/test/java/com/google/errorprone/ErrorProneJavaCompilerTest.java
@@ -487,6 +487,32 @@ public class ErrorProneJavaCompilerTest {
   }
 
   @Test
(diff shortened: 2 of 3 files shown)


check_api/src/main/java/com/google/errorprone/scanner/ScannerSupplier.java
```diff
@@ -298,7 +299,7 @@ public ScannerSupplier plus(ScannerSupplier other) {
   public ScannerSupplier filter(Predicate<? super BugCheckerInfo> predicate) {
     ImmutableSet<String> disabled =
         getAllChecks().values().stream()
-            .filter(predicate.negate())
+            .filter(bci -> disabled().contains(bci.canonicalName()) || !predicate.test(bci))
```
Stephan202, line 302: The changes in #4699 assumed that `ScannerSupplier#filter` would merely disable additional checks, but in fact it could re-enable disabled checks. This change proposes to resolve that, yielding more intuitive semantics.

core/src/test/java/com/google/errorprone/ErrorProneJavaCompilerTest.java
```diff
@@ -486,6 +486,32 @@ class StringConstantWrapper {
...
+            ImmutableList.of(AssignmentUpdater.class));
+    assertThat(result.succeeded).isTrue();
+    assertThat(Files.readString(Path.of(fileObject.toUri())))
+        .isEqualTo(
+            """
+            class StringConstantWrapper {
+              String s = "old-value";
+            }
+            """);
+  }
```
Stephan202, line 513: This new test fails prior to the proposed changes. Observe that the only difference with the preceding `patchAllWithCheckDisabled` test is the use of `-XepDisableAllChecks` instead of `-Xep:AssignmentUpdater:OFF`.

Stephan202 (commented):
Some context.

CRogers:
Thanks for the fix!

CRogers:
Would it be possible to merge this fix?

TotallyMehis:
+1

This is particularly annoying, would love to see this merged.

cushon (approved):

PR #5103, opened 2025-06-16
Create variable names that are CheckStyle compat

CVdV-au:
Existing code creates variable name for pattern match by lowercasing first char.
A common case where this rule will apply is SQLException, but sQLException fails https://checkstyle.sourceforge.io/checks/naming/abbreviationaswordinname.html#Description
where allowedAbbreviationLength is anything < 3.

This change lowercases the first char, and then any subsequent char where the char before is uppercase, and the char after is uppercase, resulting in:
SQLException -> sqlException
IOException -> ioException
InterruptedIOException -> interuptedIoException


CVdV-au:
https://github.com/google/error-prone/pull/5264 instead

PR #5121, opened 2025-06-25
Added RedundantNullCheck bug pattern

### LINKED ISSUE #5107: UnnecessaryCheckNotNull doesn't work with @NullMarked
I'm not sure how UnnecessaryCheckNotNull is supposed to work, it doesn't do anything out of the box with @NullMarked packages. 

Edit: it doesn't do anything with explicitly annotated @NotNull parameters, either.

Edit 2: unless the docs are incorrect and it's not actually on by default.

bdragan:
Closes #5107.

A null check is redundant if it is performed on a variable or method call that is known to be non-null within a `@NullMarked` scope.

Within a `@NullMarked` scope, types are non-null by default unless explicitly annotated with `@Nullable`.
Therefore, checking a variable or method return value (that isn't `@Nullable`) for nullness is unnecessary, as it should never be null.

Example:

```java
import org.jspecify.annotations.NullMarked;
import org.jspecify.annotations.Nullable;

@NullMarked
class MyClass {
  void process(String definitelyNonNull) {
    // BUG: Diagnostic contains: RedundantNullCheck
    if (definitelyNonNull == null) {
      System.out.println("This will never happen");
    }
    // ...
  }

  String getString() {
    return "hello";
  }

  @Nullable String getNullableString() {
    return null;
  }

  void anotherMethod() {
    String s = getString();
    // BUG: Diagnostic contains: RedundantNullCheck
    if (s == null) {
      // s is known to be non-null because getString() is not @Nullable
      // and we are in a @NullMarked scope.
      System.out.println("Redundant check");
    }

    String nullableStr = getNullableString();
    if (nullableStr == null) { // This check is NOT redundant
      System.out.println("Nullable string might be null");
    }
  }
}
```

This check helps to clean up code and reduce clutter by removing unnecessary null checks, making the code easier to read and maintain.
It also reinforces the contract provided by `@NullMarked` and `@Nullable` annotations.

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/nullness/RedundantNullCheck.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/nullness/RedundantNullCheck.java
@@ -52,7 +52,9 @@ import javax.inject.Inject;
 import javax.lang.model.element.ElementKind;
 
 @BugPattern(
-    summary = "Explicit null check on a variable or method call that is not @Nullable within a @NullMarked scope.",
+    summary =
+        "Null check on an expression that is statically determined to be non-null according to "
+            + "language semantics or nullness annotations.",
     severity = WARNING)
 public class RedundantNullCheck extends BugChecker
     implements BinaryTreeMatcher, MethodInvocationTreeMatcher {
@@ -84,7 +86,7 @@ public class RedundantNullCheck extends BugChecker
 
     boolean isRedundant = false;
     if (symbol instanceof VarSymbol) {
-      isRedundant = isRedundantForVarSymbol((VarSymbol) symbol, state);
+      isRedundant = !isEffectivelyNullable((VarSymbol) symbol, state);
     } else if (symbol instanceof MethodSymbol) {
       isRedundant = !isEffectivelyNullable((MethodSymbol) symbol, state);
     }
@@ -105,7 +107,7 @@ public class RedundantNullCheck extends BugChecker
     boolean isRedundant = false;
     VarSymbol varSymbol = nullCheck.varSymbolButUsuallyPreferBareIdentifier();
     if (varSymbol != null) {
-      isRedundant = isRedundantForVarSymbol(varSymbol, state);
+      isRedundant = !isEffectivelyNullable(varSymbol, state);
     } else {
       MethodSymbol methodSymbol = nullCheck.methodSymbol();
       if (methodSymbol != null) {
@@ -119,32 +121,30 @@ public class RedundantNullCheck extends BugChecker
     return NO_MATCH;
   }
 
-  private static boolean isRedundantForVarSymbol(VarSymbol varSymbol, VisitorState state) {
-    if (!NullnessUtils.isInNullMarkedScope(varSymbol, state)) {
-      return false;
+  private static boolean isEffectivelyNullable(VarSymbol varSymbol, VisitorState state) {
+    Optional<Nullness> varNullness = NullnessAnnotations.fromAnnotationsOn(varSymbol);
+    if (varNullness.isPresent()) {
+      return varNullness.get() == Nullness.NULLABLE;
     }
-    if (NullnessUtils.isAlreadyAnnotatedNullable(varSymbol)) {
-      return false;
-    }
+
     if (varSymbol.asType().getKind() == TYPEVAR) {
-      return false;
+      return true;
     }
 
     VariableTree varDecl = NullnessUtils.findDeclaration(state, varSymbol);
     if (varSymbol.getKind() == ElementKind.PARAMETER && hasImplicitType(varDecl, state)) {
-      return false;
+      return true;
     }
 
     if ((varSymbol.getKind() == ElementKind.LOCAL_VARIABLE
-        || varSymbol.getKind() == ElementKind.RESOURCE_VARIABLE)
+            || varSymbol.getKind() == ElementKind.RESOURCE_VARIABLE)
         && varDecl != null) {
-
       if (varDecl.getInitializer() == null) {
-        return false;
+        return true;
       }
 
       if (!isConsideredFinal(varSymbol)) {
-        return false;
+        return true;
       }
 
       Tree initializer = varDecl.getInitializer();
@@ -153,17 +153,18 @@ public class RedundantNullCheck extends BugChecker
         MethodInvocationTree methodInvocation = (MethodInvocationTree) initializer;
         MethodSymbol methodSymbol = ASTHelpers.getSymbol(methodInvocation);
         if (methodSymbol != null && isEffectivelyNullable(methodSymbol, state)) {
-          return false;
+          return true;
         }
       } else if (initializer instanceof LiteralTree) {
         if (initializer.getKind() == Tree.Kind.NULL_LITERAL) {
-          return false;
+          return true;
         }
       } else {
-        return false;
+        return true;
       }
     }
-    return true;
+
+    return !NullnessUtils.isInNullMarkedScope(varSymbol, state);
   }
 
   private static boolean isEffectivelyNullable(MethodSymbol methodSymbol, VisitorState state) {
--- a/core/src/main/java/com/google/errorprone/bugpatterns/nullness/RedundantNullCheck.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/nullness/RedundantNullCheck.java
@@ -126,9 +126,11 @@ public class RedundantNullCheck extends BugChecker
         varSymbol.getKind() == ElementKind.LOCAL_VARIABLE
             || varSymbol.getKind() == ElementKind.RESOURCE_VARIABLE;
 
-    Optional<Nullness> varNullness = NullnessAnnotations.fromAnnotationsOn(varSymbol);
-    if (varNullness.isPresent()) {
-      return varNullness.get() == Nullness.NULLABLE;
+    if (!isLocalOrResourceVariable) {
+      Optional<Nullness> varNullness = NullnessAnnotations.fromAnnotationsOn(varSymbol);
+      if (varNullness.isPresent()) {
+        return varNullness.get() == Nullness.NULLABLE;
+      }
     }
 
     if (varSymbol.asType().getKind() == TYPEVAR) {
--- a/core/src/main/java/com/google/errorprone/bugpatterns/nullness/RedundantNullCheck.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/nullness/RedundantNullCheck.java
@@ -1,0 +1,99 @@
+/*
+ * Copyright 2025 The Error Prone Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.errorprone.bugpatterns.nullness;
+
+import static com.google.errorprone.BugPattern.SeverityLevel.WARNING;
+import static com.google.errorprone.matchers.Description.NO_MATCH;
+
+import com.google.errorprone.BugPattern;
+import com.google.errorprone.VisitorState;
+import com.google.errorprone.bugpatterns.BugChecker;
+import com.google.errorprone.bugpatterns.BugChecker.BinaryTreeMatcher;
+import com.google.errorprone.bugpatterns.nullness.NullnessUtils.NullCheck;
+import com.google.errorprone.dataflow.nullnesspropagation.Nullness;
+import com.google.errorprone.dataflow.nullnesspropagation.NullnessAnnotations;
+import com.google.errorprone.matchers.Description;
+import com.google.errorprone.util.ASTHelpers;
+
+import com.sun.source.tree.BinaryTree;
+import com.sun.source.tree.MethodInvocationTree;
+import com.sun.source.tree.Tree;
+import com.sun.source.tree.VariableTree;
+import com.sun.tools.javac.code.Symbol.MethodSymbol;
+import com.sun.tools.javac.code.Symbol.VarSymbol;
+
+import java.util.Optional;
+
+@BugPattern(
+    summary = "Explicit null check on a variable or method call that is not @Nullable within a @NullMarked scope.",
+    severity = WARNING)
+public class RedundantNullCheck extends BugChecker implements BinaryTreeMatcher {
+
+  @Override
+  public Description matchBinary(BinaryTree tree, VisitorState state) {
+    NullCheck nullCheck = NullnessUtils.getNullCheck(tree);
+    if (nullCheck == null) {
+      return NO_MATCH;
+    }
+
+    VarSymbol varSymbol = nullCheck.varSymbolButUsuallyPreferBareIdentifier();
+    if (varSymbol != null) {
+      VariableTree varDecl = NullnessUtils.findDeclaration(state, varSymbol);
+
+      if (varDecl != null
+          && varDecl.getInitializer() != null
+          && varDecl.getInitializer().getKind() == Tree.Kind.METHOD_INVOCATION) {
+        MethodInvocationTree methodInvocation = (MethodInvocationTree) varDecl.getInitializer();
+        MethodSymbol methodSymbol = ASTHelpers.getSymbol(methodInvocation);
+
+        if (methodSymbol != null && isEffectivelyNullable(methodSymbol, state)) {
+          return NO_MATCH;
+        }
+      }
+
+      if (!NullnessUtils.isInNullMarkedScope(varSymbol, state)) {
+        return NO_MATCH;
+      }
+
+      if (NullnessUtils.isAlreadyAnnotatedNullable(varSymbol)) {
+        return NO_MATCH;
+      }
+    } else {
+      MethodSymbol methodSymbol = nullCheck.methodSymbol();
+      if (methodSymbol == null) {
+        return NO_MATCH;
+      }
+      if (isEffectivelyNullable(methodSymbol, state)) {
+        return NO_MATCH;
+      }
+    }
+    return buildDescription(tree).build();
+  }
+
+  private static boolean isEffectivelyNullable(MethodSymbol methodSymbol, VisitorState state) {
+    Optional<Nullness> returnTypeNullness = NullnessAnnotations.fromAnnotationsOn(methodSymbol);
+    if (returnTypeNullness.isPresent()) {
+      // Explicit @Nullable or @NonNull on the return type
+      return returnTypeNullness.get() == Nullness.NULLABLE;
+    }
+    // No explicit annotation on return type.
+    // Default based on the null-marked status of the method's defining scope.
+    // If the method's defining scope is NOT @NullMarked (or is @NullUnmarked),
+    // its unannotated return type is effectively nullable.
+    return !NullnessUtils.isInNullMarkedScope(methodSymbol, state);
+  }
+}
--- a/core/src/main/java/com/google/errorprone/bugpatterns/nullness/RedundantNullCheck.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/nullness/RedundantNullCheck.java
@@ -24,6 +24,7 @@ import static com.google.errorprone.util.ASTHelpers.isConsideredFinal;
 import static javax.lang.model.type.TypeKind.TYPEVAR;
 
 import com.google.errorprone.BugPattern;
+import com.google.errorprone.ErrorProneFlags;
 import com.google.errorprone.VisitorState;
 import com.google.errorprone.bugpatterns.BugChecker;
 import com.google.errorprone.bugpatterns.BugChecker.BinaryTreeMatcher;
@@ -47,6 +48,7 @@ import com.sun.tools.javac.code.Symbol.VarSymbol;
 
 import java.util.Optional;
 
+import javax.inject.Inject;
 import javax.lang.model.element.ElementKind;
 
 @BugPattern(
@@ -58,8 +60,19 @@ public class RedundantNullCheck extends BugChecker
   private static final Matcher<ExpressionTree> OBJECTS_REQUIRE_NON_NULL =
       staticMethod().onClass("java.util.Objects").named("requireNonNull");
 
+  private final boolean checkRequireNonNull;
+
+  @Inject
+  public RedundantNullCheck(ErrorProneFlags flags) {
+    this.checkRequireNonNull =
+        flags.getBoolean("RedundantNullCheck:CheckRequireNonNull").orElse(false);
+  }
+
   @Override
   public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+    if (!checkRequireNonNull) {
+      return NO_MATCH;
+    }
     if (!OBJECTS_REQUIRE_NON_NULL.matches(tree, state)) {
       return NO_MATCH;
     }
--- a/core/src/main/java/com/google/errorprone/bugpatterns/nullness/RedundantNullCheck.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/nullness/RedundantNullCheck.java
@@ -18,6 +18,7 @@ package com.google.errorprone.bugpatterns.nullness;
 
 import static com.google.errorprone.BugPattern.SeverityLevel.WARNING;
 import static com.google.errorprone.matchers.Description.NO_MATCH;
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/google/errorprone/bugpatterns/nullness/RedundantNullCheck.java
```diff
@@ -0,0 +1,109 @@
...
+
+    VarSymbol varSymbol = nullCheck.varSymbolButUsuallyPreferBareIdentifier();
+    if (varSymbol != null) {
+      if (!NullnessUtils.isInNullMarkedScope(varSymbol, state)) {
+        return NO_MATCH;
+      }
+      if (NullnessUtils.isAlreadyAnnotatedNullable(varSymbol)) {
+        return NO_MATCH;
+      }
+
```
cpovirk, line 63: The main tricky case is a type-variable type, like `E element` in the implementation of `singletonList` (though of course there are no nullness annotations on that specific API): It shouldn't be annotated as `@Nullable`, but its value can be `null` if someone creates a singleton list of a type like `List<@Nullable String>`.

The simplest thing is to avoid making suggestions for type-variable types at all:

```suggestion
      if (varSymbol.asType().getKind() == TYPEVAR) {
        return NO_MATCH;
      }

```

That matches [what we do in some other nullness checks, where I've written about how we someday _could_ do better but can't do so trivially](https://github.com/google/error-prone/blob/fbdfbb58615c52ec620c0be6928d641a1d07a80f/core/src/main/java/com/google/errorprone/bugpatterns/nullness/NullArgumentForNonNullParameter.java#L223-230).

core/src/main/java/com/google/errorprone/bugpatterns/nullness/RedundantNullCheck.java
```diff
@@ -0,0 +1,109 @@
...
+    return buildDescription(tree).build();
+  }
+
+  private static boolean isEffectivelyNullable(MethodSymbol methodSymbol, VisitorState state) {
+    Optional<Nullness> returnTypeNullness = NullnessAnnotations.fromAnnotationsOn(methodSymbol);
+    if (returnTypeNullness.isPresent()) {
+      // Explicit @Nullable or @NonNull on the return type
+      return returnTypeNullness.get() == Nullness.NULLABLE;
+    }
+    // No explicit annotation on return type.
```
cpovirk, line 103: Similarly to the other type-variable case above:

```suggestion
    if (methodSymbol.getReturnType().getKind() == TYPEVAR) {
      return true;
    }
    // No explicit annotation on return type.
```

core/src/main/java/com/google/errorprone/bugpatterns/nullness/RedundantNullCheck.java
```diff
@@ -0,0 +1,109 @@
...
+    VarSymbol varSymbol = nullCheck.varSymbolButUsuallyPreferBareIdentifier();
+    if (varSymbol != null) {
+      if (!NullnessUtils.isInNullMarkedScope(varSymbol, state)) {
+        return NO_MATCH;
+      }
+      if (NullnessUtils.isAlreadyAnnotatedNullable(varSymbol)) {
+        return NO_MATCH;
+      }
+
+      VariableTree varDecl = NullnessUtils.findDeclaration(state, varSymbol);
```
cpovirk, line 64: Another tricky case is lambda parameters: They often don't have a type declaration, so javac fills one it, and I think it normally lacks nullness annotations. We again could probably find a way to look up the appropriate types, perhaps using something like...

https://github.com/google/error-prone/blob/fbdfbb58615c52ec620c0be6928d641a1d07a80f/core/src/main/java/com/google/errorprone/bugpatterns/CheckReturnValue.java#L365

...but unless that sounds fun to you, I suggest just skipping that case, too:

```suggestion
      VariableTree varDecl = NullnessUtils.findDeclaration(state, varSymbol);
      if (hasImplicitType(varDecl, state)) {
        return NO_MATCH;
      }
```

core/src/main/java/com/google/errorprone/bugpatterns/nullness/RedundantNullCheck.java
```diff
@@ -0,0 +1,109 @@
...
+
+      if ((varSymbol.getKind() == ElementKind.LOCAL_VARIABLE
+          || varSymbol.getKind() == ElementKind.RESOURCE_VARIABLE) && varDecl != null) {
+
+        if (varDecl.getInitializer() == null) {
+          return NO_MATCH;
+        }
+
```
cpovirk, line 72: I haven't tested this one out yet, but I'm thinking that we'd want to avoid cases in which the variable is assigned a non-null value but then reassigned later. We _again_ could find clever ways to do that more precisely, but I'd again suggest just skipping the case entirely. I'm hoping that this would do it:

```suggestion
        if (!isConsideredFinal(varSymbol)) {
          return NO_MATCH;
        }

```

(I was about to say that it might render the `getInitializer() == null` check unnecessary, but I now think that that's wrong, since it's possible to write `final Foo foo;` and then assign to it later.)

core/src/main/java/com/google/errorprone/bugpatterns/nullness/RedundantNullCheck.java
```diff
@@ -0,0 +1,109 @@
...
+        return NO_MATCH;
+      }
+      if (isEffectivelyNullable(methodSymbol, state)) {
+        return NO_MATCH;
+      }
+    }
+    return buildDescription(tree).build();
+  }
+
+  private static boolean isEffectivelyNullable(MethodSymbol methodSymbol, VisitorState state) {
```
cpovirk, line 97: A totally optional thing that you could look into is handling of constructors: I think those end up in the `MethodSymbol` code path, and of course any call to a constructor produces a non-null value, regardless of annotations. Given that, you might well be able to `return false` if `methodSymbol.isConstructor()`.

cpovirk (commented):
Thank you for doing this, and I apologize again for not getting to it.

My suggested edits assume a few static imports:

```java
import static com.google.errorprone.util.ASTHelpers.hasImplicitType;
import static com.google.errorprone.util.ASTHelpers.isConsideredFinal;
import static javax.lang.model.type.TypeKind.TYPEVAR;
```

Whatever special cases you're up for adding tests for are great, but I can't complain when you already have a bunch of tests and I've left you hanging so long.

bdragan:
@cpovirk Great, thank you, I incorporated and pushed your suggestions (with two additions - implicit vars which aren't parameters and literal initializers). Also added tests for the new use cases.

bdragan:
@cpovirk Also we think that extending this with support for redundant `Objects.requireNonNull` calls would be beneficial, wdyt? If you agree I will also push it here - I already have a working solution for direct calls which are most common in practice, i.e. `Objects.requireNonNull(x)` (so method references are not covered i.e. `Object::requireNonNull` but that should be fine I guess).

cpovirk:
> @cpovirk Also we think that extending this with support for redundant `Objects.requireNonNull` calls would be beneficial, wdyt? If you agree I will also push it here - I already have a working solution for direct calls which are most common in practice, i.e. `Objects.requireNonNull(x)` (so method references are not covered i.e. `Object::requireNonNull` but that should be fine I guess).

That is a good question. I tend to have different views for different code:

In places like Guava, we actively want to make sure that we produce NPE immediately if users pass `null` where they shouldn't. This prepares us for the potential that [the JVM performs similar checks automatically and unavoidably someday](https://openjdk.org/jeps/8303099), but more immediately, it ensures that we aren't setting up a trap that will spring later in the program (like we would be doing if we were to let you put `null` into a collection that shouldn't support it). More often, the NPE does happen immediately "naturally," so there's no need to check. But we at least occasionally still check explicitly so that we can provide a nicer exception message (or so that we can support a weird environment that doesn't guarantee NPE, as we sometimes have to do).

For more typical code, I don't usually bother with defensive checks at all, since the damage from getting things "wrong" is more contained. There, I'd expect to see `requireNonNull` only for the "non-redundant" cases.

I would say that it's fine for you to do whatever suits you, but I will eventually end up running this by someone for internal review, so we may end up changing course then. We also always have the option of making the behavior configurable with a flag. That still leaves the question of which behavior to make the default, but it at least offers flexibility to users who have strong enough opinions.

core/src/main/java/com/google/errorprone/bugpatterns/nullness/RedundantNullCheck.java
```diff
@@ -0,0 +1,133 @@
...
+
+    VarSymbol varSymbol = nullCheck.varSymbolButUsuallyPreferBareIdentifier();
+    if (varSymbol != null) {
+      if (!NullnessUtils.isInNullMarkedScope(varSymbol, state)) {
+        return NO_MATCH;
+      }
+      if (NullnessUtils.isAlreadyAnnotatedNullable(varSymbol)) {
+        return NO_MATCH;
+      }
+      if (varSymbol.asType().getKind() == TYPEVAR) {
```
cpovirk, line 67: Sorry, I was just looking at this suggestion of mine again, and I think I was mistaken: We shouldn't need to check `TYPEVAR` for _variables_: I'm thinking that the only case in which we could ever produce a finding for a type-variable-typed variable is a call to a method that returns `@NonNull T` for some type-variable type `T`. In that case, we're already relying on `isEffectivelyNullable(MethodSymbol, ...)` to do the right thing for `TYPEVAR`.

So I think we can remove the `TYPEVAR` check here, leaving only the one below (the one for methods).

core/src/main/java/com/google/errorprone/bugpatterns/nullness/RedundantNullCheck.java
```diff
@@ -0,0 +1,133 @@
...
+        return NO_MATCH;
+      }
+
+      VariableTree varDecl = NullnessUtils.findDeclaration(state, varSymbol);
+      if (varSymbol.getKind() == ElementKind.PARAMETER && hasImplicitType(varDecl, state)) {
+        return NO_MATCH;
+      }
+
+      if ((varSymbol.getKind() == ElementKind.LOCAL_VARIABLE
```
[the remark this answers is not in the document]
cpovirk: Other thoughts on this proposal (in addition to the ones above on line 61), again with the motivation of capturing them for the future:

If we cover fields, we could identify _additional_ opportunities (while also keeping all the current ones) by looking for `@NonNull` annotations on them and `@NullMarked` annotations on the enclosing elements. That would let us treat even more fields as non-null, including but not limited to non-`final` fields.

That said, we do sometimes see users who claim that a field is non-null when it's merely non-null _after some initialization occurs_. So it's possible that those users would be annoyed if we were to complain that their lazy-initialization code checks for the nullness of a "non-null" field. That, at least, is an issue only for non-`final` fields.

core/src/main/java/com/google/errorprone/bugpatterns/nullness/RedundantNullCheck.java
```diff
@@ -0,0 +1,133 @@
...
+  @Override
+  public Description matchBinary(BinaryTree tree, VisitorState state) {
+    NullCheck nullCheck = NullnessUtils.getNullCheck(tree);
+    if (nullCheck == null) {
+      return NO_MATCH;
+    }
+
+    VarSymbol varSymbol = nullCheck.varSymbolButUsuallyPreferBareIdentifier();
+    if (varSymbol != null) {
+      if (!NullnessUtils.isInNullMarkedScope(varSymbol, state)) {
```
cpovirk, line 61: I hadn't thought about this before, even when I commented on the docs on the subject, but: Does checking for null-marked buy us anything? Any information that we get about a variable is either from the method whose result it's initialized to or from "language knowledge" (constructors, literals). So we're ([as usual](https://jspecify.dev/docs/api/org/jspecify/annotations/Nullable.html#applicability)) not relying on the annotations for the variable's "top-level" type.

(This would change if we were to start handling _fields_, as discussed in [a thread below](https://github.com/google/error-prone/pull/5121/files#r2373414895). That's all the more reason to _not_ do that yet :))

core/src/main/java/com/google/errorprone/bugpatterns/nullness/RedundantNullCheck.java
```diff
@@ -0,0 +1,133 @@
...
+        return NO_MATCH;
+      }
+      if (NullnessUtils.isAlreadyAnnotatedNullable(varSymbol)) {
```
cpovirk, line 64: This one is interesting: Under the logic above, we shouldn't care about annotations on the variable. (And again, JSpecify, at least, discourages annotations on local variables.) But I can see how a user who _knows_ that a value might be `null` (regardless of what the annotations say) might choose to write `@Nullable String s = something.stringThatMightSecretlyBeNull()`. If a user does that, different nullness checkers may handle it differently. Well, at least that may be true of the Checker Framework (which will _still_ treat it as non-null if that's how `stringThatMightSecretlyBeNull()` is annotated) and Kotlin (which will treat it as nullable, I _think_, or maybe that's only for _platform_ types?)—though of course Kotlin isn't relevant here, so I offer it only as an example of how different tools could make different choices.

Given that JSpecify discourages annotations on variables entirely, I'd probably drop this. What do you think?

cpovirk (commented):
Sorry, the results looked great, but I ended up with more thoughts anyway. I hope that it's just simplifications at this point.

cpovirk:
I also wanted to write down somewhere how this differs from https://errorprone.info/bugpattern/ImpossibleNullComparison, at least as I understand it:

- ImpossibleNullComparison does not trust annotations.
- ImpossibleNullComparison does not handle literals.
- ImpossibleNullComparison does not track fields after assignment.
- ImpossibleNullComparison does look for operations like `requireNonNull` that this PR at least _currently_ doesn't cover.
- ImpossibleNullComparison has some suggested fixes, at least a couple of which (the ones for `List`/`Collection` and `Optional`) we could probably someday steal for here.
- ImpossibleNullComparison handles `case null` in `switch`.
- ImpossibleNullComparison handles specific, well-known APIs.

The long-term plan should likely be to merge these, probably offering configuration options and enhanced docs. But I'm hoping to be able to make the case that we shouldn't block on that.

core/src/main/java/com/google/errorprone/bugpatterns/nullness/RedundantNullCheck.java
```diff
@@ -0,0 +1,133 @@
...
+import com.sun.source.tree.VariableTree;
+import com.sun.tools.javac.code.Symbol.MethodSymbol;
+import com.sun.tools.javac.code.Symbol.VarSymbol;
+
+import java.util.Optional;
+
+import javax.lang.model.element.ElementKind;
+
+@BugPattern(
+    summary = "Explicit null check on a variable or method call that is not @Nullable within a @NullMarked scope.",
```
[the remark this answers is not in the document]
bdragan: Indeed, we should improve the wording here. Maybe just explicitly mention both (language semantics and annotations), something like:

> Null check on an expression that is statically determined to be non-null according to language semantics or nullness annotations.

core/src/main/java/com/google/errorprone/bugpatterns/nullness/RedundantNullCheck.java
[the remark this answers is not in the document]
cpovirk: OK, let's see if I can evaluate the full logic instead of over-focusing on specific details and losing the bigger picture.

There are a few `VarSymbol` cases in which we could report that a value is never `null`:

- for a field that matches _either_ of the following conditions:
   - `@NonNull` (I'm not sure we check at the moment. That's fine)
   - a non-type-variable-typed, non-`@Nullable` field in the scope of `@NullMarked`
- for a parameter that matches either of the "field" conditions _and_ is not an implicitly typed lambda parameter
- for a variable that matches _both_ of the following conditions:
   - "considered `final`"
   - initialized to a `!isEffectivelyNullable` method or to a literal (other than `null`)
- (probably also for a _field_ that matches all the "variable" conditions, though that's another thing we don't check at the moment, and that's fine, too)

We should figure out whether that is even right. Whatever we settle on, and whatever part of that we implement now, it might be easiest to express with an `isEffectivelyNullable(VarSymbol, ...)` method. That would let us have a separate section for each case, from which we can `return` early, since we can't really do that with the logic inline in `matchBinary`. (If we have a separate method, that could also help with other possible future generalizations that we've discussed.)

core/src/main/java/com/google/errorprone/bugpatterns/nullness/RedundantNullCheck.java
cpovirk: Again, I was mixed up. See [my new comment](https://github.com/google/error-prone/pull/5121#discussion_r2376432780). We do need this logic, though arguably only for fields and parameters.

bdragan:
> > @cpovirk Also we think that extending this with support for redundant `Objects.requireNonNull` calls would be beneficial, wdyt? If you agree I will also push it here - I already have a working solution for direct calls which are most common in practice, i.e. `Objects.requireNonNull(x)` (so method references are not covered i.e. `Object::requireNonNull` but that should be fine I guess).
> 
> That is a good question. I tend to have different views for different code:
> 
> In places like Guava, we actively want to make sure that we produce NPE immediately if users pass `null` where they shouldn't. This prepares us for the potential that [the JVM performs similar checks automatically and unavoidably someday](https://openjdk.org/jeps/8303099), but more immediately, it ensures that we aren't setting up a trap that will spring later in the program (like we would be doing if we were to let you put `null` into a collection that shouldn't support it). More often, the NPE does happen immediately "naturally," so there's no need to check. But we at least occasionally still check explicitly so that we can provide a nicer exception message (or so that we can support a weird environment that doesn't guarantee NPE, as we sometimes have to do).
> 
> For more typical code, I don't usually bother with defensive checks at all, since the damage from getting things "wrong" is more contained. There, I'd expect to see `requireNonNull` only for the "non-redundant" cases.
> 
> I would say that it's fine for you to do whatever suits you, but I will eventually end up running this by someone for internal review, so we may end up changing course then. We also always have the option of making the behavior configurable with a flag. That still leaves the question of which behavior to make the default, but it at least offers flexibility to users who have strong enough opinions.

Makes sense. I pushed it now too, it is optional and disabled by default (can be enabled with `RedundantNullCheck:CheckRequireNonNull` flag).

core/src/main/java/com/google/errorprone/bugpatterns/nullness/RedundantNullCheck.java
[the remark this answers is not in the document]
bdragan: Yes, all points are correct.

> for a field that matches either of the following conditions:
> - @NonNull (I'm not sure we check at the moment. That's fine)
 
 Good catch. I was focused mostly on @NullMarked scope (where @NonNull is redundant, at least when it comes to tools implementing null-marked spec), but since I already added support for it for methods out of @NullMarked scope, I changed it now for var symbols too.
 
 > probably also for a field that matches all the "variable" conditions
 
We have to think more when it comes to fields in some situations, including alignment with other tools, e.g. a field that is not marked as nullable in @NullMarked scope will not be considered nullable even if it is not considered final (unlike local variables), etc, so we'd need to take care of it during implementation. Hence I left the logic for fields to be annotation-driven only for now.
 
 > isEffectivelyNullable(VarSymbol, ...)
 
 Good idea, I already did it for methods and the flow is easier to follow for them with that. Now I refactored the logic for var symbols that way too. :+1:

core/src/main/java/com/google/errorprone/bugpatterns/nullness/RedundantNullCheck.java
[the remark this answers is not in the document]
bdragan: Agreed; then for now I changed it (both in pattern summary and in md) to:

> Null check on an expression that is statically determined to be non-null according to language semantics or nullness annotations.

Let's think and improve more if needed.

core/src/main/java/com/google/errorprone/bugpatterns/nullness/RedundantNullCheck.java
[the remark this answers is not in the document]
bdragan: Another aspect about fields is that they are more affected by build-time or runtime code manipulation tools (i.e. like Lombok, dependency injection, bytecode weaving in runtime, etc) than local variables I guess, so the static analysis assumptions we make might not be suitable to all scenarios in practice.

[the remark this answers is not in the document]
cpovirk: I think that Error Prone conveniently blows up on Lombok code, at least :) But, yes, good point: Fields do get weird treatment from tools in a variety of situations. I also expect for _humans_ to be less reliable at annotating their fields than their APIs simply because I expect them to focus on _public_ members, while fields are usually private (aside from the occasional constants).

(I have some other distractions going on at the moment but should be able to get back to this tomorrow if I don't manage to today.)

core/src/main/java/com/google/errorprone/bugpatterns/nullness/RedundantNullCheck.java
```diff
@@ -0,0 +1,188 @@
...
+
+    boolean isRedundant = false;
+    if (symbol instanceof VarSymbol) {
+      isRedundant = !isEffectivelyNullable((VarSymbol) symbol, state);
+    } else if (symbol instanceof MethodSymbol) {
+      isRedundant = !isEffectivelyNullable((MethodSymbol) symbol, state);
+    }
+
+    if (isRedundant) {
+      return buildDescription(tree).build();
```
cpovirk, line 95:
```suggestion
      return describeMatch(tree);
```

core/src/main/java/com/google/errorprone/bugpatterns/nullness/RedundantNullCheck.java
```diff
@@ -0,0 +1,188 @@
...
+      }
+
+      if (!isConsideredFinal(varSymbol)) {
+        return true;
+      }
+
+      Tree initializer = varDecl.getInitializer();
+
+      if (initializer.getKind() == Tree.Kind.METHOD_INVOCATION) {
+        MethodInvocationTree methodInvocation = (MethodInvocationTree) initializer;
```
cpovirk, line 157:
```suggestion
      if (initializer instanceof MethodInvocationTree methodInvocation) {
```

core/src/main/java/com/google/errorprone/bugpatterns/nullness/RedundantNullCheck.java
```diff
@@ -0,0 +1,188 @@
...
+    }
+    ExpressionTree arg = tree.getArguments().get(0);
+    Symbol symbol = ASTHelpers.getSymbol(arg);
+
+    boolean isRedundant = false;
+    if (symbol instanceof VarSymbol) {
+      isRedundant = !isEffectivelyNullable((VarSymbol) symbol, state);
+    } else if (symbol instanceof MethodSymbol) {
+      isRedundant = !isEffectivelyNullable((MethodSymbol) symbol, state);
+    }
```
cpovirk, line 92:
```suggestion
    if (symbol instanceof VarSymbol varSymbol) {
      isRedundant = !isEffectivelyNullable(varSymbol, state);
    } else if (symbol instanceof MethodSymbol methodSymbol) {
      isRedundant = !isEffectivelyNullable(methodSymbol, state);
    }
```

core/src/main/java/com/google/errorprone/bugpatterns/nullness/RedundantNullCheck.java
```diff
@@ -0,0 +1,188 @@
...
+public class RedundantNullCheck extends BugChecker
+    implements BinaryTreeMatcher, MethodInvocationTreeMatcher {
+
+  private static final Matcher<ExpressionTree> OBJECTS_REQUIRE_NON_NULL =
+      staticMethod().onClass("java.util.Objects").named("requireNonNull");
+
+  private final boolean checkRequireNonNull;
+
+  @Inject
+  public RedundantNullCheck(ErrorProneFlags flags) {
```
cpovirk, line 68:
```suggestion
  RedundantNullCheck(ErrorProneFlags flags) {
```

Just to follow our internal style of reducing the visibility of constructors that are going to be called only through reflection.

core/src/main/java/com/google/errorprone/bugpatterns/nullness/RedundantNullCheck.java
```diff
@@ -0,0 +1,188 @@
...
+      isRedundant = !isEffectivelyNullable(varSymbol, state);
+    } else {
+      MethodSymbol methodSymbol = nullCheck.methodSymbol();
+      if (methodSymbol != null) {
+        isRedundant = !isEffectivelyNullable(methodSymbol, state);
+      }
+    }
+
+    if (isRedundant) {
+      return buildDescription(tree).build();
```
cpovirk, line 119:
```suggestion
      return describeMatch(tree);
```

bdragan:
> The results from running this over our depot look excellent—so many people who think that methods might return `null` when they actually throw an exception, for example, plus all the cases of missing `@Nullable` annotations that I'd mentioned before. I'm still not sure if we'll be able to justify turning it on for everyone, since some people have chosen to perform manual defensive checks on parameters, but I hope we can get there when we're finally able to focus more on nullness again!
> 
> I have run out of behavioral suggestions, so all that's left is some style-type things.

Great, I have pushed the style-type changes you suggested, all valid points, thanks!

PR #5169, opened 2025-07-22
Reproduce bug when patching and disabling through alternate name

aldexis:
This PR is meant as an example reproducer for a bug akin to https://github.com/google/error-prone/issues/3908 (issue: https://github.com/google/error-prone/issues/5170).

The aforementioned bug flagged that Error-prone was throwing an exception when patching checks that are disabled.

This was later fixed by https://github.com/google/error-prone/pull/4028.

However, this still fails when the check is disabled through an alternate name. Disabling it works fine in isolation, but it throws an exception when called with `-XepPatchChecks:`. It works fine if you use the main check name instead.

Below is an example such exception (reproduced with the `[InsecureCryptoUsage](https://github.com/google/error-prone/blob/4384ed81ed0b639307292bf5b12f820e8cfdd419/core/src/main/java/com/google/errorprone/bugpatterns/InsecureCipherMode.java)` check)
```
error: An unhandled exception was thrown by the Error Prone static analysis plugin.
        KeyFactory.getInstance(Algorithm.RSA.toString());
                              ^
  
     error-prone version: 2.40.0
     BugPattern: InsecureCryptoUsage
     Stack Trace:
        at com.google.errorprone.matchers.Description.severity(Description.java:78)
        at com.google.errorprone.ErrorProneAnalyzer.lambda$finished$3(ErrorProneAnalyzer.java:213)
        at com.google.errorprone.VisitorState.reportMatch(VisitorState.java:302)
        at com.google.errorprone.scanner.Scanner.reportMatch(Scanner.java:126)
        at com.google.errorprone.scanner.ErrorProneScanner.processMatchers(ErrorProneScanner.java:508)
        at com.google.errorprone.scanner.ErrorProneScanner.visitMethodInvocation(ErrorProneScanner.java:824)
        at com.google.errorprone.scanner.ErrorProneScanner.visitMethodInvocation(ErrorProneScanner.java:170)
        at jdk.compiler/com.sun.tools.javac.tree.JCTree$JCMethodInvocation.accept(JCTree.java:1813)

```
Side-note: The `InsecureCryptoUsage` check is a bit of an oddball in that its class name (`InsecureCipherMode`) [is the _alternate_ check name](https://github.com/google/error-prone/blob/4384ed81ed0b639307292bf5b12f820e8cfdd419/core/src/main/java/com/google/errorprone/bugpatterns/InsecureCipherMode.java#L37), not the main one, which makes it more intuitive to disable with the alternate name.

### DIFF
--- a/core/src/test/java/com/google/errorprone/ErrorProneJavaCompilerTest.java
+++ b/core/src/test/java/com/google/errorprone/ErrorProneJavaCompilerTest.java
@@ -417,7 +417,10 @@ public class ErrorProneJavaCompilerTest {
     assertThat(result.succeeded).isFalse();
   }
 
-  @BugPattern(summary = "Test bug pattern to test custom patch functionality", severity = ERROR)
+  @BugPattern(
+          summary = "Test bug pattern to test custom patch functionality",
+          severity = ERROR,
+          altNames = {"AssignmentUpdaterAltName"})
   public static final class AssignmentUpdater extends BugChecker implements VariableTreeMatcher {
     private final String newValue;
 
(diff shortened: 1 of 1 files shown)


aldexis:
Going to close this as the issue has been opened as https://github.com/google/error-prone/issues/5170 and this was only meant as a reproducer

PR #5260, opened 2025-10-04
Extend ProtoStringFieldReferenceEquality to support older versions of protobuf

styurin:
`ProtoStringFieldReferenceEquality` assumes that all protobuf messages extend `com.google.protobuf.GeneratedMessage`, but older versions of protobuf (before v26) use `com.google.protobuf.GeneratedMessageV3` as the base class for messages.

This change updates `ProtoStringFieldReferenceEquality` to use `com.google.protobuf.MessageLite` to detect protobufs. This interface is present in older versions of protobuf and it hasn't changed since then.


styurin:
Can someone take a look?

Issue #5278, opened 2025-10-19
There is no detection for `Locale.of`


mihnita:
Error-prone has `UnsafeLocaleUsage`, detecting the use of the `Locale` constructors and `Locale.toString`. \
https://errorprone.info/bugpattern/UnsafeLocaleUsage

All good...

But JDK 19 deprecated the constructors and added the static methods `Locale.of`:
* [`static Locale Locale.of(String language)`](https://docs.oracle.com/en/java/javase/19/docs/api/java.base/java/util/Locale.html#of(java.lang.String))
* [`static Locale Locale.of(String language, String country)`](https://docs.oracle.com/en/java/javase/19/docs/api/java.base/java/util/Locale.html#of(java.lang.String,java.lang.String))
* [`static Locale Locale.of(String language, String country, String variant)`](https://docs.oracle.com/en/java/javase/19/docs/api/java.base/java/util/Locale.html#of(java.lang.String,java.lang.String,java.lang.String))

The methods have the exact same problems as the constructors. \
So it would be nice if error-prone would also detect them.

Marcono1234:
Was this resolved by #5297 and should be closed now?

(it seems the PR was not automatically linked to the issue because the PR wasn't exactly using the [GitHub keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword))

cushon:
https://github.com/google/error-prone/commit/5216a438311ed3196624926a03a81ed7ce0e92ec

PR #5279, opened 2025-10-19
Update `UnsafeLocaleUsage` with detection for `Locale.of`

mihnita:
Fix for https://github.com/google/error-prone/issues/5278

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/UnsafeLocaleUsage.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/UnsafeLocaleUsage.java
@@ -38,7 +38,7 @@ import com.sun.tools.javac.tree.JCTree.JCLiteral;
 
 /** Flags unsafe usages of the {@link java.util.Locale} constructor and class methods. */
 @BugPattern(
-    summary = "Possible unsafe operation related to the java.util.Locale library.",
+    summary = "Possible unsafe operation related to the java.util.Locale class.",
     severity = WARNING)
 public final class UnsafeLocaleUsage extends BugChecker
     implements MethodInvocationTreeMatcher, NewClassTreeMatcher {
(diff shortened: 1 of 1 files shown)


docs/bugpattern/UnsafeLocaleUsage.md
```diff
@@ -51,20 +54,70 @@ There's no reliable way of getting a correct result through a `Locale`
...
+
+// Fixed code
+void someMethod(@NotNull String langId, String regionId) {
+  Locale.Builder builder = new Locale.Builder();
+  builder.setLanguage(langId);
+  if (region == null) {
+    builder.setCountry(langId);
+  }
+  Locale locale = builder.build();
+  // use the locale
```
Stephan202: IIUC:
```suggestion
void someMethod(@NotNull String langId, String regionId) {
  Locale locale (regionId == null)
      ? new Locale(langId) // or Locale.of
      : new Locale(langId, regionId); // or Locale.of
  // use the locale
}

// Fixed code
void someMethod(@NotNull String langId, String regionId) {
  Locale.Builder builder = new Locale.Builder();
  builder.setLanguage(langId);
  if (regionId == null) {
    builder.setCountry(regionId);
  }
  Locale locale = builder.build();
  // use the locale
```

core/src/main/java/com/google/errorprone/bugpatterns/UnsafeLocaleUsage.java
```diff
@@ -54,11 +57,27 @@ public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState
...
+                  + " Locale.toLanguageTag() since it produces an IETF BCP 47-formatted string"
+                  + " that can be deserialized back into a Locale.")
           .addFix(SuggestedFixes.renameMethodInvocation(tree, "toLanguageTag", state))
           .build();
     }
+    if (LOCALE_OF.matches(tree, state)) {
+      Description.Builder descriptionBuilder =
+          buildDescription(tree)
+              .setMessage(
+                  "Avoid using Locale.of static methods (which do not check their arguments for"
```
cushon: By default, the message is the `BugPattern.summary`. Is there information here that applies to all findings that could go in the summary instead?

If you want to customize this to mention e.g. `Locale.of` vs. `Locale constructors`, WDYT about refactoring and using a format string to share the text that's the same for both?

cushon (changes requested):
Thanks for this!

mihnita (commented):
Done, thank you.

core/src/main/java/com/google/errorprone/bugpatterns/UnsafeLocaleUsage.java
mihnita: Done

PR #5328, opened 2025-11-08
[rewrite] Add `Java8toJava11`

Pankraz76:
related to:

- https://github.com/diffplug/spotless/pull/2636
- https://github.com/checkstyle/checkstyle/pull/17730
- https://github.com/google/error-prone/issues/5195

Issue #5195, opened 2025-08-09
Add `rewrite` support for `errorprone.refasterrules`


Pankraz76:
## Add `rewrite` support for `errorprone.refasterrules`

I’d like to propose integrating [Google’s Error Prone](https://github.com/google/error-prone) and its Picnic extension (demonstrated in [Automating Away Bugs with Error Prone | PlatformCon 2023](https://www.youtube.com/watch?v=6llnrUtVlrE)) to enable automated bug fixes via `rewrite` rules. This would complement Checkstyle’s static analysis capabilities by addressing semantic bugs rather than stylistic issues.

## Motivation
Error Prone’s `refaster`/`rewrite` rules can automatically:
- Fix common bug patterns (e.g., `String.equals()` misuse)
- Modernize code (e.g., JDK migration helpers)
- Enforce best practices (e.g., null-check improvements)

Real-world adoptions show tangible benefits:
- https://github.com/checkstyle/checkstyle/pull/17545
- https://github.com/pmd/pmd/pull/5956
- https://github.com/opensearch-project/OpenSearch/pull/18791
- https://github.com/diffplug/spotless/pull/2576
- https://github.com/apache/kafka/pull/20219

## Proposal
1. Add support for `errorprone.refasterrules`-based rewrites
2. Implement with opt-in adoption (no breaking changes)
3. Include suppression mechanisms for API constraints

## Discussion Points
- Need consensus on:
- Scope of auto-fixes
- Preferred suppression strategy
- Integration approach

## Next Steps
I’m happy to:
- Prepare a PoC demonstrating the value
- Collaborate on implementation strategy
- Address any concerns about compatibility

relates to:

- https://github.com/checkstyle/checkstyle/issues/16543

cushon:
Hi, I'm not sure I understand what the proposal is. This is the repository for Error Prone. We already run Error Prone checks on this codebase.

Can you clarify what the proposal is here? Do you want to integrate Error Prone into something? Or add support for a specific feature to Error Prone? Or something else?

Pankraz76:
> We already run

was not aware about this, so its fixed already thanks. Any more changes are then individual approaches.

Thanks.

Pankraz76:
this is one way to integrate. i could not find picnic so i tried this way:

- https://github.com/google/error-prone/pull/5328

cushon:
[...]

PR #5343, opened 2025-11-12
[rewrite] Add `UpdateMavenWrapper`

Pankraz76:
related to:

- https://github.com/diffplug/spotless/pull/2636
- https://github.com/checkstyle/checkstyle/pull/17730
- https://github.com/google/error-prone/issues/5195
- https://github.com/google/error-prone/pull/5328


Pankraz76:
kindly requesting any feedback on this, considering:

- https://github.com/junit-team/junit-framework/pull/5193

it seems there is something for every project to gain upon, thus embracing the conventional principle.

cushon:
It would be be helpful to provide a summary in PRs and issues of what the goal is, instead of linking to discussions in other issues and PRs, it isn't obvious how this is related to the other linked issues.

Pankraz76:
Yes, of course, please excuse my shortcomings on this.

I'm wondering if the Maven wrapper is something for this project to consider, like other Google projects and most of the major Maven projects around, using this in regards to the convention principle.

This PR and most of the other related ones demonstrate how each recipe could serve and address a dedicated increment, considered a level up in terms of the same convention principle again.

Starting with some Java migration as the recently pulled changes in the dedicated PR, thanks for this consideration. Going into other details we might find considering the picnic rules.

Pankraz76:
Having now provided a dedicated issue for these two cents (#5421), I hope we can move into action.

There’s no need to use a rewrite for this; still it would benefit the project more to update things directly rather than introducing another issue while fixing an existing one.

Issue #4334, opened 2024-03-20
MultipleNullnessAnnotations shouldn't consider Validation Annotations


xenoterracide:
The Jakarta (formerly JavaX) Validation spec shouldn't be considered for this bug. My rationale is that it is a runtime validation which may not occur until late in a process. 

This is not, perhaps the best example of code, but it's what my brain is currently coming up with.

```java
public class Foo {

  @NotNull
  @Nullable
  Long getId();
} 
```

```
if ( foo.getId() == null ) {
   foo.set(1L); // yes this is terrible, but you could query a database sequence, or this could be a UUID
}
repository.save(foo); // if I remove the above if statement, and a null is passed this would throw an error
```

so, getId is validly nullable at runtime, but at the point of validation we want it to be non null.

note: you could ignore the javax/jakarta namespaced annotations explicitly, but as far as I'm aware `NotNull` is always the name of the runtime checks, and some case insensitive variant of `NonNull` is the "compile time" variant

cushon:
cc @cpovirk 

Thanks for the report. The use-case for MultipleNullnessAnnotations is primarily nullness annotations that have semantics similar to the https://github.com/jspecify/jspecify ones. We could look at having the check skip annotations with incompatible semantics, or clarifying what the use-case for it is.

cpovirk:
Yes, thanks. I'm largely clueless about the validation annotations, which I have probably seen as many times in the past couple weeks as I had in my life previously.

We have at least one [list of nullness annotations](https://github.com/jspecify/jspecify-reference-checker/blob/75aa1de2a8f88f8c23347d86465844df9887874c/src/main/java/com/google/jspecify/nullness/NullSpecAnnotatedTypeFactory.java#L1995), which sadly includes at least one notably "`@NotNull`" annotation that is "similar to JSpecify"  more than "similar to Jakarta Validation," `org.jetbrains.annotations.NotNull`. That doesn't make this proposal impossible to implement, but it keeps it from being a one-liner. For now, you might want to set `-Xep:MultipleNullnessAnnotations:OFF`.

xenoterracide:
Ugh, thanks jetbrains. Fortunately you can probably just use the two official validation spec paths for that. I don't know if I would bother going and tracking down other vendor specific ones. Like I'm sure hibernate had one at some point.

PR #5406, opened 2025-12-25
Nullness: ignore Jakarta Validation

commonquail:
Teach checks like `MultipleNullnessAnnotations` and `RedundantNullCheck`
to disregard annotations that look like nullness specification
annotations but that actually are not; namely the Jakarta Validation
`@NotNull` constraint.

Resolves: https://github.com/google/error-prone/issues/4334

### DIFF
--- a/check_api/src/main/java/com/google/errorprone/dataflow/nullnesspropagation/NullnessAnnotations.java
+++ b/check_api/src/main/java/com/google/errorprone/dataflow/nullnesspropagation/NullnessAnnotations.java
@@ -20,6 +20,7 @@ import static com.google.common.collect.ImmutableList.toImmutableList;
 import static javax.lang.model.element.ElementKind.TYPE_PARAMETER;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
 import com.sun.source.tree.AnnotationTree;
 import com.sun.source.tree.IdentifierTree;
 import com.sun.source.tree.MemberSelectTree;
@@ -37,6 +38,7 @@ import javax.lang.model.element.ElementKind;
 import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.element.Name;
 import javax.lang.model.element.Parameterizable;
+import javax.lang.model.element.QualifiedNameable;
 import javax.lang.model.element.TypeParameterElement;
 import javax.lang.model.type.IntersectionType;
 import javax.lang.model.type.TypeMirror;
@@ -59,6 +61,9 @@ public class NullnessAnnotations {
                   + "ProtoMethodMayReturnNull|ProtoMethodAcceptsNullParameter|"
                   + "ProtoPassThroughNullness")
           .asMatchPredicate();
+  private static final ImmutableSet<String> FALSE_NULLNESS_ANNOTATIONS =
+      ImmutableSet.of(
+          "jakarta.validation.constraints.NotNull", "javax.validation.constraints.NotNull");
 
   private NullnessAnnotations() {} // static methods only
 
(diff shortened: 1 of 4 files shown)


commonquail:
https://errorprone.info/bugpattern/RedundantNullCheck has the same limitation as https://github.com/google/error-prone/issues/4334, but it's more disruptive because it happens at the call site instead of at the declaration site, and dangerous because the flagged null check is actually necessary. It would be swell to have some solution to that (and I'm probably looking at having to disable `RedundantNullCheck` otherwise).

There are [a few other constraint annotations that match "null"](https://jakarta.ee/specifications/bean-validation/3.1/apidocs/jakarta/validation/constraints/package-summary#:~:text=null,-nor) but I believe none of those are an issue. Unfortunately I'm not aware of any other offenders either, so the case is not very strong.

check_api/src/main/java/com/google/errorprone/dataflow/nullnesspropagation/NullnessAnnotations.java
```diff
@@ -84,12 +89,13 @@ public static boolean annotationsAreAmbiguous(
...
       Collection<? extends AnnotationMirror> annotations) {
     return annotations.stream()
-        .filter(a -> ANNOTATION_RELEVANT_TO_NULLNESS.test(simpleName(a).toString()))
+        .filter(NullnessAnnotations::isAnnotationRelevant)
         .collect(toImmutableList());
   }
 
   public static ImmutableList<AnnotationTree> annotationsRelevantToNullness(
       List<? extends AnnotationTree> annotations) {
+    // Missing support for FALSE_NULLNESS_ANNOTATIONS
```
commonquail: I'm not sure what to do about this or how much it matters. There are test cases that hit this method but none that hit the `MemberSelectTree` inside `simpleName(AnnotationTree)`. I can get a qualified name from a `MemberSelectTree` but I don't see how that can be accomplished for an `IdentifierTree`.

check_api/src/main/java/com/google/errorprone/dataflow/nullnesspropagation/NullnessAnnotations.java
```diff
@@ -216,7 +222,10 @@ public static Optional<Nullness> getUpperBound(TypeVariable typeVar) {
 
   private static Optional<Nullness> fromAnnotationStream(
       Stream<? extends AnnotationMirror> annotations) {
-    return fromAnnotationSimpleNames(annotations.map(a -> simpleName(a).toString()));
+    return fromAnnotationSimpleNames(
+        annotations
+            .filter(NullnessAnnotations::isAnnotationRelevant)
```
commonquail: Alternatively, perhaps `simpleName` could be taught to return `Optional.empty()` in equivalent cases. A bigger alteration to existing code but at least nominally prevents calling `simpleName` before establishing the precondition.

commonquail: Oh, that doesn't work so well with `annotationsRelevantToNullness` which returns the `AnnotationMirror` (`AnnotationTree`) rather than the simple name&hellip;

check_api/src/main/java/com/google/errorprone/dataflow/nullnesspropagation/NullnessAnnotations.java
commonquail: That must be what `Matchers::isType` is for. This looks like it could work:

```java
  private static final Matcher<AnnotationTree> NOT_FALSE_NULLNESS_ANNOT =
      Matchers.not(
          Matchers.anyOf(FALSE_NULLNESS_ANNOTATIONS.stream().map(Matchers::isType).toList()));

  public static ImmutableList<AnnotationTree> annotationsRelevantToNullness(
      List<? extends AnnotationTree> annotations, final VisitorState state) {
    return annotations.stream()
        .filter(a -> NOT_FALSE_NULLNESS_ANNOT.matches(a, state))
        .filter(a -> ANNOTATION_RELEVANT_TO_NULLNESS.test(simpleName(a)))
        .collect(toImmutableList());
  }
```

but requires passing `VisitorState` to (one) `annotationsRelevantToNullness`.

PR #5525, opened 2026-02-22
[UnusedMethod] Handle reflective method references

electrum:
Handles `Class.getMethod`, `getDeclaredMethod`, `getConstructor`, `getDeclaredConstructor`, and `MethodHandles.Lookup.findStatic`, `findVirtual`, `findSpecial`, `findConstructor`. Only suppresses findings when the method name, owner class, and parameter types can all be statically resolved.

This is cleaner than adding `@Keep` everywhere, and avoids the problem that methods with `@Keep` can later become unused. IntelliJ IDEA understands static reflection as method usage, and this gives Error Prone the same ability.

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/UnusedMethod.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/UnusedMethod.java
@@ -26,14 +26,19 @@ import static com.google.errorprone.BugPattern.SeverityLevel.WARNING;
 import static com.google.errorprone.fixes.SuggestedFix.emptyFix;
 import static com.google.errorprone.fixes.SuggestedFixes.replaceIncludingComments;
 import static com.google.errorprone.matchers.Matchers.SERIALIZATION_METHODS;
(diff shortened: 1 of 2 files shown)


cushon:
This approach only handles reflective references from the same file as the method declaration. Error Prone only analyzes one file at a time, it won't be able to detect reflective references from other files. So this isn't a replacement for `@Keep`, `@Keep` is still necessary for references from other files, and reflective references to methods declared in the same file are usually rarer.

electrum:
@cushon For [Trino](https://github.com/trinodb/trino), using references in the same file to lookup `MethodHandle` is very common. We have over 100 such usages, and probably zero usages of reflection across files. Here's an [example file](https://github.com/trinodb/trino/blob/master/core/trino-main/src/main/java/io/trino/operator/scalar/ArrayConcatFunction.java). This feature would allow us to remove ~173 usages of our `@Keep`-extending `@UsedByGeneratedCode` annotation.

PR #5648, opened 2026-03-24
Fixes #5558 : Skip JdkObsolete Charset API suggestions when source level lacks them

### LINKED ISSUE #5558: URLEncoder.encode(String, String) false positive
We're using the following code in our project:

```java
return URLEncoder.encode(string, UTF_8.name());
````

When using errorprone with gradle we get a warning message that is obsolete. This is [not the case](https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/net/URLEncoder.html#encode(java.lang.String,java.lang.String)) however.

### Build output
```
     /jobrunr/core/src/main/java/org/jobrunr/utils/StringUtils.java:112: warning: [JdkObsolete] Use URLEncoder.encode(String, Charset) instead.
            return URLEncoder.encode(string, UTF_8.name());
                                    ^
    (see https://errorprone.info/bugpattern/JdkObsolete)
```

### Our setup:
```gradle
plugins {
    id 'net.ltgt.errorprone' version '5.1.0' apply false
}

dependencies {
   errorprone 'com.google.errorprone:error_prone_core:2.48.0'
}
```

mirkoalicastro:
The JdkObsolete checker recommends Charset-accepting API alternatives (for example `URLEncoder.encode(String, Charset)`) that were only added in JDK 10. Users compiling with `--release 9` or lower **cannot** use the suggested alternatives.

_Changes:_ 
- Adds a `jdkVersion` field to `ObsoleteApi` (analogous to the existing `androidMinSdkVersion`) and checks the compilation source level before reporting these APIs as obsolete.
- Sets `jdkVersion=10` for all entries whose Charset alternatives were introduced in JDK 10.
- Entries whose Charset alternatives have existed since JDK 1.6 (String, InputStreamReader, ..) are unaffected.

Fixes #5558

### DIFF
--- a/check_api/src/main/java/com/google/errorprone/util/SourceVersion.java
+++ b/check_api/src/main/java/com/google/errorprone/util/SourceVersion.java
@@ -56,7 +56,8 @@ public final class SourceVersion {
     return sourceIsAtLeast(context, 25);
   }
 
-  private static boolean sourceIsAtLeast(Context context, int version) {
+  /** Returns true if the compiler source version level is at least the given version. */
+  public static boolean sourceIsAtLeast(Context context, int version) {
     Source lowerBound = Source.lookup(Integer.toString(version));
     return lowerBound != null && Source.instance(context).compareTo(lowerBound) >= 0;
   }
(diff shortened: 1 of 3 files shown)


mirkoalicastro:
Hi @cushon . Any chance these small changes could be imported and internally reviewed? 🙏

PR #5649, opened 2026-03-24
Fixes #5553 : Skip UnnecessaryOptionalGet when lambda parameter is unnamed

### LINKED ISSUE #5553: Improved suggestions when unnamed variables are used
In Java, the `_` is not an implicit variable like `it` so the substitution does not work. The recommendation is otherwise excellent.

```console
warning: [UnnecessaryOptionalGet] This code can be simplified by directly using the lambda parameters instead of calling get..() on optional.
    return firmId.map(_ -> ShareIdentity.builder()
                     ^
    (see https://errorprone.info/bugpattern/UnnecessaryOptionalGet)
  Did you mean '.sharedToFirmId(_)'?
```

```diff
     return firmId.map(_ -> ShareIdentity.builder()
         .sharedToUserId(edgeId(sharedTo.getUser()))
-        .sharedToFirmId(firmId.get())
+        .sharedToFirmId(_)
         .identityType(PHONE_NUMBER)
         .identity(phoneNumber)
         .build());
```

mirkoalicastro:
`UnnecessaryOptionalGet` suggested replacing `.get()` with the lambda parameter name, but the parameter `_` (unnamed variable, Java 22+) produces uncompilable code since `_` cannot be referenced.

Changes:
- Added an early return when the lambda parameter source is `_`
- Added a test guarded by `assume().that(Runtime.version().feature()).isAtLeast(22)`

Fixes #5553

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/UnnecessaryOptionalGet.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/UnnecessaryOptionalGet.java
@@ -91,7 +91,7 @@ public final class UnnecessaryOptionalGet extends BugChecker
       return buildDescription(tree)
           .setMessage(
               "This code can be simplified by naming the unnamed lambda parameter and using it"
-                  + " directly instead of calling get() on the optional.")
+                  + " directly instead of calling get..() on the optional.")
           .build();
     }
     SuggestedFix.Builder fix = SuggestedFix.builder();
@@ -112,7 +112,8 @@ public final class UnnecessaryOptionalGet extends BugChecker
     return describeMatch(tree, fix.build());
   }
 
-  private static boolean usesOptionalGet(MethodInvocationTree tree, VisitorState state, ExpressionTree onlyArg) {
+  private static boolean usesOptionalGet(
+      MethodInvocationTree tree, VisitorState state, ExpressionTree onlyArg) {
     AtomicBoolean foundGet = new AtomicBoolean(false);
     new TreeScanner<Void, VisitorState>() {
       @Override


mirkoalicastro:
Hi @cushon . Any chance you could have a look at this MR? 🙏

core/src/test/java/com/google/errorprone/bugpatterns/UnnecessaryOptionalGetTest.java
```diff
@@ -388,4 +390,24 @@ private void home() {
...
+    refactoringTestHelper
+        .addInputLines(
+            "Test.java",
+            """
+            import java.util.Optional;
+
+            public class Test {
+              private void home() {
+                Optional<String> op = Optional.of("hello");
+                op.ifPresent(_ -> System.out.println(op.get()));
```
cushon, line 449: Instead of skipping the diagnostic, couldn't we suggest renaming the variable and using it instead of `.get()`?

```
    op.ifPresent(value -> System.out.println(value));
```

core/src/main/java/com/google/errorprone/bugpatterns/UnnecessaryOptionalGet.java
```diff
@@ -83,6 +83,9 @@ public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState
       return Description.NO_MATCH;
     }
     VariableTree arg = getOnlyElement(lambdaExpressionTree.getParameters());
+    if ("_".equals(state.getSourceForNode(arg))) {
+      return Description.NO_MATCH;
+    }
```
cushon, line 96: Prefer `arg.getName().isEmpty()` to test for unnamed variables, it avoids having to look at the source text.

I think the current approach doesn't handle

```
op.ifPresent((String _) -> System.out.println(op.get()));
```

could you add a test case for that?

mirkoalicastro: Ah, much better. I've made the changes and added a unit test for it, thanks.

core/src/test/java/com/google/errorprone/bugpatterns/UnnecessaryOptionalGetTest.java
mirkoalicastro: That's a nice improvement! 

Since there could be cases like:
```
String value = "";
Optional<String> op = Optional.of("hello");
op.ifPresent(_ -> System.out.println(op.get()));
```
the suggested fix cannot blindly be:
```
String value = "";
Optional<String> op = Optional.of("hello");
op.ifPresent(value -> System.out.println(value));
```
but I used a more generic message:
> This code can be simplified by naming the unnamed lambda parameter and using it directly instead of calling get..() on the optional.

mirkoalicastro:
Hey @cushon . I've made the requested changes. I'll leave the threads for you to resolve if everything looks good 🙏

core/src/test/java/com/google/errorprone/bugpatterns/UnnecessaryOptionalGetTest.java
cushon: Yes there could be existing variables to consider when picking a name, there are some examples of checks that handle that:

* https://github.com/google/error-prone/blob/7c4b0406b851d5423762b0deb82e8ab7395e79ee/core/src/main/java/com/google/errorprone/bugpatterns/PatternMatchingInstanceof.java#L178
* https://github.com/google/error-prone/blob/7c4b0406b851d5423762b0deb82e8ab7395e79ee/core/src/main/java/com/google/errorprone/bugpatterns/AssertThrowsMinimizer.java#L383

We should probably have a shared helper for this, I could do that as a follow-up. How does that sound?

cushon (approved):

core/src/test/java/com/google/errorprone/bugpatterns/UnnecessaryOptionalGetTest.java
mirkoalicastro: I see. Sounds great, thanks!

cushon: The proposed changes are here: https://github.com/google/error-prone/pull/5717

PR #5787, opened 2026-05-09
Fix BigDecimalLiteralDouble for boxed floating-point values

### LINKED ISSUE #5759: `BigDecimalLiteralDouble `  false negative for the boxed value
### Error Prone version

2.49.0 (error_prone_core)

### Check name

BigDecimalLiteralDouble

### Description

The check reports `new BigDecimal(double)` and `new BigDecimal(float)` for primitive arguments, but misses the same constructions when the value is wrapped in `Double` or `Float`. Since `BigDecimal` has no `Double`/`Float` constructor, the boxed value is automatically unboxed and the same lossy conversion still occurs. This causes a false negative for the boxed forms.

### Reproducer

```java
import java.math.BigDecimal;

class Repro {
  // BEFORE — correctly flagged
  BigDecimal a = new BigDecimal(0.1);
  BigDecimal b = new BigDecimal(0.3f);
  BigDecimal c;
  {
    float f = 0.1f;
    c = new BigDecimal(f);
  }

  // AFTER — not flagged, but semantically identical (auto-unboxing)
  Double boxedD = 0.1;
  BigDecimal a2 = new BigDecimal(boxedD);
  BigDecimal b2 = new BigDecimal((Float) 0.3f);
  Float boxedF = 0.1f;
  BigDecimal c2 = new BigDecimal(boxedF);
}
```

### Actual behavior

Only the `BEFORE` primitive-literal constructions are flagged; every boxed form — variable-typed, cast, or initialized from a literal — passes without a warning.

codingkiddo:
Fixes #5759.

## Summary

`BigDecimalLiteralDouble` reported precision-loss diagnostics for direct `double` / `float` literals passed to `BigDecimal(double)`, but it missed boxed `Double` / `Float` values.

For example:

```java
Double boxedD = 0.1;
new BigDecimal(boxedD);
```

This still invokes the `BigDecimal(double)` constructor after unboxing, so it has the same precision-loss problem as passing a primitive floating-point literal directly.

## Changes

- Added regression coverage for boxed `Double` / `Float` arguments.
- Updated `BigDecimalLiteralDouble` to detect boxed floating-point arguments consistently with primitive floating-point arguments.
- Kept the existing suggested fix behavior limited to direct literal cases, where the replacement source is safe to construct.

## Test

```bash
mvn -pl core -Dtest=BigDecimalLiteralDoubleTest test
```

Result:

```text
Tests run: 3, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/BigDecimalLiteralDouble.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/BigDecimalLiteralDouble.java
@@ -27,6 +27,8 @@ import com.google.errorprone.matchers.Description;
 import com.google.errorprone.matchers.Matcher;
 import com.google.errorprone.matchers.Matchers;
 import com.google.errorprone.util.ASTHelpers;
+import com.sun.tools.javac.code.Type;
+import javax.lang.model.type.TypeKind;
 import com.sun.source.tree.ExpressionTree;
 import com.sun.source.tree.NewClassTree;
 import com.sun.source.tree.Tree.Kind;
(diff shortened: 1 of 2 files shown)


codingkiddo:
Thanks. I have signed/updated the Google CLA check.

One note on the implementation: the diagnostic is intentionally extended to boxed `Double` / `Float`
arguments because overload resolution still selects `BigDecimal(double)` after unboxing. The suggested
fix remains limited to direct literal/unary literal cases because only those have safe replacement
source available. For boxed or variable arguments, the PR reports the precision-loss risk without
offering an automatic replacement.

PR #5868, opened 2026-06-12
Fix NPE in `VarWithPrimitive` when lambda parameters have inferred primitive types

### LINKED ISSUE #5857: [VarWithPrimitive] java.lang.NullPointerException: Cannot invoke "com.sun.tools.javac.tree.JCTree.getStartPosition()" because "tree" is null
error-prone version: 2.50.0
 BugPattern: VarWithPrimitive
 Stack Trace:
 java.lang.NullPointerException: Cannot invoke "com.sun.tools.javac.tree.JCTree.getStartPosition()" because "tree" is null

```java
import module java.base;

class Test {

    void foo() {
        byte[] bar = new byte[6];
Map<Byte, List<Byte>> indicesMap = IntStream.range(0, 6)
                .mapToObj(n -> n)
                .collect(Collectors.groupingBy(n -> bar[n], Collectors.mapping(n -> (byte) n.intValue(), Collectors.toList())));
    }
}
```

Cf. #5122

HarshMehta112:
## Problem

Fixes #5857

`VarWithPrimitive` throws a `NullPointerException` when a lambda has an implicit
parameter whose inferred type is a primitive:

```java
byte[] bar = new byte[6];
Map<Byte, List<Byte>> indicesMap =
    IntStream.range(0, 6)
        .mapToObj(n -> n)
        .collect(Collectors.groupingBy(
            n -> bar[n],
            Collectors.mapping(
                n -> (byte) n.intValue(),
                Collectors.toList())));

// Exception:
java.lang.NullPointerException: Cannot invoke
"com.sun.tools.javac.tree.JCTree.getStartPosition()"
because "tree" is null
```

## Root Cause

`hasImplicitType()` returns `true` for both:

- `var` declarations
- Implicit lambda parameters (when `VariableTree.getType()` is `null`, per JDK-8268850)

`VarWithPrimitive` used `hasImplicitType()` as its guard, so lambda parameters
with inferred primitive types passed the check and reached
`replaceVariableType()`.

Inside `replaceVariableType()`, `tree.getType()` returns `null` for these
parameters. `hasExplicitSource(null, state)` was then called unconditionally,
which invokes:

```java
getStartPosition(null)
```

resulting in a `NullPointerException`.

## Fix

Add a null guard for `type` in `SuggestedFixes.replaceVariableType()` at the
actual crash site.

If `type == null`:

- Skip the `hasExplicitSource()` call.
- Fall through to `getStartPosition(tree)`.

The subsequent token scan finds no `var` keyword for implicit lambda
parameters and correctly returns `Optional.empty()`, producing `NO_MATCH`
without throwing an exception.

This is the minimal fix because:

- It addresses the actual crash site.
- It is defensive against any other callers that may pass a `VariableTree`
  with a null type.

### `implicitLambdaParameter_noMatch`

Regression test reproducing the exact code from the bug report:

- Single-parameter lambdas
- Inferred `int` and `byte` types
- Verifies `NO_MATCH` and no exception

### `multiParamImplicitLambda_noMatch`

Tests a multi-parameter implicit lambda:

```java
IntBinaryOperator op = (a, b) -> a + b;
```

- Both parameters inferred as `int`
- Verifies `NO_MATCH` and no exception

### `explicitLambdaParam_noMatch`

Tests an explicit lambda parameter:

```java
(int n) -> n
```

- `hasImplicitType()` returns `false`
- Verifies no match is reported

### `enhancedForLoop`

Tests:

```java
for (var x : intArray) {
    // ...
}
```

Expected fix:

```java
for (int x : intArray) {
    // ...
}
```

### `forLoopInitializer`

Tests:

```java
for (var i = 0; i < 10; i++) {
    // ...
}
```

Expected fix:

```java
for (int i = 0; i < 10; i++) {
    // ...
}
```

### DIFF
--- a/core/src/test/java/com/google/errorprone/bugpatterns/VarWithPrimitiveTest.java
+++ b/core/src/test/java/com/google/errorprone/bugpatterns/VarWithPrimitiveTest.java
@@ -17,6 +17,7 @@
 package com.google.errorprone.bugpatterns;
 
 import com.google.errorprone.BugCheckerRefactoringTestHelper;
+import com.google.errorprone.CompilationTestHelper;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
@@ -26,6 +27,9 @@ import org.junit.runners.JUnit4;
 public final class VarWithPrimitiveTest {
   private final BugCheckerRefactoringTestHelper refactoringHelper =
       BugCheckerRefactoringTestHelper.newInstance(VarWithPrimitive.class, getClass());
+
+  private final CompilationTestHelper compilationHelper =
+      CompilationTestHelper.newInstance(VarWithPrimitive.class, getClass());
 
   // from https://openjdk.org/projects/amber/guides/lvti-style-guide#G7
   @Test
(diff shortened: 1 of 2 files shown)


cushon:
Thanks for the fix!

Could you also add a couple more test cases:

* a lambda that has a `var` parameter like `IntUnaryOperator op = (var n) -> n * 2;`
* with an annotation like `@interface A { int var() default 0; }` on the lambda parameter `IntUnaryOperator op = (@A(var = 0) var n) -> n * 2;`, to make sure the replacement doesn't replace the occurrence of `var` in the annotation instead of the type
* a local variable test with `@A(var = 0) var n` also

core/src/test/java/com/google/errorprone/bugpatterns/VarWithPrimitiveTest.java
```diff
@@ -194,6 +198,198 @@ long getAgeAsLong() {
...
+              }
+            }
+            """)
+        .doTest();
+  }
+
+  @Test
+  public void annotatedVarLocalVariable() {
+    // The token scan finds two 'var' tokens (@A(var=0) attribute + type keyword) and safely
+    // returns no fix rather than risk replacing the annotation attribute instead of the type.
```
cushon: The CI shows this test fails on JDK 27 because there's now an AST node for the `var` type, so the replacement succeeds

Can you split this test into separate tests for JDK < 27 and JDK >= 27, and use `assume().that(Runtime.version().feature()).{isAtLeast,isLessThan}(27)` to run the tests on the different JDK versions?

cushon (changes requested):
It looks like the author information for the commits are from two different github accounts and emails, and there is only a CLA on file for one of them

Can you sign the CLA for the other account or update the author information for the commits if it is incorrect?

core/src/test/java/com/google/errorprone/bugpatterns/VarWithPrimitiveTest.java
HarshMehta112: Done — I split the test into JDK < 27 and JDK ≥ 27 variants using assume().that(Runtime.version().feature()).isLessThan(27) and isAtLeast(27).

cushon (approved):

HarshMehta112:
Hi @cushon,

The feedback/copybara service workflow is failing. It was passing a few days ago, but it's now failing. Is there anything I need to do to fix it?

Thanks.

cushon (approved):

PR #5884, opened 2026-06-17
Make `TypeParameterQualifier` Refaster-aware

Stephan202:
For method references, there are valid use cases within Refaster template methods.

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/TypeParameterQualifier.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/TypeParameterQualifier.java
@@ -22,6 +22,7 @@ import static com.google.errorprone.matchers.Description.NO_MATCH;
 import static com.google.errorprone.util.ASTHelpers.findEnclosingMethod;
 import static com.google.errorprone.util.ASTHelpers.getSymbol;
 import static com.google.errorprone.util.ASTHelpers.hasAnnotation;
+import static com.google.errorprone.util.AnnotationNames.AFTER_TEMPLATE_ANNOTATION;
 import static com.google.errorprone.util.AnnotationNames.BEFORE_TEMPLATE_ANNOTATION;
 import static javax.lang.model.element.ElementKind.TYPE_PARAMETER;
 
@@ -61,16 +62,17 @@ public class TypeParameterQualifier extends BugChecker
 
   @Override
   public Description matchMemberReference(MemberReferenceTree tree, VisitorState state) {
-    if (!matchMethodReferences || isInRefasterBeforeTemplate(state)) {
+    if (!matchMethodReferences || isInRefasterTemplate(state)) {
       return NO_MATCH;
     }
     return match(tree, tree.getQualifierExpression(), state);
   }
 
-  private static boolean isInRefasterBeforeTemplate(VisitorState state) {
+  private static boolean isInRefasterTemplate(VisitorState state) {
     MethodTree enclosingMethod = findEnclosingMethod(state);
     return enclosingMethod != null
-        && hasAnnotation(enclosingMethod, BEFORE_TEMPLATE_ANNOTATION, state);
+        && (hasAnnotation(enclosingMethod, BEFORE_TEMPLATE_ANNOTATION, state)
+            || hasAnnotation(enclosingMethod, AFTER_TEMPLATE_ANNOTATION, state));
   }
 
   private Description match(ExpressionTree tree, ExpressionTree qualifier, VisitorState state) {
--- a/core/src/main/java/com/google/errorprone/bugpatterns/TypeParameterQualifier.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/TypeParameterQualifier.java
@@ -19,7 +19,10 @@ package com.google.errorprone.bugpatterns;
 import static com.google.errorprone.BugPattern.SeverityLevel.ERROR;
 import static com.google.errorprone.fixes.SuggestedFixes.qualifyType;
 import static com.google.errorprone.matchers.Description.NO_MATCH;
+import static com.google.errorprone.util.ASTHelpers.findEnclosingMethod;
 import static com.google.errorprone.util.ASTHelpers.getSymbol;
+import static com.google.errorprone.util.ASTHelpers.hasAnnotation;
+import static com.google.errorprone.util.AnnotationNames.BEFORE_TEMPLATE_ANNOTATION;
 import static javax.lang.model.element.ElementKind.TYPE_PARAMETER;
 
 import com.google.errorprone.BugPattern;
@@ -32,6 +35,7 @@ import com.google.errorprone.matchers.Description;
 import com.sun.source.tree.ExpressionTree;
 import com.sun.source.tree.MemberReferenceTree;
 import com.sun.source.tree.MemberSelectTree;
+import com.sun.source.tree.MethodTree;
 import com.sun.tools.javac.code.Symbol;
 import javax.inject.Inject;
 
@@ -57,10 +61,16 @@ public class TypeParameterQualifier extends BugChecker
 
   @Override
   public Description matchMemberReference(MemberReferenceTree tree, VisitorState state) {
-    if (!matchMethodReferences) {
+    if (!matchMethodReferences || isInRefasterBeforeTemplate(state)) {
       return NO_MATCH;
     }
     return match(tree, tree.getQualifierExpression(), state);
+  }
+
+  private static boolean isInRefasterBeforeTemplate(VisitorState state) {
+    MethodTree enclosingMethod = findEnclosingMethod(state);
+    return enclosingMethod != null
+        && hasAnnotation(enclosingMethod, BEFORE_TEMPLATE_ANNOTATION, state);
   }
 
   private Description match(ExpressionTree tree, ExpressionTree qualifier, VisitorState state) {
--- a/core/src/test/java/com/google/errorprone/bugpatterns/TypeParameterQualifierTest.java
+++ b/core/src/test/java/com/google/errorprone/bugpatterns/TypeParameterQualifierTest.java
@@ -144,6 +144,25 @@ public class TypeParameterQualifierTest {
   }
 
   @Test
+  public void methodReference_inAfterTemplate() {
+    compilationHelper
+        .addSourceLines(
+            "Test.java",
+            """
+            import com.google.errorprone.refaster.annotation.AfterTemplate;
+            import java.util.function.Function;
+
+            class Test {
+              @AfterTemplate
+              <T extends Enum<T>> Function<T, String> rule() {
+                return T::name;
+              }
+            }
+            """)
+        .doTest();
+  }
+
+  @Test
   public void memberSelect_inBeforeTemplate() {
     compilationHelper
         .addSourceLines(
--- a/core/src/test/java/com/google/errorprone/bugpatterns/TypeParameterQualifierTest.java
+++ b/core/src/test/java/com/google/errorprone/bugpatterns/TypeParameterQualifierTest.java
@@ -125,6 +125,44 @@ public class TypeParameterQualifierTest {
   }
 
   @Test
+  public void methodReference_inBeforeTemplate() {
+    compilationHelper
+        .addSourceLines(
+            "Test.java",
+            """
+            import com.google.errorprone.refaster.annotation.BeforeTemplate;
+            import java.util.function.Function;
+
+            class Test {
+              @BeforeTemplate
+              <T extends Enum<T>> Function<T, String> rule() {
+                return T::name;
+              }
+            }
+            """)
+        .doTest();
+  }
+
+  @Test
+  public void memberSelect_inBeforeTemplate() {
+    compilationHelper
+        .addSourceLines(
+            "Test.java",
+            """
+            import com.google.errorprone.refaster.annotation.BeforeTemplate;
+
+            class Test {
+              @BeforeTemplate
+              <T extends Enum<T>> T rule(Class<T> clazz, String value) {
+                // BUG: Diagnostic contains: Enum.valueOf(clazz, value);
+                return T.valueOf(clazz, value);
+              }
+            }
+            """)
+        .doTest();
+  }
+
+  @Test
   public void negative() {
     compilationHelper
         .addSourceLines(


Stephan202:
See PicnicSupermarket/error-prone-support#2274 for context. An alternative to this change is #5883, though these PRs are not mutually exclusive.

core/src/test/java/com/google/errorprone/bugpatterns/TypeParameterQualifierTest.java
```diff
@@ -124,6 +124,44 @@ static <T extends Foo> void get() {
...
+        .addSourceLines(
+            "Test.java",
+            """
+            import com.google.errorprone.refaster.annotation.BeforeTemplate;
+            import java.util.function.Function;
+
+            class Test {
+              @BeforeTemplate
+              <T extends Enum<T>> Function<T, String> rule() {
+                return T::name;
```
cpovirk, line 139: Just to be clear: This one could use `Enum::name`, right? It's still reasonable to test it; it's just that the case below is the one in which Refaster needs you to use `T`?

core/src/main/java/com/google/errorprone/bugpatterns/TypeParameterQualifier.java
```diff
@@ -57,12 +61,18 @@ public Description matchMemberSelect(MemberSelectTree tree, VisitorState state)
...
+    if (!matchMethodReferences || isInRefasterBeforeTemplate(state)) {
       return NO_MATCH;
     }
     return match(tree, tree.getQualifierExpression(), state);
   }
 
+  private static boolean isInRefasterBeforeTemplate(VisitorState state) {
+    MethodTree enclosingMethod = findEnclosingMethod(state);
+    return enclosingMethod != null
+        && hasAnnotation(enclosingMethod, BEFORE_TEMPLATE_ANNOTATION, state);
```
cpovirk: Obvious question that might have an obvious answer: Would it make sense to check `@AfterTemplate`, too?

core/src/test/java/com/google/errorprone/bugpatterns/TypeParameterQualifierTest.java
cpovirk: Ah, sorry: Since (related to my other comment!) this is a `@BeforeTemplate`, then [the goal is to match when people write `Foo::name`](https://github.com/PicnicSupermarket/error-prone-support/pull/2274/changes#diff-882708d619bef0d632a4427b56e368a9524d67cf382ce879d59ac2609b3f55bdR64), even though they _could_ write `Enum::name` instead. So, in a Refaster template, unlike in normal code, you very much do want to be able to write `T::name`.

cpovirk:
I also observe that:

- ~At least in the case I tested, our google3 Refaster integration does run Turbine on the Refaster templates, so we would care about references there.~
  - \[edit: I was wrong. See [below](https://github.com/google/error-prone/pull/5884#issuecomment-4733815541).\]
- We could still rely on Turbine to reject them itself. The nightmare scenario would presumably arise if you were to successfully build your own code (but not any tests for it or anything else that depends on it!) and then submit, only for your _users_ to break when they run Turbine on it for the first time. Error Prone protects against that edge case.
- There was a time when Turbine _crashed_ on such code, so we're already in a better position now that it's been reporting a proper error for a long time.
- Perhaps we could allow suppression for _expressions_, which Turbine doesn't care about, but not for _types_—at least not in signatures, since Turbine looks at those?

cpovirk:
> At least in the case I tested, our google3 Refaster integration does run Turbine on the Refaster templates, so we would care about references there.

My mistake: It does _not_ run Turbine on them (in the case I tested); I just saw an error and assumed it was from Turbine when in fact it was from... `TypeParameterQualifier` :)

I do notice that (for better or for worse) it's currently possible to _disable_ `TypeParameterQualifier` with `-Xep:TypeParameterQualifier:OFF`. It's possible that, if we don't do https://github.com/google/error-prone/pull/5884, we should instead eliminate the way of disabling it. Or maybe the current state of affairs is a pragmatic compromise, one that would get even better if we were to merge this particular PR?

cpovirk:
And wait... :)

If I insert the a Refaster template of...

```java
    @BeforeTemplate
    private <T extends java.util.Map<?, ?>> Object o(T.Entry<?, ?> t) {
      return t;
    }
```

...into our `EnumsShortcuts`, then I get _Refaster crashes_ in a couple targets, including `OrderingShortcutsTest_refasters`. This might be the root failure:

```
Jun 17, 2026 11:11:34 AM com.google.devtools.javatools.refactory.MultipassRefactoring analyzeCompilationUnit
SEVERE: Unexpected error in javac analysis of , pass: Test
java.lang.LinkageError
  at com.google.errorprone.refaster.Template.callCheckMethod(Template.java:547)
  at com.google.errorprone.refaster.Template.infer(Template.java:478)
  at com.google.errorprone.refaster.Template.typecheck(Template.java:209)
  at com.google.errorprone.refaster.ExpressionTemplate$2.apply(ExpressionTemplate.java:208)
  at com.google.errorprone.refaster.ExpressionTemplate$2.apply(ExpressionTemplate.java:170)
  at com.google.errorprone.refaster.Choice$2.mapIfPresent(Choice.java:126)
  at com.google.errorprone.refaster.ExpressionTemplate.unify(ExpressionTemplate.java:169)
  at com.google.errorprone.refaster.ExpressionTemplate.match(ExpressionTemplate.java:128)
  at com.google.errorprone.refaster.RefasterScanner.scan(RefasterScanner.java:108)
  at com.google.errorprone.refaster.RefasterScanner.scan(RefasterScanner.java:52)
  at jdk.compiler/com.sun.source.util.TreeScanner.scanAndReduce(TreeScanner.java:95)
  at jdk.compiler/com.sun.source.util.TreeScanner.visitPackage(TreeScanner.java:168)
  at jdk.compiler/com.sun.tools.javac.tree.JCTree$JCPackageDecl.accept(JCTree.java:664)
  at jdk.compiler/com.sun.source.util.TreeScanner.scan(TreeScanner.java:91)
  at com.google.errorprone.refaster.RefasterScanner.scan(RefasterScanner.java:133)
  at com.google.errorprone.refaster.RefasterScanner.scan(RefasterScanner.java:52)
  at jdk.compiler/com.sun.source.util.TreeScanner.visitCompilationUnit(TreeScanner.java:149)
  at jdk.compiler/com.sun.tools.javac.tree.JCTree$JCCompilationUnit.accept(JCTree.java:627)
  at jdk.compiler/com.sun.source.util.TreeScanner.scan(TreeScanner.java:91)
  at com.google.errorprone.refaster.RefasterScanner.scan(RefasterScanner.java:133)
  at com.google.errorprone.refaster.RefasterRule.apply(RefasterRule.java:139)
  at com.google.devtools.javatools.refactory.refaster.RefasterMatcher.analyzeFile(RefasterMatcher.java:71)
  at com.google.devtools.javatools.refactory.refaster.RefasterMatcher.analyzeFile(RefasterMatcher.java:53)
  at com.google.devtools.javatools.refactory.refaster.RefasterMatcher.analyzeCompilationUnit(RefasterMatcher.java:41)
  at com.google.devtools.javatools.refactory.Refactorings.refactorAndReturnReplacements(Refactorings.java:108)
  at com.google.devtools.javatools.refactory.MultipassRefactoring.lambda$analyzeCompilationUnit$0(MultipassRefactoring.java:130)
  at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:328)
  at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1090)
  at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:614)
  at java.base/java.lang.Thread.run(Thread.java:1547)
Caused by: java.lang.AssertionError: isSubtype PACKAGE
  at jdk.compiler/com.sun.tools.javac.code.Types$4.visitType(Types.java:1130)
  at jdk.compiler/com.sun.tools.javac.code.Types$4.visitType(Types.java:1107)
  at jdk.compiler/com.sun.tools.javac.code.Types$DefaultTypeVisitor.visitPackageType(Types.java:4941)
  at jdk.compiler/com.sun.tools.javac.code.Type$PackageType.accept(Type.java:1592)
  at jdk.compiler/com.sun.tools.javac.code.Types$DefaultTypeVisitor.visit(Types.java:4936)
  at jdk.compiler/com.sun.tools.javac.code.Types.isSubtype(Types.java:1103)
  at jdk.compiler/com.sun.tools.javac.code.Types.isSubtypeUncheckedInternal(Types.java:1029)
  at jdk.compiler/com.sun.tools.javac.code.Types.isSubtypeUnchecked(Types.java:1015)
  at jdk.compiler/com.sun.tools.javac.code.Types.isConvertible(Types.java:610)
  at jdk.compiler/com.sun.tools.javac.comp.Resolve$MethodCheckContext.compatible(Resolve.java:1052)
  at jdk.compiler/com.sun.tools.javac.comp.Check.checkType(Check.java:615)
  at jdk.compiler/com.sun.tools.javac.comp.Attr$ResultInfo.check(Attr.java:531)
  at jdk.compiler/com.sun.tools.javac.comp.Resolve$MethodResultInfo.check(Resolve.java:1098)
  at jdk.compiler/com.sun.tools.javac.comp.Resolve$4.checkArg(Resolve.java:918)
  at jdk.compiler/com.sun.tools.javac.comp.Resolve$AbstractMethodCheck.argumentsAcceptable(Resolve.java:803)
  at jdk.compiler/com.sun.tools.javac.comp.Resolve$4.argumentsAcceptable(Resolve.java:927)
  at jdk.compiler/com.sun.tools.javac.comp.Infer.instantiateMethod(Infer.java:184)
  at jdk.compiler/com.sun.tools.javac.comp.Resolve.rawInstantiate(Resolve.java:633)
  at jdk.compiler/com.sun.tools.javac.comp.Resolve.checkMethod(Resolve.java:672)
  at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
  at java.base/java.lang.reflect.Method.invoke(Method.java:565)
  at com.google.errorprone.refaster.Template.callCheckMethod(Template.java:532)
  ... 29 more
```

cpovirk:
For the crash, Gemini says:

```patch
--- a/core/src/main/java/com/google/errorprone/refaster/Template.java  2026-01-29 08:07:06.000000000 -0500
+++ b/core/src/main/java/com/google/errorprone/refaster/Template.java  2026-06-17 14:46:11.000000000 -0400
@@ -203,6 +203,12 @@
       Warner warner,
       List<Type> expectedTypes,
       List<Type> actualTypes) {
+    for (Type type : actualTypes) {
+      TypeTag tag = type.getTag();
+      if (tag == TypeTag.PACKAGE || tag == TypeTag.METHOD || tag == TypeTag.MODULE || tag == TypeTag.FORALL) {
+        return Optional.empty();
+      }
+    }
     try {
       ImmutableList<UTypeVar> freeTypeVars = freeTypeVars(unifier);
       var unused =
```

Maybe? I didn't test whether the template from https://github.com/PicnicSupermarket/error-prone-support/pull/2274 would still work with that change.

cpovirk:
~[That change](https://github.com/google/error-prone/pull/5884#issuecomment-4734199781) breaks a number of existing Refaster tests.~

\[edit: Wrong again; see [below](https://github.com/google/error-prone/pull/5884#issuecomment-4734246895).\]

cpovirk:
Wait, no, sorry again: What breaks the tests is my test template with `Map.Entry`, which matches `null` (which is a `Map.Entry`!) all over. The actual attempted fix doesn't break any tests that I found.

core/src/test/java/com/google/errorprone/bugpatterns/TypeParameterQualifierTest.java
Stephan202: Exactly :)

core/src/main/java/com/google/errorprone/bugpatterns/TypeParameterQualifier.java
Stephan202: Yeah, good question. I thought I convinced myself that the answer is "no", but now that I think back to what I tested: the answer is actually "yes". `T::method` will be replaced with `ConcreteType::method`, and one may reasonably want that in case e.g. a rule wishes to rewrite some context around arbitrary `T::method`. I've pushed another commit. :)

Stephan202 (commented):
Tnx for the review @cpovirk!

> Maybe? I didn't test whether the template from [PicnicSupermarket/error-prone-support#2274](https://github.com/PicnicSupermarket/error-prone-support/pull/2274) would still work with that change.

I can quickly try that in a bit; we have test cases for each rule, so shouldn't be hard.

core/src/main/java/com/google/errorprone/bugpatterns/TypeParameterQualifier.java
Stephan202: (Generalized the PR description accordingly.)

Stephan202:
> > Maybe? I didn't test whether the template from [PicnicSupermarket/error-prone-support#2274](https://github.com/PicnicSupermarket/error-prone-support/pull/2274) would still work with that change.
> 
> I can quickly try that in a bit; we have test cases for each rule, so shouldn't be hard.

That change doesn't break our build :heavy_check_mark:

core/src/main/java/com/google/errorprone/bugpatterns/TypeParameterQualifier.java
cpovirk: In hindsight, I guess that [my other thread](https://github.com/google/error-prone/pull/5884#discussion_r3430321827) implies that an _`@AfterTemplate`_ could always use `BoundOfTheTypeParameter::method` instead of `T::method`.

The former might actually be safer: If `BoundOfTheTypeParameter::method` compiles in the template, then it's very likely to compile in the output, whereas `T::method` could perhaps become ambiguous when replaced with `ConcreteType::method`. For example, `Enum::valueOf` is clearly [`valueOf(Class, String)`](https://docs.oracle.com/en/java/javase/26/docs/api/java.base/java/lang/Enum.html#valueOf(java.lang.Class,java.lang.String)), but `ConcreteType::valueOf` could be either that or the `valueOf(String)` method generated on the enum class `ConcreteType`.

The former also feels less like a spiritual violation of [NonCanonicalType](https://errorprone.info/bugpattern/NonCanonicalType) / [NonCanonicalStaticImport](https://errorprone.info/bugpattern/NonCanonicalStaticImport). But when we discussed that very thing (in the context of `List::stream` vs. `Collection::stream`) internally a couple months ago in bug 411135416, we decided that we didn't actually care :)

Those don't seem like deal-breakers. If someone wants to produce `ConcreteType::name`, then sure, I _guess_?

It does seem believable that there would be cases in which we don't have a specific known type for `T`, in which case generating a literal "`T::whatever`" would be bad. Still... I guess it's OK-ish? Or we should catch it inside Refaster if we don't already? But I guess I could see that as a somewhat more compelling reason to have not allowed `T` qualifiers in the `@AfterTemplate`.

And finally: I am belatedly recognizing that your implementation and your test take the position that there might very well _not_ be a good case for either a `@BeforeTemplate` or an `@AfterTemplate` that contains `T.someMethod()`: `someMethod` would have to be a static method that is probably better referred to through the actual declaring type, as in [the other comment thread](https://github.com/google/error-prone/pull/5884/changes/BASE..d4b553c5fd6fb66de8c02bb3c5bf51b5769cd60c#r3431165668). Additionally, the ban of `T.<anything>` has the virtue of preserving the ban on `T.Foo` in the _signature_ (return types / parameters / type parameters) of the Refaster template, allowing qualification on `T` only for method references. (So [my earlier Turbine experiments](https://github.com/google/error-prone/pull/5884#issuecomment-4733756919) were focused on a case whose behavior you haven't even been proposing to change!)

After all that, I'm probably more on the side of your original approach of omitting this from the PR :) But I don't feel confident after all the knots I've unnecessarily tied myself into here. Let's at least wait to see if @cushon has higher-level concerns about the direction of the two PRs.

core/src/test/java/com/google/errorprone/bugpatterns/TypeParameterQualifierTest.java
cpovirk: Thanks. I'm now grasping the other test shows that we continue _not_ to try to support matches on `Foo.valueOf(Foo.class, value)` (or especially `Foo.valueOf(Bar.class, value)`!), which is _definitely_ weirder than writing `Enum.valueOf(Foo.class, value)` :) Now I suppose that we could still want to provide a way to match the latter (to steer people away from it!), but I am happy to stop here now that I hopefully actually understand what's going on.

cpovirk (commented):
Thanks for testing the possible fix!

core/src/main/java/com/google/errorprone/bugpatterns/TypeParameterQualifier.java
Stephan202: I'm totally fine either way :). You make a compelling argument, though dropping the `@AfterTemplate` exemption does make it harder to write a rule that matches-without-modification of that particular sub-expression. There does exist an escape hatch for that, though, through the introduction of a `@Matches(IsSomeMethodReference.class)`-guarded template parameter. Significantly more work for the user, but as this would be rare, it is not a strong argument for keeping the PR as-is. 

If we drop the second commit, let's remember to also revert the PR description body (as Copybara transfers that to the final commit message, I've observed).

core/src/test/java/com/google/errorprone/bugpatterns/TypeParameterQualifierTest.java
Stephan202: Since there is (IIRC) another Error Prone rule (or is it CheckStyle?) that nudges users to using `DeclaringType.staticMethod` instead of `SomeSubType.staticMethod`, I suspect there is less need for supporting this. But I also can't come up with a strong argument for _not_ supporting this. Sometimes those static methods are "overridden". 

Note that if we _do_ also want to allow this, the PR actually becomes simpler, and the code likely more efficient. Cause we can then simply set:
```java
@BugPattern(
    summary = "Type parameter used as type qualifier",
    severity = ERROR,
    suppressionAnnotations = BeforeTemplate.class)
```

That would make the change set class with #5883, but that's not a biggie.

Let's unresolve this thread for input by @cushon.

core/src/main/java/com/google/errorprone/bugpatterns/TypeParameterQualifier.java
cpovirk: Good call on the PR description.

I'm realizing that the bottom line here might be this: Turbine doesn't care about anything that this CL affects.

PR #5889, opened 2026-06-19
fix(#5845): strip parentheses before matching in 6 bug checks to eliminate false positives/negatives

### LINKED ISSUE #5845: Multiple checks are not parenthesis-invariant: matchers don't look through ParenthesizedTree
### Error Prone version

Error Prone 2.49.0 (javac 21.0.8, `--source/--target 21`)

### Affected checks

`LoopConditionChecker`, `NonAtomicVolatileUpdate`, `PatternMatchingInstanceof`, `DuplicateBranches`, `InlineTrivialConstant`, `AssignmentExpression`

### Description

A number of unrelated checks share a single root cause: their AST matchers do not look through `ParenthesizedTree`. Per JLS §15.8.5, parentheses around an expression do not change its value, type, or constant-expression status — they are semantically inert. Wrapping (or unwrapping) an operand in redundant parentheses is therefore a no-op rewrite, yet across at least six checks it flips the verdict — silencing true findings (false negatives), and in one case manufacturing a false positive.

Each case below is a self-contained, minimal `BEFORE`/`AFTER` pair where the only difference is one pair of redundant parentheses. I verified all of them on Error Prone 2.49.0; the per-check counts are shown in the table.

| Check | Rewrite | BEFORE | AFTER | Effect |
|---|---|:--:|:--:|---|
| `LoopConditionChecker` | `i++` → `(i)++` | 0 | 1 | **false positive** |
| `NonAtomicVolatileUpdate` | `x++` → `(x)++` | 4 | 0 | false negative |
| `PatternMatchingInstanceof` | `(String) o` → `(String)(o)` | 1 | 0 | false negative |
| `DuplicateBranches` | `1 * 5` → `(1 * 5)` | 1 | 0 | false negative |
| `InlineTrivialConstant` | `""` → `("")` | 1 | 0 | false negative |
| `AssignmentExpression` | `a = b = 0` ↔ `a = (b = 0)` | 2 | 3 | verdict drift (both directions) |

The common fix is to skip/unwrap enclosing `ParenthesizedTree` nodes before matching the relevant operand (e.g. via `ASTHelpers.stripParentheses` on the operand of the increment/cast/branch/initializer/assignment being inspected).

#### 1. `LoopConditionChecker` — `(i)++` produces a false positive

`(i)++` denotes the same variable update as `i++`, so the loop's condition variable `i` is still modified in the update. The check reports it as never modified (this manifests when the loop bound is a compile-time constant).

```java
// BEFORE — no finding
public class LoopParen {
  void m() {
    for (int i = 0; i < 10; i++) {
      System.out.println(i);
    }
  }
}
```

```java
// AFTER — false positive: "condition variable(s) never modified in loop body: i"
public class LoopParen {
  void m() {
    for (int i = 0; i < 10; (i)++) {
      System.out.println(i);
    }
  }
}
```

#### 2. `NonAtomicVolatileUpdate` — `(x)++` is missed

`(x)++` is the same non-atomic read-modify-write of a volatile field as `x++`.

```java
// BEFORE — 4 findings
public class Vol {
  volatile int x;
  volatile long y;
  void bad() { x++; y++; x--; y--; }
}
```

```java
// AFTER — 0 findings
public class Vol {
  volatile int x;
  volatile long y;
  void bad() { (x)++; (y)++; (x)--; (y)--; }
}
```

#### 3. `PatternMatchingInstanceof` — `(String)(o)` is missed

Per JLS §15.16 the parentheses around a cast operand are a no-op, but the instanceof-and-cast pattern is no longer recognized.

```java
// BEFORE — 1 finding (suggests `if (o instanceof String s)`)
public class Sample {
  public int f(Object o) {
    if (o instanceof String) {
      String s = (String) o;
      return s.length();
    }
    return o.hashCode();
  }
}
```

```java
// AFTER — 0 findings
public class Sample {
  public int f(Object o) {
    if (o instanceof String) {
      String s = (String)(o);
      return s.length();
    }
    return o.hashCode();
  }
}
```

#### 4. `DuplicateBranches` — `(1 * 5)` defeats the branch comparison

The two ternary branches remain identical; only one is parenthesized.

```java
// BEFORE — 1 finding
public class Sample {
  int x;
  void m() { x = true ? 1 * 5 : 1 * 5; }
}
```

```java
// AFTER — 0 findings
public class Sample {
  int x;
  void m() { x = true ? 1 * 5 : (1 * 5); }
}
```

#### 5. `InlineTrivialConstant` — `("")` is missed

```java
// BEFORE — 1 finding
public class Sample {
  private static final String EMPTY = "";
}
```

```java
// AFTER — 0 findings
public class Sample {
  private static final String EMPTY = ("");
}
```

#### 6. `AssignmentExpression` — verdict is not parenthesis-invariant

The check flags an assignment used inside a larger expression, but the result depends on redundant parentheses around the inner assignment — in **both** directions: `a = b = 0` is not flagged while `a = (b = 0)` is, whereas `result = (bresult = new byte[len])` is flagged while `result = bresult = new byte[len]` is not.

```java
// BEFORE — 2 findings (on `int i = j = 0;` and `result = (bresult = ...)`)
public class Sample {
  int a, b, j, l;
  byte[] bresult;
  void m1() { a = b = 0; }                        // NOT flagged
  void m2() { if ((a = b = 0) != 1) {} }          // NOT flagged
  void m3() { int i = j = 0; }                    // flagged
  void m4(int len) {
    int k = (l += 1);                             // NOT flagged
    byte[] result;
    result = (bresult = new byte[len]);           // flagged
  }
}
```

```java
// AFTER — 3 findings; per-site verdicts flip as parentheses are added/removed
public class Sample {
  int a, b, j, l;
  byte[] bresult;
  void m1() { a = (b = 0); }                      // now flagged
  void m2() { if ((a = (b = 0)) != 1) {} }        // now flagged
  void m3() { int i = (j = 0); }                  // flagged
  void m4(int len) {
    int k = l += 1;                               // NOT flagged
    byte[] result;
    result = bresult = new byte[len];             // now NOT flagged
  }
}
```

### Expected behavior

Each check should produce the same findings on the `BEFORE` and `AFTER` of every pair above. Redundant parentheses do not change the value, type, or constant-ness of the wrapped expression, so they must not change any check's verdict.

### Actual behavior

See the table: each rewrite that only adds or removes one pair of redundant parentheses changes the finding set — five checks drop true findings, one (`LoopConditionChecker`) raises a false positive, and `AssignmentExpression` drifts in both directions. The shared cause is that the matchers inspect the raw operand AST instead of unwrapping enclosing `ParenthesizedTree` nodes.

HarshMehta112:
Fixes: #5845 

## Problem
                                                                                                                                                                                          
  Six unrelated checks share a root cause: their AST matchers inspect raw `ExpressionTree`
  operands without unwrapping enclosing `ParenthesizedTree` nodes. Per JLS §15.8.5, a
  parenthesized expression denotes the same variable as its contained expression —
  parentheses are semantically inert. Because `ASTHelpers.getSymbol(ParenthesizedTree)`
  returns `null` (no case for `PARENTHESIZED` in the dispatch), wrapping or unwrapping one
  pair of redundant parentheses around an operand silently flips the verdict.

  | Check | Reproducer | Direction |
  |---|---|---|
  | `LoopConditionChecker` | `for (int i = 0; i < 10; (i)++) {}` | false positive |
  | `NonAtomicVolatileUpdate` | `(x)++` on a volatile field | false negative |
  | `PatternMatchingInstanceof` | `String s = (String)(o)` | false negative |
  | `DuplicateBranches` | `true ? 1 * 5 : (1 * 5)` | false negative |
  | `InlineTrivialConstant` | `private static final String EMPTY = ("")` | false negative |
  | `AssignmentExpression` | `a = b = 0` vs `a = (b = 0)` verdict drift | both directions |

  ---

  ## Fix
                                                                                                                                                                                          
  Apply `ASTHelpers.stripParentheses()` at the point where each check extracts the operand
  it will inspect.

  ### `LoopConditionChecker`
  `UpdateScanner.check()`: strip before `getSymbol`. One change covers all four call sites
  (`visitUnary`, `visitAssignment`, `visitCompoundAssignment`, `visitMethodInvocation`).

  ### `NonAtomicVolatileUpdate`
  Strip in all three extractor lambdas (`expressionFromUnaryTree`,
  `variableFromCompoundAssignmentTree`, `variableFromAssignmentTree`) before the
  `hasModifier(VOLATILE)` check.

  ### `PatternMatchingInstanceof`
  Strip `node.getExpression()` in `findAllCasts.visitTypeCast` before the
  `constantExpression` symbol lookup.

  ### `DuplicateBranches`
  Strip both arms before `toString()` comparison. Source positions for the suggested fix
  still use the original (possibly-parenthesized) tree, so the replacement is always
  syntactically valid.

  ### `InlineTrivialConstant`                                                                                                                                                
  Strip `tree.getInitializer()` before the `STRING_LITERAL` kind check and value                                                                                                          
  extraction.

  ### `AssignmentExpression`
  Walk `getParentPath()` in a loop until the leaf is no longer `ParenthesizedTree`, then                                                                                     
  use the resulting `effectiveParent` for all three downstream decisions:                                                                                                                 
  - binary-context exemption (`(foo = bar) != null`)
  - duplicate-assignment detection (`a = a = foo()`)
  - chained-assignment exemption (`x = y = 0`)

  ---

  ## Tests
                                                                                                                                                                                          
  32 new test cases across all six checkers:

  - Single and multiple layers of redundant parentheses
  - All operator forms: postfix/prefix increment/decrement, compound assignment (`+=`, `-=`),
    simple assignment
  - Negative cases confirming legitimate no-find verdicts are unaffected (e.g. non-volatile
    fields, synchronized blocks, different branch content after stripping)
  - Realistic patterns: negated if/else, ternary arms, do-while/while loop bodies, chained
    assignments, inline cast use, duplicate assignment through parentheses

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/AssignmentExpression.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/AssignmentExpression.java
@@ -33,6 +33,7 @@ import com.sun.source.tree.LambdaExpressionTree;
 import com.sun.source.tree.ParenthesizedTree;
 import com.sun.source.tree.Tree;
 import com.sun.source.tree.VariableTree;
+import com.sun.source.util.TreePath;
 import com.sun.tools.javac.code.Symbol.VarSymbol;
 
 /** A BugPattern; see the summary. */
@@ -55,18 +56,23 @@ public final class AssignmentExpression extends BugChecker implements Assignment
     if (parent instanceof LambdaExpressionTree) {
       return Description.NO_MATCH;
     }
+    // Walk through any enclosing parentheses to find the effective parent context.
+    TreePath effectiveParentPath = state.getPath().getParentPath();
+    while (effectiveParentPath.getLeaf() instanceof ParenthesizedTree) {
+      effectiveParentPath = effectiveParentPath.getParentPath();
+    }
+    Tree effectiveParent = effectiveParentPath.getLeaf();
     // Exempt the C-ism of (foo = getFoo()) != null, etc.
-    if (parent instanceof ParenthesizedTree
-        && state.getPath().getParentPath().getParentPath().getLeaf() instanceof BinaryTree) {
+    if (effectiveParent instanceof BinaryTree) {
       return Description.NO_MATCH;
     }
     // Detect duplicate assignments: a = a = foo() so that we can generate a fix.
-    if (isDuplicateAssignment(tree, parent)) {
+    if (isDuplicateAssignment(tree, effectiveParent)) {
       return describeMatch(
           tree, SuggestedFix.replace(tree, state.getSourceForNode(tree.getExpression())));
     }
     // If we got here it's something like x = y = 0, which is odd but not disallowed.
-    if (parent instanceof AssignmentTree) {
+    if (effectiveParent instanceof AssignmentTree) {
       return Description.NO_MATCH;
     }
     return describeMatch(tree);
--- a/core/src/main/java/com/google/errorprone/bugpatterns/DuplicateBranches.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/DuplicateBranches.java
@@ -20,6 +20,7 @@ import static com.google.common.collect.Iterables.getLast;
 import static com.google.errorprone.BugPattern.SeverityLevel.WARNING;
 import static com.google.errorprone.matchers.Description.NO_MATCH;
 import static com.google.errorprone.util.ASTHelpers.getStartPosition;
+import static com.google.errorprone.util.ASTHelpers.stripParentheses;
 import static java.util.stream.Collectors.joining;
 
 import com.google.errorprone.BugPattern;
@@ -31,6 +32,7 @@ import com.google.errorprone.matchers.Description;
 import com.google.errorprone.util.ErrorProneTokens;
 import com.sun.source.tree.BlockTree;
 import com.sun.source.tree.ConditionalExpressionTree;
+import com.sun.source.tree.ExpressionTree;
 import com.sun.source.tree.IfTree;
 import com.sun.source.tree.Tree;
 
@@ -66,7 +68,9 @@ public class DuplicateBranches extends BugChecker
     // do exactly what we want here, which is to compare the syntax including of identifiers and
     // not their underlying symbols, and it would require a lot of case work to implement for all
     // AST nodes.
-    if (!thenTree.toString().equals(elseTree.toString())) {
+    Tree strippedThen = thenTree instanceof ExpressionTree et ? stripParentheses(et) : thenTree;
+    Tree strippedElse = elseTree instanceof ExpressionTree et ? stripParentheses(et) : elseTree;
+    if (!strippedThen.toString().equals(strippedElse.toString())) {
       return NO_MATCH;
     }
     int start = getStartPosition(elseTree);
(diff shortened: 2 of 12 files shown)


HarshMehta112:
Hi @cushon , when you get a chance, could you please have a look at the PR and share any feedback or suggestions? Thanks!

Issue #5938, opened 2026-07-01
GuardedBy false positive with Runnable passed to a user method known to invoke synchronously


EamonTracey:
The `@GuardedBy` annotation has a false positive when we pass a lambda or method reference that accesses guarded items, even if we know it is invoked immediately synchronously.

Consider the following case:
```java
class Transaction {
  @GuardedBy("this")
  int x;

  public synchronized void handle() {
    doSomething(() -> {
      x++;  // compilation error, requires extraneous synchronized block to satisfy the checker
    });
  }

  private void doSomething(Runnable r) {
    r.run();
  }
}
```

The above code fails to compile without adding a `synchronized` block inside the lambda. This grows cumbersome in codebases where this pattern is common, for example, a `safeRun` method that wraps a `Runnable` with error handling and logging.

I suggest we add a parameter annotation `@RunsImmediately` to indicate the call is safe without the extraneous synchronization block. There already exists a [whitelist for common built-in Java functions](https://github.com/google/error-prone/blob/be8896aee16f3de5f95d188f3ba1045e36e872ac/core/src/main/java/com/google/errorprone/bugpatterns/threadsafety/HeldLockAnalyzer.java#L68-L89), so this would allow people to whitelist their own methods.

cushon:
Instead of an annotation, we could add a flag to the check to make that list of well known APIs configurable.

For the error prone annotations artifact we try to guarantee there will never be breaking changes to the annotations, and there’s a higher bar to adding annotations compared to flags.

EamonTracey:
Makes sense, will update the PR to introduce a new flag rather than annotation

PR #5940, opened 2026-07-01
Add KnownImmediateMethods flag to avoid @GuardedBy false positives on immediate invocations of guarded access

EamonTracey:
Addresses https://github.com/google/error-prone/issues/5938

Adds the `GuardedBy:KnownImmediateMethods` flag, which marks methods that invoke lambdas / method references / runnables immediately in the caller's lock scope. This allows users to avoid false positive `@GuardedBy` checks.

For example, with `-XepOpt:GuardedBy:KnownImmediateMethods=com.example.Transaction#doSomething`:
```java
class Transaction {
  @GuardedBy("this")
  int x;

  public synchronized void handle() {
    doSomething(() -> {
      ++x;  // OK: 'doSomething' is configured to run the lambda immediately.
    });
  }

  private void doSomething(Runnable r) {
    r.run();
  }
}
```


core/src/main/java/com/google/errorprone/bugpatterns/threadsafety/HeldLockAnalyzer.java
```diff
@@ -88,6 +90,32 @@ public final class HeldLockAnalyzer {
...
+    MethodSymbol sym = ASTHelpers.getSymbol(invocation);
+    if (sym == null) {
+      return false;
+    }
+    List<VarSymbol> params = sym.getParameters();
+    int index = invocation.getArguments().indexOf(functionalArgument);
+    if (index < 0 || params.isEmpty()) {
+      return false;
+    }
+    VarSymbol param = index < params.size() ? params.get(index) : params.getLast();
```
EamonTracey: this logic specifically supports varargs, which is the case where `index >= params.size()`

EamonTracey:
Hey @kluever and @cpovirk 👋! I am seeking a code review for this contribution; are either of you are available to do it? This addresses https://github.com/google/error-prone/issues/5938

EamonTracey:
Politely bumping @kluever @cpovirk 😃

cushon:
I suggested an alternative in the bug that I’d prefer to an annotation for this: https://github.com/google/error-prone/issues/5938#issuecomment-5131298373

core/src/main/java/com/google/errorprone/bugpatterns/threadsafety/HeldLockAnalyzer.java
EamonTracey: no longer relevant

EamonTracey:
@cushon Thanks for the suggestion, updated to introduce a new flag rather than annotation. This PR is ready for review

EamonTracey:
@cushon Bumping again!

EamonTracey:
Hey @kluever @cpovirk @cushon could I get a review here?

Issue #1431, opened 2019-12-03
PreferJavaTimeOverload false positive for Duration#dividedBy(long)


michaelhixson:
`PreferJavaTimeOverload` reports an incorrect warning when it sees `Duration#dividedBy(long)`.

Example code:
```java
Duration totalTime = Duration.ofSeconds(10);
long numberOfPeople = 5; // does not represent a Duration
Duration timePerPerson = totalTime.dividedBy(numberOfPeople);
```

Expected behavior: No warnings.

Actual behavior:
> [PreferJavaTimeOverload] If the numeric primitive (numberOfPeople) represents a Duration, please call totalTime.dividedBy(Duration) instead.

I'm using Error Prone 2.3.4.

---

The bug pattern seems to want me to use `Duration#dividedBy(Duration)` instead, but these two methods aren't interchangeable.  They answer different questions.

https://docs.oracle.com/en/java/javase/13/docs/api/java.base/java/time/Duration.html#dividedBy(long)
https://docs.oracle.com/en/java/javase/13/docs/api/java.base/java/time/Duration.html#dividedBy(java.time.Duration)

I notice there is a TODO for inspecting the return types in the bug pattern:
https://github.com/google/error-prone/blob/df4cdefbf351c76098562c2d83eebf4cafaa8eea/core/src/main/java/com/google/errorprone/bugpatterns/time/PreferJavaTimeOverload.java#L345-L348

I tried resolving that TODO by adding another condition...
```java
&& isSameType(input.getReturnType(), calledMethod.getReturnType(), state)
```

...and that made the incorrect warning go away.

kluever:
Thanks - will get this fixed up ASAP

kluever:
Will be mirrored out shortly.

PR #5987, opened 2026-07-31
Address `zizmor` findings:

cpovirk:
- [`persist-credentials: false`](https://docs.zizmor.sh/audits/#artipacked)
- [pinning `uses`](https://docs.zizmor.sh/audits/#unpinned-uses)

There may be other findings, but these are the easy ones.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,39 +47,41 @@ jobs:
     continue-on-error: ${{ matrix.experimental }}
     steps:
       - name: Cancel previous
-        uses: styfle/cancel-workflow-action@0.13.1
+        uses: styfle/cancel-workflow-action@d07a454dad7609a92316b57b23c9ccfd4f59af66 # 0.13.1
         with:
           access_token: ${{ github.token }}
       - name: 'Check out repository'
-        uses: actions/checkout@v7
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+        with:
+          persist-credentials: false
       - name: 'Set up JDK 21'
-        uses: actions/setup-java@v5
+        uses: actions/setup-java@03ad4de0992f5dab5e18fcb136590ce7c4a0ac95 # v5
         with:
           java-version: 21
           distribution: 'zulu'
           cache: 'maven'
       - name: 'Set up JDK 25'
-        uses: actions/setup-java@v5
+        uses: actions/setup-java@03ad4de0992f5dab5e18fcb136590ce7c4a0ac95 # v5
         with:
           java-version: 25
           distribution: 'zulu'
           cache: 'maven'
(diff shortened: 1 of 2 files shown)


cpovirk:
JDK 23 CI has been failing since https://github.com/google/error-prone/pull/5979. Separate from this PR, I'm going to fix the failure and fix the misconfiguration that allowed Copybara to merge PRs that were failing that one particular check.

cpovirk:
Now actually passing the `zizmor` (GitHub Actions Scan) check.

I have gotten myself mixed up a few times with `zizmor`, and I'm not entirely sure whether it's actually misbehaving in some cases, whether I'm holding it wrong, or whether I've just been mistaken about things as I juggle changes to various repos.

One place where I _have_ just been mistaken is that I've assume that `zizmor` is blocking internal submits via the Copybara presubmit. That's not true, since Copybara waits only for CI tasks that it's explicitly configured to wait for (as seen recently in our discussions about JDK 23). I also have come to _suspect_ that it's not blocking even merges in GitHub, since I don't see Branch Protection rules that request that, even though it shows up as a failing check. (I don't know what Branch Protection rules Error Prone has in place at the moment, anyway.) But I don't have a handy PR available to check that on, and I'm not invested enough to create one.

All that is separate from the question of whether we might _want_ `zizmor` failures to block changes, such as if someone sends us a PR to allegedly update some GitHub Actions action to a new release but that actually [points to something evil](https://instatunnel.my/blog/github-ghost-commit-smuggling-hiding-in-the-detached-head). `zizmor` might catch cases in which the comment is wrong (though I haven't verified _that_ to 100% confidence, either).

One thing I note on this specific PR is that it seems a bit strange for `zizmor` to [insert comments of the  form"`v7`,"](https://github.com/google/error-prone/pull/5987/changes#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR54), since I wonder if that tag might get force-pushed to the latest `v7.x` release periodically? But I'm happy to move in the general correct direction and to satisfy `zizmor` today. We can always do more in the future.

(IIRC, we merge changes to `.github` directly in GitHub, since we don't have those files in our internal repo? I would do it myself for this + https://github.com/google/turbine/pull/466 + https://github.com/google/google-java-format/pull/1431 (and for https://github.com/google/error-prone/pull/5988 once we have a direction set there), but I don't have permissions.)

cushon:
Thanks!

> IIRC, we merge changes to .github directly in GitHub, since we don't have those files in our internal repo?

Yes these files are extenal SoT and aren't synced, to make it easier to iterate on CI config changes.

I can merge this one and [#1431](https://github.com/google/google-java-format/pull/1431), are they ready to go?

cpovirk:
Thanks, yes, they should be good, along with [the Turbine one](https://github.com/google/turbine/pull/466) that you may have had reasons for not mentioning yet.

PR #5988, opened 2026-07-31
Simplify JDK matrix:

cpovirk:
- We don't need to list "JDK 25 on ubuntu-latest" in `include` because it's covered by the generated `os`-`java` cross product.
- We can add 23 to the `java` list so that "JDK 23 on ubuntu-latest" is generated automatically instead of having to list it in `include`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-latest ]
-        java: [ 25, 21 ]
+        java: [ 25, 23, 21 ]
         experimental: [ false ]
         # Only test on macos and windows with a single recent JDK to avoid a
         # combinatorial explosion of test configurations.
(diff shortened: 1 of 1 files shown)


cushon:
Thanks. I'm trying to remember why we even have 23, it may have been to navigate breaking changes in 23 before JDK 25 was released. At this point it might make sense to just have the two latest LTSs and EA, or maybe also the latest STS. I don't expect a lot of usage of JDK 23 now that there's a newer LTS.

cpovirk:
That is pretty much exactly how I have looked at it for Guava. So naturally I'm here because of [a test that fails under JDK 23](https://github.com/google/error-prone/pull/5989) :) But that's unlikely to happen with any frequency, and also it doesn't at all invalidate your point that zero end users will be affected.

Whatever we do here with 23, we should reconcile it with the Copybara config that I changed in cl/957613778 but lacked permissions to put into effect.

The _other_ failing check here will be addressed by https://github.com/google/error-prone/pull/5987, which is mostly about pinning actions by hash. After that, we should tell Dependabot to recreate https://github.com/google/error-prone/pull/5990 so that it will use hashes for it.

PR #6055, opened 2026-08-21
Fix a Refaster crash on `module-info.java` files

rickie:
`UMemberSelect#visitIdentifier` unconditionally dereferences `sym.owner`, but module symbols have no owner: `Symbol.ModuleSymbol#create` always passes `null`. Any Refaster rule whose `@BeforeTemplate` is rooted at an instance member select then crashes the compiler as soon as it scans a `requires` directive whose module name is a single identifier.

Module names that contain a dot are parsed as member selects instead, and `#visitMemberSelect` is already null-safe; as every JDK module name contains a dot, this has so far gone unnoticed.

Reported at
https://github.com/PicnicSupermarket/error-prone-support/issues/2316.


cpovirk (approved):
Thanks.

(I see [a very similar check in `EqualsNaN`](https://github.com/google/error-prone/blob/372f5551a27102bc965b856a26cccb2495f353dc/core/src/main/java/com/google/errorprone/bugpatterns/EqualsNaN.java#L80), and I'll bet it's not the only existing place where we do that.)

PR #6069, opened 2026-08-26
Recognize split declaration/assignment in UnnecessaryStringBuilder

tanvir-ux:
## Description

Fixes #5828.

`UnnecessaryStringBuilder` already flags locals like:

```java
StringBuilder sb = new StringBuilder("x");
```

but misses the equivalent split form:

```java
StringBuilder sb;
sb = new StringBuilder("x");
```

This change treats a sole initializing assignment the same as a declarator
initializer, and ignores that assignment when checking whether the local
still needs to stay a `StringBuilder`.

## Testing

```
mvn -pl core -am -Dtest=UnnecessaryStringBuilderTest -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false test
```

All `UnnecessaryStringBuilderTest` cases passed (including new split-form coverage).

PR #6070, opened 2026-08-26
Recognize split declaration/assignment in UnnecessaryAsync

### LINKED ISSUE #5829: `UnnecessaryAsync` misses the split declaration/assignment form
### Error Prone version

Error Prone 2.49.0 (javac plugin) — javac 21.0.8, source/target 21.

### Check name

[UnnecessaryAsync](https://errorprone.info/bugpattern/UnnecessaryAsync)

### Description

`UnnecessaryAsync` flags atomic variables that are initialized and do not escape the current scope (so the atomicity is unnecessary). The matcher recognizes the combined declarator-with-initializer form `AtomicInteger a = new AtomicInteger(0)`. Splitting it into `AtomicInteger a; a = new AtomicInteger(0);` is a semantics-preserving rewrite — the variable is still initialized exactly once before use and still does not escape — yet the rule stops firing.

### Reproducer

```java
// BEFORE — 1 finding
import java.util.concurrent.atomic.AtomicInteger;
class C {
  int f() {
    AtomicInteger a = new AtomicInteger(0);
    a.incrementAndGet();
    return a.get();
  }
}
```

```java
// AFTER — equivalent rewrite, 0 findings
import java.util.concurrent.atomic.AtomicInteger;
class C {
  int f() {
    AtomicInteger a;
    a = new AtomicInteger(0);
    a.incrementAndGet();
    return a.get();
  }
}
```

### Expected behavior

`UnnecessaryAsync` should report the same finding on BEFORE and AFTER, since the variable remains initialized-once and non-escaping in both forms.

### Actual behavior

- BEFORE: 1 finding (`Variables which are initialized and do not escape the current scope ...`).
- AFTER: 0 findings.

The matcher only recognizes the combined declarator-with-initializer form and misses the equivalent split-assignment form.

tanvir-ux:
## Description

Fixes #5829.

`UnnecessaryAsync` already flags locals like:

```java
AtomicInteger a = new AtomicInteger(0);
```

but misses the equivalent split form:

```java
AtomicInteger a;
a = new AtomicInteger(0);
```

This change finds a sole initializing assignment when the declarator has
no initializer, treats that assignment as non-escaping (same as the
combined form), and removes the assignment statement when applying the
suggested rewrite to a plain local.

## Testing

```
mvn -pl core -am -Dtest=UnnecessaryAsyncTest -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false test
```

All 13 `UnnecessaryAsyncTest` cases passed (including new split-form coverage).

### DIFF
--- a/core/src/main/java/com/google/errorprone/bugpatterns/UnnecessaryAsync.java
+++ b/core/src/main/java/com/google/errorprone/bugpatterns/UnnecessaryAsync.java
@@ -33,6 +33,8 @@ import com.google.errorprone.bugpatterns.BugChecker.VariableTreeMatcher;
 import com.google.errorprone.fixes.SuggestedFix;
 import com.google.errorprone.matchers.Description;
 import com.google.errorprone.matchers.Matcher;
+import com.sun.source.tree.AssignmentTree;
+import com.sun.source.tree.ExpressionStatementTree;
 import com.sun.source.tree.ExpressionTree;
 import com.sun.source.tree.IdentifierTree;
 import com.sun.source.tree.LambdaExpressionTree;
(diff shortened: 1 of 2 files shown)


tanvir-ux:
Ready for review when you have bandwidth — adds split declaration/assignment coverage for `UnnecessaryAsync` (paired with #6069 for `UnnecessaryStringBuilder`).

Issue #6079, opened 2026-08-31
Per-check timing spans cost a few percent of the compile, and nothing reads the timings


vlsi:
`ErrorProneScanner.processMatchers` opens a timing span for every matcher it runs on every AST node:

```java
try (AutoCloseable unused = oldState.timingSpan(matcher)) {
```

`VisitorState.timingSpan` forwards to `ErrorProneTimings.span`, which looks the check up by canonical name, starts a `Stopwatch`, and returns a closure that stops it. Every span is therefore a `HashMap` lookup, two `System.nanoTime` calls, and a closure.

Nothing in this repository reads what the timers record: `ErrorProneTimings.timings()` and `initializationTime()` have no callers in `error_prone_check_api` or `error_prone_core`, and no test asserts on either. I assume there is an internal consumer, since the class arrived in cl/245821511 for that purpose, but every build that uses the released artifact pays for data it has no way to read.

`span` has one caller, so it is not that some spans are coarse and useful while others are fine-grained: all of them wrap one matcher on one node.

## Cost

Two workloads, both compiled in process so that the numbers are the compiling thread's own, on macOS on Apple silicon.

**`error_prone_core` itself**, 948 sources, JDK 24, default check set. One compilation opens 14771508 spans, so it makes 29543016 `System.nanoTime` calls. Comparing that build against one whose `span` returns a shared no-op, alternating the two inside a single JVM so each pair meets the same machine state, the spans cost a median of **656 ms against a floor of 13.0 s**, positive in 8 pairs out of 8.

**Calcite's `:core`**, 1655 sources, JDK 21, with NullAway and the project's own check selection. One compilation opens 28493608 spans. The spans cost a median of **1216 ms against a floor of 37 s**, positive in 4 pairs out of 4, and — unlike the first workload — **0.42 GiB of allocation**:

| | allocated by the compiling thread |
| --- | --- |
| spans on | 21.03 GiB |
| spans off | 20.61 GiB |

The closure is what accounts for that. It does not show up on `error_prone_core`, where C2 inlines `span` into `processMatchers` and scalar-replaces the closure; it does show up on Calcite, whose larger check set defeats that inlining. So whether the allocation is real depends on the JIT, and a project cannot tell which case it is in.

## What the spans measure

Recording each span's duration alongside the per-check totals, on `error_prone_core`:

| span duration | share of spans | share of measured time |
| --- | --- | --- |
| under 128 ns | 45.1% | 2.9% |
[...]

PR #6080, opened 2026-08-31
Record check timings only when -XepPrintTimings asks for them

vlsi:
## Why

`ErrorProneScanner.processMatchers` opens a timing span for every matcher it runs on every AST node, and nothing reads what the spans record: `ErrorProneTimings.timings()` and `initializationTime()` have no callers in `error_prone_check_api` or `error_prone_core`, and no test asserts on either.

The spans are not cheap at that frequency. Compiling `error_prone_core` itself opens 14771508 of them, and Calcite's `:core` opens 28493608, each a `HashMap` lookup, two `System.nanoTime` calls, and a closure. Against a build whose `span` returns a shared no-op, alternating the two inside one JVM so each pair meets the same machine state, they cost a median of 656 ms of a 13.0 s compile and 1216 ms of a 37 s one, and on Calcite 0.42 GiB of allocation.

[#6079](https://github.com/google/error-prone/issues/6079) has the full measurements.

## What

`-XepPrintTimings` gates the spans, so a compilation that does not pass it opens none. The same flag prints the per-check totals when the compilation finishes, so the data reaches whoever paid for it:

```text
Error Prone ran 440 checks in 582 ms, and spent 338 ms initializing
       172 ms   29.6%  ParameterName
        27 ms    4.8%  AvoidCommonTypeNames
        24 ms    4.2%  AutoValueSubclassLeaked
```

Every check that ran gets a row, so the count in the header is the number of rows below it.

An embedder that reads `timings()` programmatically now needs the flag too. That is the part worth arguing with, and #6079 says why I think the default belongs this way round: the collection has no consumer in this repository, and the flag is what gives it one.

Making the collection itself cheap enough to leave on is a larger change, and #6079 proposes it separately.

## How to verify

```bash
mvn -pl check_api,core test -Dtest=ErrorProneOptionsTest,ErrorProneJavaCompilerTest
```

`recognizesPrintTimings` and `printTimingsIsOffByDefault` cover the option. `printTimingsReportsEveryCheckThatRan` compiles a real file and asserts both a non-zero count in the header and at least one row: the header alone is printed before the rows, so a build that collected nothing would still print one. `withoutPrintTimingsNoReportIsPrinted` covers the default.

Each test was watched failing: deleting the option's parse case fails the first, inverting the gate in `VisitorState.timingSpan` fails the third on both of its assertions, and dropping the `Kind.COMPILATION` branch fails it on the absent header.

One limitation the tests do not cover: under Gradle the report does not reach the build log, because Gradle passes no writer to `JavaCompiler.getTask` and javac's `NOTICE` writer then goes to the compiler daemon's stderr. It arrives through any caller that supplies a writer. Emitting it as a diagnostic instead would fix that, and the same applies to what `RefactoringCollection` already prints; I left both alone here.

### DIFF
--- a/check_api/src/main/java/com/google/errorprone/ErrorProneAnalyzer.java
+++ b/check_api/src/main/java/com/google/errorprone/ErrorProneAnalyzer.java
@@ -19,9 +19,11 @@ package com.google.errorprone;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Throwables.getStackTraceAsString;
 import static com.google.common.base.Verify.verify;
+import static java.util.Comparator.comparing;
 
 import com.google.common.base.Supplier;
 import com.google.common.base.Suppliers;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.errorprone.BugPattern.SeverityLevel;
 import com.google.errorprone.ErrorProneOptions.Severity;
@@ -45,7 +47,10 @@ import com.sun.tools.javac.util.Log;
 import com.sun.tools.javac.util.Log.WriterKind;
 import com.sun.tools.javac.util.PropagatedException;
 import java.io.PrintWriter;
+import java.time.Duration;
 import java.util.HashSet;
+import java.util.Locale;
+import java.util.Map;
 import java.util.Set;
 import javax.tools.JavaFileObject;
 import org.safere.Pattern;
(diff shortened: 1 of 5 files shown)


Stephan202:
This looks to be an alternative to #1474. Would still be nice to have something like this :)

vlsi:
I should have checked the past PRs on that matter. I did not expect there would be 6-year-old PR doing the same thing 😂

cushon (approved):

cushon:
I'm going to try to get this merged. I'm planning to separate out the configuration for recording and printing timing information with something like `-XepPrintTimings` and `-XepRecordTimings`. (I'm happy to make those changes when importing it, just mentioning it here if there's feedback.)

vlsi:
>I'm happy to make those changes when importing it, just mentioning it here if there's feedback.

Thank you for picking this up. Yeah, separate options are good as well.

PR #6108, opened 2026-09-14
Don't treat java.lang.Record#equals as a custom equals implementation.

### LINKED ISSUE #6107: FN: `EqualsIncompatibleType` misses comparisons between unrelated record types
## What version of Error Prone are you using?

2.50.0

## Does this issue reproduce with the latest release?

Yes.

## What did you do?

```java
public class Test {
  record UserId(long id) {}

  record OrderId(long id) {}

  boolean f(UserId userId, OrderId orderId) {
    return userId.equals(orderId);
  }
}
```

## What did you expect to see?

`[EqualsIncompatibleType] Calling equals on incompatible types UserId and OrderId`

## What did you see instead?

No diagnostic. Comparisons where exactly one side is a record are flagged as expected,
so the gap is specific to record-to-record comparisons:

| comparison | flagged |
| --- | --- |
| `record` vs `java.util.UUID` | yes |
| `record` vs `enum` | yes |
| unrelated `final class` vs unrelated `final class` (both with a hand-written `equals`) | yes |
| `record` vs unrelated `record` | **no** |

`TypeCompatibility#isFeasiblyCompatible` treats two types as compatible when an
`equals(Object)` override on the left type is inherited by the right type.
`java.lang.Record` declares `public abstract boolean equals(Object)`, and
`customEqualsMethod` only excludes `Object` and `Enum`, so every pair of records
shares `Record.equals` and is considered compatible.

Like `Object.equals` and `Enum.equals`, `Record.equals` specifies no equality
semantics that two distinct record types could share, so it looks like it belongs
in the same exclusion. This is the same shape as #679 (enums) and #2384 (protos).

Records are increasingly used as wrapper types for IDs and other values, where
mixing two of them up is exactly the mistake this check is meant to catch.

rincho0617:
Fixes #6107.

`isFeasiblyCompatible` considers two types compatible when an `equals(Object)`
override on one is inherited by the other. `java.lang.Record` declares
`public abstract boolean equals(Object)`, so every pair of records shares it and
`EqualsIncompatibleType` never fires on record-to-record comparisons.

`Record.equals` is abstract and specifies no equality semantics that two distinct
record types could share, so this excludes it alongside `Object` and `Enum`.

Comparisons where only one side is a record were already flagged; this only adds
the record-to-record case. `EqualsIncompatibleTypeTest` covers both the new
positive case and that two values of the same record type stay comparable.


cpovirk (approved):
Yikes, thanks. 

We may need to flag-guard this for at least our internal rollout, but I can deal with that there.

PR #6131, opened 2026-09-19
[MisformattedTestData] avoid re-formatting diagnostic message comments

### LINKED ISSUE #6130: MisformattedTestData suggested fix wraps long `// BUG: Diagnostic contains: ` lines
For example:
```diff
-                // BUG: Diagnostic contains: Method violates has @Contract(-> !null), but this appears to be violated, as a @Nullable value may be returned when the contract preconditions are true.
+                // BUG: Diagnostic contains: Method violates has @Contract(-> !null), but this appears to be
+                // violated, as a @Nullable value may be returned when the contract preconditions are true.
```
This changes what gets asserted by the test, making it risky to apply the suggested fix.

msridhar:
Fixes #6130 

Works by swapping out such comments with placeholders before invoking GJF and then restoring them.

Assisted-by: Codex (GPT-5)


core/src/main/java/com/google/errorprone/bugpatterns/MisformattedTestData.java
```diff
@@ -156,6 +159,63 @@ public Void visitLiteral(LiteralTree node, Void unused) {
...
+      placeholderPrefix += "_";
+    }
+
+    List<String> comments = new ArrayList<>();
+    List<String> lines = LINE_SPLITTER.splitToList(source);
+    StringBuilder result = new StringBuilder(source.length());
+    boolean inBugMarkerComment = false;
+    for (int i = 0; i < lines.size(); i++) {
+      String line = lines.get(i);
+      String strippedLine = line.stripLeading();
```
cpovirk: Maybe we should just stick with `trim()`, gross an API as it is, to match `DiagnosticTestHelper`?

core/src/main/java/com/google/errorprone/bugpatterns/MisformattedTestData.java
```diff
@@ -156,6 +159,63 @@ public Void visitLiteral(LiteralTree node, Void unused) {
...
+              || strippedLine.startsWith(DIAGNOSTIC_MATCHES_MARKER);
+      if (!startsBugMarkerComment && !(inBugMarkerComment && strippedLine.startsWith("//"))) {
+        inBugMarkerComment = false;
+      } else {
+        inBugMarkerComment = true;
+        int commentStart = line.indexOf("//");
+        comments.add(line.substring(commentStart));
+        line =
+            line.substring(0, commentStart)
+                + SourceWithProtectedBugMarkers.placeholder(placeholderPrefix, comments.size() - 1);
```
cpovirk: Thoughts on whether we need a unique marker for each comment? Since we replace them all in order, it seems like we could probably use the same marker for each. But I guess then we'd either have to use `replaceFirst` (which uses regex, so we'd need `Pattern.quote`) or roll our own `indexOf` + `substring` approach (though perhaps we could use that as an opportunity to make the code slightly more efficient with `StringBuilder` instead of repeated `replace`, even though I doubt that that matters). I guess that overall pushes me toward your current approach?

core/src/main/java/com/google/errorprone/bugpatterns/MisformattedTestData.java
```diff
@@ -156,6 +159,63 @@ public Void visitLiteral(LiteralTree node, Void unused) {
 
   private static final Splitter LINE_SPLITTER = Splitter.on('\n');
 
+  private static final String DIAGNOSTIC_CONTAINS_MARKER = "// BUG: Diagnostic contains:";
```
cpovirk: Do you think it would be worth moving these constants and `protectBugMarkerComments` to live inside `SourceWithProtectedBugMarkers`?

core/src/main/java/com/google/errorprone/bugpatterns/MisformattedTestData.java
msridhar: Yeah, I was a bit concerned about the efficiency of doing a repeated `replace`.  Probably we could re-implement `replace` to go line by line, finding relevant comment markers and replacing them in order, so we wouldn't need unique markers then either.  I don't think this'll make much performance difference in the average case given the typical size of a test input, but I guess there could be extreme cases where it'd make a difference.  It'll be a bit more complicated code-wise also.  Do you prefer I change `replace` to the line-by-line approach?  If so, happy to do it.

core/src/main/java/com/google/errorprone/bugpatterns/MisformattedTestData.java
msridhar: Sure, done

core/src/main/java/com/google/errorprone/bugpatterns/MisformattedTestData.java
msridhar: This one is done

core/src/main/java/com/google/errorprone/bugpatterns/MisformattedTestData.java
```diff
@@ -156,6 +160,66 @@ public Void visitLiteral(LiteralTree node, Void unused) {
...
+
+      List<String> comments = new ArrayList<>();
+      List<String> lines = LINE_SPLITTER.splitToList(source);
+      StringBuilder result = new StringBuilder(source.length());
+      boolean inBugMarkerComment = false;
+      for (int i = 0; i < lines.size(); i++) {
+        String line = lines.get(i);
+        String trimmedLine = line.trim();
+        boolean startsBugMarkerComment =
+            trimmedLine.startsWith(DIAGNOSTIC_CONTAINS_MARKER)
```
vlsi, line 188: `DiagnosticTestHelper` recognizes a marker anywhere on a line (`line.contains(BUG_MARKER_COMMENT_INLINE)`), but this checks `startsWith` on the trimmed line, so a marker that follows code still gets wrapped. With google-java-format 1.36.1, this test data:

```java
    try {} catch (Exception e) {
    } // BUG: Diagnostic contains: expected phrase that is sufficiently long that google-java-format would otherwise wrap it
```

is formatted as:

```java
    } // BUG: Diagnostic contains: expected phrase that is sufficiently long that google-java-format
      // would otherwise wrap it
```

The helper then reads `would otherwise wrap it` as a second expected pattern, which is the #6130 problem again. There's at least one marker like this in the repository (`EmptyCatchTest.java:56`). Could `protectBugMarkerComments` find the marker with `indexOf` and protect the text from the marker onward? If trailing markers are out of scope on purpose, a test that pins that down would help.

core/src/main/java/com/google/errorprone/bugpatterns/MisformattedTestData.java
```diff
@@ -156,6 +160,66 @@ public Void visitLiteral(LiteralTree node, Void unused) {
...
+                || trimmedLine.startsWith(DIAGNOSTIC_MATCHES_MARKER);
+        if (!startsBugMarkerComment && !(inBugMarkerComment && trimmedLine.startsWith("//"))) {
+          inBugMarkerComment = false;
+        } else {
+          inBugMarkerComment = true;
+          int commentStart = line.indexOf("//");
+          comments.add(line.substring(commentStart));
```
vlsi, line 195: `line.substring(commentStart)` keeps any trailing whitespace, including a `\r`, and `restore` puts it back after the formatter has stripped it. For `// BUG: Diagnostic contains: foo   ` the suggested text block keeps the three trailing spaces. Test data whose only problem is that trailing whitespace is no longer reported. Calling `stripTrailing()` here would fix both, and the helper trims each pattern anyway.

core/src/test/java/com/google/errorprone/bugpatterns/MisformattedTestDataTest.java
```diff
@@ -152,6 +152,102 @@ void method() {
         .doTest();
   }
 
+  @Test
+  public void bugMarkerComment_notWrapped() {
```
vlsi, line 156: Neither new test has a long ordinary `//` comment outside a marker block. So both still pass if the checker protected every `//` line, or if the `inBugMarkerComment = false` reset were removed, and either change would stop the checker from reformatting ordinary comments. A test with a marker, a statement, and then a long ordinary comment that still gets wrapped would catch both.

`Diagnostic matches:` has no test either: dropping `|| trimmedLine.startsWith(DIAGNOSTIC_MATCHES_MARKER)` leaves both tests green.

core/src/test/java/com/google/errorprone/bugpatterns/MisformattedTestDataTest.java
```diff
@@ -152,6 +152,102 @@ void method() {
...
+
+            class Test {
+              void method(BugCheckerRefactoringTestHelper h) {
+                h.addInputLines(
+                    "Test.java",
+                    \"""
+                    class Test {
+                      void method() {
+                        // BUG: Diagnostic contains: expected phrase that is sufficiently long that google-java-format would otherwise wrap it onto another comment line and change the test's meaning
+                        int a =
```
vlsi, line 171: Nit: `int a =` / `1;` is only there to make the checker suggest a fix, and a reader could take it for part of what's being tested. `.addInputLines(...).expectUnchanged()` would state the contract directly, that a long marker alone isn't a misformatting. It fails the same way without the fix and would cut the test in half. Otherwise, a one-line comment explaining why the statement is misformatted would help.

core/src/main/java/com/google/errorprone/bugpatterns/MisformattedTestData.java
```diff
@@ -156,6 +160,66 @@ public Void visitLiteral(LiteralTree node, Void unused) {
...
 
+  private record SourceWithProtectedBugMarkers(
+      String source, String placeholderPrefix, List<String> comments) {
+
+    private static final String DIAGNOSTIC_CONTAINS_MARKER = "// BUG: Diagnostic contains:";
+
+    private static final String DIAGNOSTIC_MATCHES_MARKER = "// BUG: Diagnostic matches:";
+
+    /**
+     * Replace error diagnostic comments with placeholders so they are not modified during
```
vlsi, line 171: Nit: these Javadoc summaries could be sentences, for example `Replaces bug marker comments with placeholders so the formatter leaves them unchanged.` and `Restores the original bug marker comments in the formatted source.`

vlsi (changes requested):
Thanks, this fixes the case from #6130, and the continuation rule matches `DiagnosticTestHelper.extractPatterns`. I ran the protect/restore logic against google-java-format 1.36.1 on a few edge cases; the inline comments have the details. The trailing-marker gap and the missing negative test are the two I'd address before merging.

One related problem this PR doesn't cause: a marker inside Javadoc, as in `InvalidParamTest`, `InvalidBlockTagTest`, and `EmptyBlockTagTest`, is reflowed by the Javadoc formatter. With a long enough pattern, `* @param <T> baz // BUG: Diagnostic contains: <long text>` gets its tail wrapped onto the next `*` line. The helper reads only the marker line, so the test keeps passing and asserts a truncated pattern. That may be worth a separate issue.

core/src/main/java/com/google/errorprone/bugpatterns/MisformattedTestData.java
msridhar: Thanks for identifying this case!  Probably we can handle as you suggest; I'll look into it.

core/src/main/java/com/google/errorprone/bugpatterns/MisformattedTestData.java
msridhar: Do you mean just add a `.` to the end?  If we start with `Replaces` / `Restores` then I don't think it's a valid English sentence anymore?

msridhar:
@vlsi I believe I addressed all your feedback, thanks for the review!

PR #6137, opened 2026-09-22
Don't suggest making JUnit 5 test methods `private` in `ExposedPrivateType`

rickie:
`shouldReduceVisibility` exempts reflectively invoked test methods using `JUnitMatchers.TEST_CASE`, which covers JUnit 3 and 4 only. Jupiter test methods were therefore privatised, which silently stops them running.

Matching `@ParameterizedTest` requires looking through the `@TestTemplate` meta-annotation, hence `TypePredicates#hasAnnotation` and `Matchers#hasMetaAnnotation`.

This PR ports over a few Matchers we in Error Prone Support.


check_api/src/main/java/com/google/errorprone/predicates/TypePredicates.java
```diff
@@ -50,6 +50,11 @@ public static TypePredicate isPrimitive() {
     return (type, state) -> type.isPrimitive();
   }
 
+  /** Match types that are annotated with the given annotation. */
+  public static TypePredicate hasAnnotation(String annotationType) {
```
rickie, line 54: Didn't introduce a test for this method, as I suspect there is a test class that's not open sourced right now. Let me know if I should add it :)

rickie (commented):
Let me know if it's preferred to keep the PR smaller and some stuff should be simplified. This however, seemed like a better step towards using more of these JUnit 5 matchers in the future.

cushon (approved):
Thanks!

rickie:
Thanks @cushon !

